PR #6, opened 2023-12-04
feat: transition force exercise to solvency-check model

dyedm1:


contracts/PanopticPool.sol
```diff
@@ -1220,6 +1167,82 @@ contract PanopticPool is ERC1155Holder, Multicall {
         s_collateralToken0.refund(account, msg.sender, refundAmounts.rightSlot());
```
guil-lambert, line 1059: Can you move the `getRefundAmounts` to `PanopticPool.sol`? It feels a bit weird to call `s_collateralToken1` from `s_collateralToken0` and I think we should have enough bytecode space to call it locally

contracts/PanopticPool.sol
```diff
@@ -1220,6 +1167,82 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+                msg.sender,
+                twapTick,
+                positionBalanceArray,
+                portfolioPremium.leftSlot()
+            );
+
+            (uint256 balanceCross, uint256 thresholdCross) = _getSolvencyBalances(
+                // multiply the collateral requirement by 1.333... to create a buffer (CR*1.3333... = CR + CR/3)
+                tokenData0.toLeftSlot(tokenData0.leftSlot() / 3),
+                tokenData1.toLeftSlot(tokenData1.leftSlot() / 3),
```
guil-lambert: Try to avoid hardcoding it. Can you pass on a new parameter in `getAccountMarginDetails` called `bool maintenanceCalculation` which adds the maintenance margin buffer there instead? 

The same `maintenanceCalculation=true` flag could be used in `takeCommissionAddData` when it calls `getAccountMarginDetails` there too

guil-lambert (changes requested):
See comments

contracts/PanopticPool.sol
dyedm1: I did a big internal logic refactor and directed all the solvency check paths into one function with an adjustable buffer - it actually removes a ton of data being passed down functions and between contracts that way. The pathway for minting options is much simpler now that we have removed the rolls and moved the calculations out of the collateralTracker

contracts/CollateralTracker.sol
```diff
@@ -1485,18 +1392,23 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
-                        (s_maintenanceMarginRatio * uint128(-premiumAllPositions)) /
-                        DECIMALS;
+                    tokenRequired += uint128(-premiumAllPositions);
                 }
             }
         }
 
-        // get the user's shares balance (amount of collateral);
-        uint256 collateralAmount = balanceOf[user];
+        uint256 collateralAmount = convertToAssets(balanceOf[user]);
```
guil-lambert, line 1400: Good catch!

contracts/PanopticPool.sol
```diff
@@ -548,13 +553,25 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
         // update the users options balance of position 'tokenId'
         // note: user can't mint same position multiple times, so set the positionSize instead of adding
         _setUserOptionsBalance(msg.sender, tokenId, positionSize, poolUtilizations);
 
+        if (
+            !_checkSolvency(
+                msg.sender,
+                positionIdList,
+                newTick,
+                tickStateCallContext.medianTick(),
```
guil-lambert, line 570: medianTick is not up to date (as of the `556: updateMedian(newTick);`)

However, do we really want to use the median tick for solvency? I know we need to make sure the price is not in a manipulated state, but that could cut us the other way too.

**eg:** a user could open up a _lot_ of new positions after seeing a large price move, and do so within the few minutes necessary for the medianTick to catch up with the "real" tick, effectively knowing the future price and buying calls for essentially free money?

contracts/CollateralTracker.sol
dyedm1: It was actually correct before lol, it was just converted while adding it to the LeftRight. I downstreamed this change from the liquidations since it was the only other place we used the maintenance margin in

contracts/PanopticPool.sol
dyedm1: I have my concerns about the median tick being inconsistent with the TWAP tick solvency checks used elsewhere, but this is not one of them. 

1. We have the 100% collateral requirement when the deviation exceeds the maximum tick deviation
2. The median tick is used for solvency, not settlement, so the only consequence would be an increased amount of leverage compared to what is normally available. There's no mechanism that would actually result in them making more profit based on the medianTick

The reason we use the median tick is to prevent the current tick from being manipulated for more leverage on low-liquidity pools

> medianTick is not up to date (as of the 556: updateMedian(newTick);)

It's not supposed to be - this is logically identical to the previous sequence of operations. The updateMedian is just a poke we include in there to increase the update frequency.

guil-lambert: We were doing the checks using the "after-mint" `currentTick` before:

https://github.com/panoptic-labs/panoptic-v1-core-private/blob/63d4a23aa208305c5fc09d6779648f713834ddb1/contracts/PanopticPool.sol#L599C29-L599C29

>We have the 100% collateral requirement when the deviation exceeds the maximum tick deviation

That final `solvencyCheck` would require 100% collateralization for the new positions anyways, so using the `medianTick` could make new positions solvent when they're not.

>There's no mechanism that would actually result in them making more profit based on the medianTick

~~Maybe there is? Let me know if that makes sense:~~ Nah, this attack would also require 100% collateral requirement, so the check would require 100% for new medianTick options, not 20%.. 

1. A natural spike in price leads to a increases 50% in 1 block. It'll take 4mins for the medianTick to catch up.
2. Account A mints several call options near medianTick. Collateral requirement thinks the position is OTM (20% of notional) but it receives an ITM amounts when swapping (50% of notional). They do this in a loop & mint infinite call options or drain the call side of the collateralTokens.
4. Do nothing and wait 4 minutes. Median tick is newTick but the account is still solvent because they own all token1 and collateralRequirement < requiredBalance

Compare this with the collateral requirement calculation at `newTick` but with collateral requirement at 100%, and ~~that attack is not possible~~. The difference here is that they'd get the ITM amount "for free" so maybe they can lever-up or grift by making the protocol more insolvent?

dyedm1: > We were doing the checks using the "after-mint" currentTick before:
Oh, I see where the confusion is
We were using
- the currentTick to determine the position requirements
- the medianTick for the cross-collateral values of the tokens

And now we're just using the medianTick for everything

> Do nothing and wait 4 minutes. Median tick is newTick but the account is still solvent because they own all token1 and collateralRequirement < requiredBalance

No, they wouldn't be solvent if they didn't have the basal collateral requirement in their account in the first place - any ITM amounts they receive will be matched by a corresponding increase in their collateral requirement once at `newTick`. The only consequences of this "attack" are an inordinate amount of leverage for that account and a quick liquidation. 

(tokenType=ETH 1E=1KUSD)
CR=100% (medianTick is too far away)
Mint 1 ETH call, CR=1ETH, get 0.5 ETH in ITM amount
use USDC to collateralize the other 0.5 eth, so 500USDC
*price moves back to 50% up*, 1500USD/ETH
CR=100% (it doesn't change after you mint the pos)
CR =  1ETH(basal) + 0.5 ETH (ITM)
CB = 500USDC=0.333ETH + 0.5ETH
Very insolvent

PR #7, opened 2023-12-06
feat: add a solvency check at burn

dyedm1:

PR #8, opened 2023-12-11
Liquidations V2

guil-lambert:


guil-lambert:
built on the https://github.com/panoptic-labs/panoptic-v1-core-private/pull/7 PR, new commits start here: [335e6e7](https://github.com/panoptic-labs/panoptic-v1-core-private/pull/8/commits/335e6e77d53baca3297fb8fabb6984f550d7113e)

contracts/CollateralTracker.sol
```diff
@@ -1292,6 +1199,9 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                 tokenToPay += intrinsicValue;
             }
 
+            // only return positive paidAmount
+            paidAmount = tokenToPay < 0 ? int128(0) : int128(tokenToPay);
```
dyedm1: The paidAmount should be accurate. I'm not sure exactly what the intention of this is, but the reason we have the paidAmount is so we can get an accurate representation of the user's balance post the transaction  without querying their balances. We make decisions on how to allocate tokens to pay back the liquidator + bonus based on this, so it needs to be correct.

dyedm1:
Added working pass/fail fuzz tests for the liquidations and fixed the CollateralTracker tests/removed the long-runnning/broken tests. Had to make a few fixes based on the results:

 - `exercise` was returning the paid amount as math.max(0, tokenToPay) instead of just tokenToPay which was screwing up the results.
 - the balance in `tokenData` includes positive accumulated premium, which is also part of the `netExchanged` amount (i.e double-counted). It has to be in `tokenData` for solvency calculations, so I just subtracted the positive premium where it's used to determine the balance while doing the protocol loss compensation.
 - we were using the final price instead of the TWAP price when converting tokens to calculate the solvency-based liquidation bonus. this makes no sense because the collateral requirements are themselves calculated at the TWAP price, meaning that the bonus is not reflective of the actual solvency at *any* price and can even be negative. I just set up the bonus calculation to use the TWAP price to convert tokens for solvency instead. We can talk about calculating the bonus value at both prices and maybe taking the minimum or something, but being internally consistent is good enough, so let's ship this change and think about that later.
 - Not a bug, but I converted the codebase over to math.min/max/abs instead of ternaries since we're using them a lot more and the math functions are a lot more readable.

dyedm1 (approved):
LGTM, lets ship

contracts/CollateralTracker.sol
dyedm1: fixed

guil-lambert:
I'll merge this one but I'll test out a few more features in a separate PR 👍

PR #9, opened 2023-12-18
fix: prevent transfers of removedLiquidity

dyedm1:
We need to disable transfers of removedLiquidity because they can result in misrepresentation of the liquidity utilization/premium multiplier to protocols if chunks are transferred to them before users mint any positions there.

Also, they shouldn't be possible anyway in general because having a long position means that there's a corresponding short position (i.e >1 touches to a chunk)


guil-lambert:
LGTM!

PR #11, opened 2023-12-20
fix: prevent reentrancy on uninitialized pools

dyedm1:
It's possible to reenter an uninitialized SFPM pool by initializing it during a reentrant call because the lock is reset. To fix this, we just persist the lock state through initialization.


guil-lambert (approved):
LGTM

PR #12, opened 2023-12-20
fix: add reentrancy checks to SFPM transfer functions

dyedm1:
The SFPM transfers allow transferring tokens of reentrancy locked pools right now, which can create issues with ERC777 tokens that can make reentrant calls to transfer tokens before the fee growth accumulators are updated. If the fee growth is transferred to another account prematurely, that account will potentially be able to collect more fees than they are due since the differences between the current and stored fee growths will be incorrect.


guil-lambert (approved):
LGTM

PR #17, opened 2023-12-20
fix: more robust validation for v3 callbacks to avoid birthday attack

dyedm1:
Although Uniswap uses the same method to validate router callbacks as we do, it's still vulnerable to a birthday attack feasible with (a significant portion of) today's computing power. Obviously, it's practically impossible to find a key that matches with a targeted address, but if you mine two sets: 
1. deployed Uniswap addresses with the canonical factory, a real token and fee, and one token used as a fake.
2. EOA addresses

you can steal funds from the real token just by finding a collision between those two sets. (big argument over this in Kyber audit contest: https://github.com/sherlock-audit/2023-07-kyber-swap-judging#issue-m-2-routersol-is-vulnerable-to-address-collission)

Also, some chains like ZKSync have weird/nonstandard contract address derivation mechanisms, so if we want to accommodate those chains in the future we can't rely on repeating the derivation ourselves to recover the correct pool address.

The solution is just to call `getPool` on Uniswap to check if the calling address is the canonical pool. This does increase the gas a little bit, but it's well worth it for the added robustness we get out of it.


contracts/libraries/CallbackLib.sol
```diff
@@ -1,9 +1,12 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity ^0.8.0;
 
+// Interfaces
+import {IUniswapV3Factory} from "univ3-core/interfaces/IUniswapV3Factory.sol";
 // Libraries
 import {Constants} from "@libraries/Constants.sol";
 import {Errors} from "@libraries/Errors.sol";
+import "forge-std/Test.sol";
```
guil-lambert: remove Test.sol

guil-lambert (approved):

PR #18, opened 2023-12-20
fix: round collectedAmount calculation down for Uniswap fee parity

dyedm1:
[A] - [B] <= [A - B] does not always hold true


contracts/SemiFungiblePositionManager.sol
```diff
@@ -1407,6 +1430,16 @@ contract SemiFungiblePositionManager is ERC1155, Multicall {
                         tempChunk
                     );
                     amountToCollect = feesBase.sub(s_accountFeesBase[positionKey]);
+
+                    // If the current feesBase is close or identical to the stored one, the amountToCollect can be negative.
+                    // This is because the stored feesBase is rounded up, and the current feesBase is rounded down.
+                    // When this is the case, we want to behave as if there are 0 fees, so we just rectify the values.
+                    amountToCollect = amountToCollect.rightSlot() >= 0
```
guil-lambert: This rectification logic is repeated twice: here and on lines 1229-1234.

Would it make sense to create a new function in LeftRight.sol called `subRect` which would perform this rectification in that call, without requiring to re-extract the `rightSlot()+leftSlot()`? 

We would call
- `int256 amountToCollect = _getFeesBase(univ3pool, startingLiquidity, liquidityChunk, false)
            .subRect(s_accountFeesBase[positionKey]);`
- `amountToCollect = feesBase.subRect(s_accountFeesBase[positionKey])`

guil-lambert (changes requested):
See comment above about `subRect`

contracts/SemiFungiblePositionManager.sol
dyedm1: Fair point, we already mask out the values in `sub` so we can save a few ops by reusing.

PR #19, opened 2024-01-02
fix: prevent overflows of removedLiquidity

dyedm1:
removedLiquidity can be overflowed by accumulating tokens corresponding to a greater amount of liquidity than what was initially removed. For some liquidity chunks, 1 unit of positionSize corresponds to <1 unit of liquidity, so the liquidity returned can be greater than the liquidity removed if that accumulated fractional liquidity adds up to 1 or more.


guil-lambert (approved):
A-OK

PR #21, opened 2024-01-10
feat: use virtual shares for force exercises instead of delegation

dyedm1:
Currently, we require force exercisors to delegate 100% of the collateral that could be required to exercise an option to ensure it can be burned (the exercised account must be solvent, but they may not have the right tokens to exercise the option on their own). 

This results in force exercising being significantly capital-intensive, especially for larger options that could require a large amount of tokens to be provided upfront. 

Because the only purpose of the delegation is to ensure that the burn settlement logic can execute without reverting, it's not necessarily important for the force exercisor to be the one that provides the tokens. In fact, we can even accomplish the same purpose by minting temporary (or "virtual") shares to the exercised account, without requiring anyone to provide tokens upfront!

This change can make force exercises less capital-intensive as well as simplifying the force exercise calculation on the UI and increasing the probability a user will be able to generate a valid force exercise sequence.

To accomplish this, we add counterparts of the `delegate` and `refund` function that mint and burn shares to the exercised account without updating `totalSupply`. The flow goes like so:
1. Call `delegate(delegatee, uint256)` to mint virtual shares to the delegatee, ensuring the position burn can be settled
2. Burn the position
3. Call `refund(delegator, delegatee, uint256)` to settle the force exercise fee and any substitute tokens (i.e if the position needed USDC to burn but the account only had ETH, so we send USDC to the delegator from the delegatee and get ETH back) between the delegator and delegatee
4. Call `refund(delegatee, uint256)` to burn the virtual shares that were minted to the delegatee

We can do this for liquidations too, but it's less of an urgent UX improvement and the delegatedAmounts are more difficult to get.

### DIFF
--- a/test/foundry/core/CollateralTracker.t.sol
+++ b/test/foundry/core/CollateralTracker.t.sol
@@ -194,6 +194,14 @@ contract PanopticPoolHarness is PanopticPool {
         collateralToken.delegate(delegator, delegatee, assets);
     }
 
+    function delegate(
+        address delegatee,
+        uint128 assets,
+        CollateralTracker collateralToken
+    ) external {
+        collateralToken.delegate(msg.sender, delegatee, assets);
+    }
+
     // mimics an internal Panoptic pool  revoke call onto the collateral tracker
     function revoke(
         address delegator,
@@ -202,6 +210,15 @@ contract PanopticPoolHarness is PanopticPool {
         CollateralTracker collateralToken
     ) external {
         collateralToken.revoke(delegator, delegatee, requestedAmount);
+    }
+
+    function refund(
+        address delegator,
+        address delegatee,
+        int256 requestedAmount,
+        CollateralTracker collateralToken
+    ) external {
+        collateralToken.refund(delegator, delegatee, requestedAmount);
     }
 
     function getTWAP() external returns (int24 twapTick) {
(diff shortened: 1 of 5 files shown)


test/foundry/core/CollateralTracker.t.sol
```diff
@@ -1319,7 +1336,48 @@ contract CollateralTrackerTest is Test, PositionUtils {
...
+        // check delegatee balance before
+        uint256 sharesBefore0 = collateralToken0.balanceOf(Bob);
+        uint256 sharesBefore1 = collateralToken1.balanceOf(Bob);
+
+        uint256 convertedShares = convertToShares(1_000_000_000, collateralToken0);
+
+        // invoke delegate transactions from the Panoptic pool
+        panopticPool.delegate(Alice, assets, collateralToken0);
+
+        panopticPool.delegate(Alice, assets, collateralToken1);
```
guil-lambert, line 1375: This doesn't look like it's testing the virtual shares delegation, shouldn't we use the two input call `delegate(address delegatee, uint256 assets)`?

guil-lambert (changes requested):
Add more tests explicitly testing flows for

- `delegate(address delegatee, uint256 assets)` 
- `refund(address delegatee, uint256 assets)`

test/foundry/core/CollateralTracker.t.sol
dyedm1: These functions can only be called by the Panoptic Pool, so we have a custom Panoptic Pool contract with the functions exposed publicly.

The first line calls ct0.delegate(Alice, assets)
The second line calls ct1.delegate(Alice, assets)

PR #22, opened 2024-01-10
feat: premium overhaul

dyedm1:
Under the current model, sellers can collect all of the premium they are owed (including Uniswap fees, long premium, and long premium spread) at any time, as long as there are enough settled premium tokens available in the pool (collected from Uniswap/paid by sellers across all chunks).

This is problematic for a few reasons:
1. If many sellers close around the same time, the rest of the sellers may be stuck waiting for more premium to settle into the pool to collect everything they are owed, even if no one bought their options.
2. If someone can sell and then buy an option (different accounts) that composes a large fraction of the active Uniswap liquidity, they can inflate and collect a large quantity of premium from the short option and then liquidate the long option without actually paying all the premium, which, depending on how the situation is handled, can result in the theft of funds from either PLPs or the general settled premium pool.

To rectify these issues, this PR replaces the current premium distribution system with a more robust model that distributes premium proportionally and fairly to sellers on a chunk-by-chunk basis according to the ratio of available (settled) tokens in the chunk to total tokens owed. (To balance out the UX hit for sellers, #26 will allow premium owed by eligible buyers to be collected early/before closing a sold option) 

To compute this ratio, we need two things (for each token):
1. The current amount of settled tokens available for withdrawal in the chunk
2. The total premium owed to all sellers in the chunk

This PR also adds a requirement (in `validate()`) that all legs in a `tokenId` must be of a unique chunk, which reduces a lot of technical complexity.  Theoretically, any tokenId that touches the same chunk twice can be replicated in a more gas-efficient tokenId that only has one touch, so this ban will have little-to-no UX impact.

For 1, we introduce a new storage variable `s_settledTokens` (LeftRight, left:token1, right:token0) that is updated when premium is collected from the Uniswap pool, paid by buyers, or collected by sellers. To retrieve the per-chunk premium collected from Uniswap, the SFPM `totalCollected` return value was changed to a per-leg array, `collectedByLeg`, which is then added to the corresponding chunk for each tokenId leg. 

For 2, we introduce another storage variable `s_grossPremiumLast` (LeftRight, left:token1, right:token0). When used as a `last` value with the current gross premium accumulator and `shortLiquidity`(`totalLiquidity`), the total premium owed to all sellers is the result. 

To accomplish that, we take advantage of the fact that *given a constant shortLiquidity*, `ΔgrossPremium*shortLiquidity` is equal to the total premium accrued to sellers. shortLiquidity only changes when sold options are minted or burned, so to keep `s_grossPremiumLast` accurate we algebraically solve for a new `grossPremiumLast` value that yields the same premium with the new shortLiquidity value as the *old* `grossPremiumLast` yielded from the *old* shortLiquidity value (computed from the position liquidity delta) after each mint, and do the same after each burn (with the premium target being corrected for the amount that was owed to the closed position).

Whenever sold options are closed and collateral is checked, the available premium (what is actually paid) is now defined by `available = min(owed, owed*s_settledTokens/(grossPremiumCurrent - grossPremiumLast)*shortLiquidity)`

Note: in a future PR, tokens owed by buyers over the protocol loss threshold in a liquidation *will not* be paid. If this is the result of premium manipulation, sellers in the chunk will still receive all the "real" premium owed to them, but if it occurs naturally due to a less-than-timely liquidation, sellers will not receive the premium that was exempted during the liquidation.


contracts/PanopticPool.sol
```diff
@@ -1649,75 +1746,356 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+                //
+                // (old)
+                // (C(T - R) - G)/(T - R) = Ln
+                uint256[2] memory grossNew;
+                unchecked {
+                    // if there's still liquidity, compute the new grossPremiumLast
+                    // otherwise, we just reset grossPremiumLast to the current grossPremium
+                    if (totalLiquidity != 0) {
+                        // get G
+                        grossNew[0] =
```
dyedm1: I need to give the rounding here some more thought; there's an edge case where G > C(T-R) because of terms being rounded differently. That's easy to patch (math.max(C(T-R) - G, 0)), but we need to make sure there's no griefing attack vector (erroneously decreasing grossPremiumLast, which increases the size of the owed premium beyond what can be settled). Usually rounding in favor of the protocol is accceptable and even prudent, but the calculations have to be tight here and not accumulate leaks of >1 accumulator point downward.

contracts/PanopticPool.sol
```diff
@@ -1723,4 +1639,315 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+            bytes32 chunkKey = keccak256(
+                abi.encodePacked(tokenId.strike(leg), tokenId.width(leg), tokenId.tokenType(leg))
+            );
+
+            // collected from Uniswap
+            uint256 settledTokens = s_settledTokens[chunkKey].add(collectedByLeg[leg]);
+
+            int256 legPremia = premiaByLeg[leg];
+
+            // (will be) paid by long legs
```
guil-lambert: This need to be negated for each slot:
```
            if (tokenId.isLong(leg) == 1) {
                settledTokens = settledTokens.add(uint256(0).toRightSlot(uint128(-legPremia.rightSlot())).toLeftSlot(-legPremia.leftSlot())));
                realizedPremia = realizedPremia.add(legPremia);
```

contracts/PanopticPool.sol
```diff
@@ -1649,72 +1590,47 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
                                 )
                             )
                         );
 
-                    if (isLong == 0) {
-                        premia = premia.add(legPremia);
-                    } else {
-                        premia = premia.sub(legPremia);
+                    if (isLong == 1) {
+                        premiaByLeg[leg] = -premiaByLeg[leg];
```
guil-lambert: This need to be negated for each slot:
`premiaByLeg[leg] = int256(0).sub(premiaByLeg[leg]);`

PR #23, opened 2024-01-11
Safer collateral requirements

guil-lambert:
Increased the base collateral requirement for ITM options.

1. Increase in collateral requirement, help reduce buffer between liquidation and protocol loss: https://www.desmos.com/calculator/xsu8podgaw
2. Increased utilization as tick deviation increases


dyedm1 (approved):
LGTM

PR #25, opened 2024-01-16
fix: consistent behavior on premium rollover

dyedm1:
It turns out that it is actually very possible to overflow the x64 premium accumulators in the SFPM by minting positions in low-liquidity areas on pools with a small full-range deployment, even with our general assumptions about the number of tokens passing through the system not exceeding 2**127-1 holding

In the SFPM, premium accumulator rollovers leak between slots (tokens) b/c unchecked addition of leftRights. Also, getAccountPremium (with the correct atTick passed) reverts on overflow instead of rolling over, which is not consistent with the actual behavior when the positions are closed. I added a new `addUnchecked` function that adds leftRights while allowing them to rollover without leaking into other slots, so we just use this whenever adding up premium in the SFPM to make the behavior consistent.

In the PanopticPool, rollovers are not handled at all and actually cause erroneous transfers from sellers to PLPs (negative fees):
```
// tough luck... PLPs just stole ~2**64 tokens per liquidity Alice had because of an overflow
// Alice can be frontrun if her transaction goes to a public mempool (or is otherwise anticipated),
// so the cost of the attack is just ~2**64 * active liquidity (shown here to be as low as 1 even with initial full-range!) 
// + fee to move price initially (if applicable)
// The solution is to wrap around the overflow once (so if the accumulator goes down, the fees are acc_current + (acc_max - acc_prev)
// If it overflows multiple times, we leave some fees unclaimed, but that's fine. Can't be exploited.
```
 
Run `test_success_PremiumRollover` on the previous version of the contracts to demonstrate the issues

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1649,7 +1649,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
                     positionSize,
                     s_tickSpacing
                 );
-                
+
                 uint256[2] memory premiumAccumulators;
                 {
                     uint256 tokenType = TokenId.tokenType(tokenId, leg);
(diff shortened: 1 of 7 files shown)


guil-lambert (approved):
Added a PR to remove the memory array and fix the stack too deep error, otherwise everything looks A-OK!

PR #26, opened 2024-01-24
feat: add `settleLongPremium` feature

dyedm1:
This PR adds a new `settleLongPremium` function that takes a list of owners, a list of tokenIds for each owner,  an exhaustive list of their positions (for collateral checking), and a chunk identity. The function settles all the premium owed by long chunks in the list of tokenIds for each owner matching the chunk identity passed in and updates the s_settledTokens/s_options accumulators accordingly. 

Any option seller operating in a chunk with a non-negligible liquidity utilization will want to call this function on some or all of the positions with long legs of their chunk before burning their position (probably using the `multicall` function, e.g `multicall([settleLongPremium(), settleLongPremium(), ...(however many chunks were utilized), burnOptions()])`.

Without this function, sellers (even small ones) are forced to wait for buyers to close their position, or force exercise them (expensive) if they want to receive the full amount of premium they are owed. This is because the premium (owed to sellers) collected at burn is now distributed proportionally based on the settled premium available.


contracts/PanopticPool.sol
```diff
@@ -1643,6 +1643,89 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+
+        if (tokenId.isLong(legIndex) == 0 || legIndex > 3) revert Errors.NotALongLeg();
+
+        // 0: currentTick, 1: medianTick, 2: tickSpacing
+        (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
+        int24 tickSpacing = s_tickSpacing;
+
+        uint256 accumulatedPremium;
+        {
+            (int24 tickLower, int24 tickUpper) = tokenId.asTicks(legIndex, tickSpacing);
```
guil-lambert: I'm using `tokenId.asTicks` here instead of the previous `PanopticMath.getTicks`. Do we need the `PanopticMath.getTicks` function anymore at all?

contracts/libraries/PanopticMath.sol
```diff
@@ -216,6 +217,40 @@ library PanopticMath {
         liquidityChunk = liquidityChunk.createChunk(tickLower, tickUpper, legLiquidity);
     }
 
+    /// @notice Extract the tick range specified by `strike` and `width` for the given `tickSpacing`, if valid.
+    /// @param strike the strike price of the option
+    /// @param width the width of the option
+    /// @param tickSpacing the tick spacing of the underlying Uniswap v3 pool
+    /// @return tickLower the lower tick of the liquidity chunk.
+    /// @return tickUpper the upper tick of the liquidity chunk.
+    function getTicks(
```
guil-lambert, line 226: See comment in `PanopticPool:1669`

contracts/PanopticPool.sol
dyedm1: We don't use it here anymore, but I think it's still better to keep it that way. The `asTicks` function calls `getTicks` instead of doing the tick calculations directly, which is a little cleaner IMO and lets you convert a strike/width into a tickLower/upper without needing to pack them into a tokenId. Also helps with #28.

contracts/PanopticPool.sol
```diff
@@ -1620,6 +1644,83 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+            accumulatedPremium = uint256(0).toRightSlot(uint128(premiumAccumulator0)).toLeftSlot(
+                uint128(premiumAccumulator1)
+            );
+
+            // update the premium accumulator for the long position to the latest value
+            // (the entire premia delta will be settled)
+            uint256 premiumAccumulatorsLast = s_options[owner][tokenId][legIndex];
+            s_options[owner][tokenId][legIndex] = accumulatedPremium;
+
+            accumulatedPremium = accumulatedPremium.sub(premiumAccumulatorsLast);
```
guil-lambert, line 1690: Right now, we don't collect any outstanding fees in the SFPM when we call `settleLongPremium`, I think we should. Ie. we need to actually collect in SFPM and update the SFPM accumulators and then read the accumulators in the PanopticPool.

It would make it a bit safe because the stored `s_options` values will be exactly the same as in the SFPM (which they aren't right now). The `atTick < type(int24).max` condition is only used when reading collateral requirement values, but ultimately all values are read from storage when it's actually time to exercise.

We may need an external function 

```
sfpm.collectAndSettle(
        address univ3pool,
        address owner,
        uint256 tokenType,
        int24 tickLower,
        int24 tickUpper,
        int24 atTick,
        uint256 isLong
    ) external { ...}
```
Which takes care of writing the latest collected to the sfpm accumulators

dyedm1: :thinking: as long as our calculated `feesBase` matches post-poke `feesBase`, it should be the same.
Even if it wasn't the same, we could just poke it and use the non-pending flag to guarantee the same result without actually having to implement a `collectAndSettle` function.

dyedm1: If you log out all the pending accumulators (during `settleLongPremium`) in the `test_success_settleLongPremium` function and compare to the updated accumulator values (post burn/touch) they are the same. 
 `_getAMMSwapFeesPerLiquidityCollected` should always yield the same result — it does the same math Uniswap does internally to construct the `feesBase` accumulators.

PR #27, opened 2024-01-24
feat: add `assertAccountValue` as optional collateral slippage check

dyedm1:
While you can't donate directly to the balance anymore, an attacker could still make a "donation" by paying the MEV tax or inducing a precision loss on withdrawal. Like other inflation attacks, this can lead to the first depositor getting frontrun for the value of their deposit.

To address this, we create 10^6 virtual shares corresponding to 1 asset by setting the `totalSupply` and `s_poolAssets` variable to 10^6 and 1 respectively during `startToken`. This makes the inflation attack uneconomical by increasing the capital requirement 1_000_000x, and provides some extra initial precision on shares, improving UX for users. There was an alternate solution based on user-provided slippage checks proposed below, but virtual shares are more versatile and do not require frontend intervention or extra gas (besides at initialization).

Also, the MEV tax calculations were improved. Previously, it was possible for the MEV tax, but not the assets amount,  to round down to zero. This property could be leveraged to avoid the MEV tax and provide JIT liquidity. Additionally, the MEV tax on `mint` was higher than on `deposit` because the same percentage was charged on the variable asset input in `mint` as the variable share output in `deposit`. Effectively, this results in being charged based on the final value of the shares you receive in `mint`, but *more* than the final value received in `deposit`.

To rectify this, all the terms of the `previewDeposit` and `previewMint` expressions were combined into a single appropriately-rounded `mulDiv` to preserve precision, and the `previewMint` expression was changed from `shares * totalAssets / totalShares * (1 + commissionFee)` to `shares * totalAssets / totalShares / (1 - commissionFee)` to match the MEV taxes calculated in `previewDeposit`.

~~Virtual shares don't work very well on a permissionless protocol because we don't know the decimals/value of the tokens (if you have ie GUSD your virtual share value will be huge and much of the yield will be diverted away from real users), so I added an `assertAccountValue` function that can be called with `multicall` if the pool state is vulnerable to frontrunning.~~

~~The good thing about this is there's only a (small) gas impact and only when actually needed, so the pattern for the vast majority of users is exactly the same.~~

~~For frontends, the pattern is:~~
~~1. Use a heuristic (is the user the first depositor/is the pool controlled by one large entity) to determine if a slippage check is needed~~
~~2. If that heuristic is true, do `multicall[deposit(...), assertAccountValue(minExpected, maxExpected)]), which will revert if the account value is outside the expected bounds. Generally for `deposit` you would want to set a minimum bound in particular because you receive a variable amount of shares, and vice versa for `mint` (protects against the MEV tax being unexpectedly high)~~
~~3. The inflation attack cannot be executed because the `assertAccountValue` call would revert~~

~~ERC4626 router contracts also do the same thing, but because the protection is never needed for most depositors, it makes more sense to have the functionality in the same contract so everything goes through the same interface.~~

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -425,6 +425,18 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     }
 
     /*//////////////////////////////////////////////////////////////
+                          SLIPPAGE PROTECTION
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Asserts that the asset value of msg.sender is at least `minAssets`.
+    /// @dev This function is used to protect against unexpected slippage from precision loss/rounding (e.g inflation attacks).
+    /// @dev The intended pattern is to call this function after a series of deposit/mint with `multicall`
+    /// @param minAssets The minimum amount of assets that the caller's account must be worth.
+    function assertAccountValue(uint256 minAssets) external view {
+        if (convertToAssets(balanceOf[msg.sender]) < minAssets) revert Errors.AccountValueTooLow();
+    }
+
+    /*//////////////////////////////////////////////////////////////
                      STANDARD ERC4626 INTERFACE
     //////////////////////////////////////////////////////////////*/
 
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -428,12 +428,14 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                           SLIPPAGE PROTECTION
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice Asserts that the asset value of msg.sender is at least `minAssets`.
+    /// @notice Asserts that the asset value of msg.sender is between `minAssets` and `maxAssets`.
     /// @dev This function is used to protect against unexpected slippage from precision loss/rounding (e.g inflation attacks).
-    /// @dev The intended pattern is to call this function after a series of deposit/mint with `multicall`
-    /// @param minAssets The minimum amount of assets that the caller's account must be worth.
-    function assertAccountValue(uint256 minAssets) external view {
-        if (convertToAssets(balanceOf[msg.sender]) < minAssets) revert Errors.AccountValueTooLow();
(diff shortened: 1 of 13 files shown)


contracts/CollateralTracker.sol
```diff
@@ -228,12 +229,21 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+        PanopticPool panopticPool,
+        uint256 virtualShares
     ) external {
         // fails if already initialized
         if (s_initialized) revert Errors.CollateralTokenAlreadyInitialized();
         s_initialized = true;
 
+        // the virtual shares function as a multiplier for the capital requirement to manipulate the pool price
+        // e.g if the virtual shares are 10**6, then the capital requirement to manipulate the price to 10**12 is 10**18
+        totalSupply = Math.max(virtualShares, 1);
```
guil-lambert: Why can't we hardcode this to `1e4` or `1e6`. That won't affect the low decimals tokens because `s_poolAssets` is set to `1`

contracts/CollateralTracker.sol
```diff
@@ -473,7 +495,11 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     function previewDeposit(uint256 assets) public view returns (uint256 shares) {
         // compute the MEV tax, which is equal to a single payment of the commissionRate BEFORE adding the funds
         unchecked {
-            shares = convertToShares(assets - (assets * (uint128(s_commissionFee))) / DECIMALS);
+            shares = Math.mulDiv(
```
guil-lambert, line 482: I changed the logic for previewDeposit, makes most sense when virtualShares are larger than 1e4

dyedm1: I agree, you get some free precision by doing it this way

contracts/CollateralTracker.sol
dyedm1: Oh, good point. The only thing that dilutes the yield is the assets...

Yeah I'm fine with this being fixed at 1e6 in that case, I'll update that

dyedm1:
Force-merging since we ended up agreeing on all changes here in the end and @guil-lambert is on vacation 🏖️

PR #28, opened 2024-01-25
feat: enable odd ts pool support

PeakyCryptos:
A tokenId which is the unit of creating positions is encoded with a strike parameter. This strike is used to describe the bounds of the position within Uniswap (lower tick, upper tick). The width parameter determines how many ticks wide the position is, depending on the tickspacing of the underlying pool.

i.e if the width is 2 and tickspacing is 10 and we have a strike of 9, this would be interpreted to have a one sided range of 
2 * 10 / 2 = 10. Positional bounds of (9 - 10, 9 + 10) -> (-1, 19). 

If we have a pool where the tickspacing is odd such as a 1 bps pool the problem is that there cannot be half ticks.
i.e width = 1 and tickspacing = 1 with a strike of 9. 1 * 1 / 2 = 0.5 (which is invalid).

To support odd tickspacing pools we do as follows:
If the product of the width and tick spacing is even, the tick range is (strike - range, strike + range); if it's odd, the tick range is (strike - range rounded down, strike + range rounded up). In cases where the product is even, both rangeUp and rangeDown will be the same.

Given the example of width = 1 and tickspacing = 1 with a strike of 9:
1 * 1 / 2 = 0.5 -> lower range is 0(round down) and upper range is 1 (round up), preserving the width of 1.
Giving a position bounds of (9 - 0, 9 + 1) -> (9, 10)
__

Other changes include using the more accurate getLiquidityForAmounts when querying the amounts moved in the opposite token when token type != asset. As well as using range up where collateral calculations are impacted to favor a higher requirement/exercise cost

### DIFF
--- a/contracts/libraries/PanopticMath.sol
+++ b/contracts/libraries/PanopticMath.sol
@@ -241,11 +241,6 @@ library PanopticMath {
 
             (tickLower, tickUpper) = (strike - rangeDown, strike + rangeUp);
 
-            console2.log("strike", strike);
-            console2.log("width", width);
-            console2.log("tickLower", tickLower);
-            console2.log("tickUpper", tickUpper);
-            console2.log("tickSpacing", tickSpacing);
             // Revert if the upper/lower ticks are not multiples of tickSpacing
             // Revert if the tick range extends from the strike outside of the valid tick range
             // These are invalid states, and would revert silently later in `univ3Pool.mint`
--- a/contracts/libraries/PanopticMath.sol
+++ b/contracts/libraries/PanopticMath.sol
@@ -12,7 +12,6 @@ import {Math} from "@libraries/Math.sol";
 import {LeftRight} from "@types/LeftRight.sol";
 import {LiquidityChunk} from "@types/LiquidityChunk.sol";
 import {TokenId} from "@types/TokenId.sol";
-import "forge-std/Test.sol";
 
 /// @title Compute general math quantities relevant to Panoptic and AMM pool management.
 /// @author Axicon Labs Limited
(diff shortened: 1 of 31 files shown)


contracts/libraries/PanopticMath.sol
```diff
@@ -234,10 +236,9 @@ library PanopticMath {
             int24 minTick = (Constants.MIN_V3POOL_TICK / tickSpacing) * tickSpacing;
             int24 maxTick = (Constants.MAX_V3POOL_TICK / tickSpacing) * tickSpacing;
 
-            // The width is from lower to upper tick, the one-sided range is from strike to upper/lower
-            int24 oneSidedRange = (width * tickSpacing) / 2;
+            (int24 rangeDown, int24 rangeUp) = PanopticMath.getRangesFromStrike(width, tickSpacing);
```
guil-lambert, line 239: Any chance we save some gas by looking at a condition that checks is `tickSpacing % 2 == 0` to compute `oneSidedRange = (width * tickSpacing) / 2` and goes to the `getRangesFromStrike` branch otherwise?

guil-lambert:
LGTM besides that one gas comment!

contracts/libraries/PanopticMath.sol
dyedm1: MOD (5) + EQ(3) + JUMPI(10) + another JUMP somewhere(8) = 26 vs. ADD + DIV + GT + MOD = 16 (assuming funcs are inlined, if not it's more).

It depends on the compiler profile, so I would say that we should err towards the branchless form since it's simpler and needs less bytecode (lower contract size). `getRangesFromStrike` is really just a couple cheap arithmetic ops on the upper tick and doesn't have any branches itself.

PR #29, opened 2024-02-05
feat: implement haircut on premium accumulated above protocol loss threshold

dyedm1:
We can't allow premium to be paid by liquidated accounts above the protocol loss threshold because buyers and sellers could collude to manipulate the premium (between themselves) and force the protocol to pay. Thus, if there is protocol loss, we haircut premium paid by the liquidatee until there is none left or the protocol loss threshold has been crossed back.

On average, this should not increase the gas cost for liquidations significantly because protocol loss is (or should be) a relatively long-tail scenario due to the generous liquidation buffer. However, it does have a worst-case gas increase of ~25600 if premium from every leg of all 32 tokens needs to be haircut. We may need to recalibrate the 32 positions limit, especially if we're launching on chains with lower gas limits.


dyedm1:
Q1: How do we deal with liquidatees paying themselves premium and then revoking it?
Q2: Does Ethereum (30m limit 15m target) need a new max positions limit?

dyedm1:
OK I figured out Q1! We will deal with Q2 in a separate PR later on - the gas costs in this PR are not final so there's no point in benchmarking now.

The solution for Q1 is to simply refrain from committing any long premium to the `s_settledTokens` accumulator in storage (but still include collected fees from Uniswap) during the burning process if a liquidation is occuring. For liquidatees who have long and short positions of the same chunk, the available premium ratio on their short positions could be lower and if so they would receive less premium. This does not affect the liquidation buffer because settled premium is not counted in solvency calculations.

In fact, the same effect could also be achieved by the liquidator if they modify the order of `positionIdList` such that all the short chunks are burned first. 

To accomplish this, we pass a flag down to `updateSettlementPostBurn` that indicates whether to include long premium in the settledTokens delta to commit.

We now commit the long premium paid (minus the haircut) during the `haircutPremia` mechanism instead of subtracting it from the premium committed during `updateSettlementPostBurn`

dyedm1:
We could probably also push the optimizer runs back up, but again will wait until after all this is merged in so we are working with final numbers.

dyedm1:
I'll resolve the conflicts once #22 is finalized - they are not significant.

PR #30, opened 2024-02-05
feat: implement more conservative collateral requirements for long risk-partnered positions

dyedm1:
Spreads - Changed the "base" collateral requirement to be that of the long leg. The long leg caps the maximum risk of the short leg, but the long leg itself still has premium risk. The new collateral requirement is the long leg requirement + 100% collateralized spread risk (token delta between the bounds of the "risky" portion of the spread)

Strangles/straddles - removed long strangles from the set of valid positions. Unlike short strangles, long strangles do not have any meaningful difference in risk (especially straddles - premium is accumulated at twice the rate) because the risk is derived only from the premium and not the intrinsic value of the position. 

Thanks to Robert from Numoen for asking an important question. This PR makes the answer 40x instead of infinity :)
image

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1696,6 +1696,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
 
         uint256 tokenType = tokenId.tokenType(index);
 
+        // compute the max loss of the spread
+
         // if asset is NOT the same as the tokenType, the required amount is simply the difference in notional values
         // ie. asset = 1, tokenType = 0:
         if (tokenId.asset(index) != tokenType) {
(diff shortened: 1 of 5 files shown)


contracts/CollateralTracker.sol
```diff
@@ -1789,6 +1733,13 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                     : ((notional - notionalP) * contracts) / notionalP;
             }
         }
+
+        // add the collateral requirement for the long leg as a base
+        spreadRequirement += _getRequiredCollateralAtUtilization(
```
guil-lambert: Do we want to make it: `spreadRequirement = max(longLegRequirement, movedPartnerRight - movedRight)`

Would make it worthwhile for a bit longer to create spreads vs. the naked option --ie. spreads up to 20% wide vs. spreads up to 10% wide assuming 10% base collateral requirement for the long option


--
Small nitpick: it is better for readability if the base requirement is added before line `1701`: 

ie. `base + extra` as opposed to `extra + base`.

guil-lambert (changes requested):
See comment, everything looks good otherwise

contracts/CollateralTracker.sol
dyedm1: >Do we want to make it: spreadRequirement = max(longLegRequirement, movedPartnerRight - movedRight)
Would make it worthwhile for a bit longer to create spreads vs. the naked option --ie. spreads up to 20% wide vs. spreads up to 10% wide assuming 10% base collateral requirement for the long option

I'm not sure that particular formula makes sense - a 5% wide spread would have the same collateral requirement as a 10% wide while being significantly more risky. 

There's no reason why we need to force users to 100% collateralize the risky portion of the spread though when the same standard is not applied to normal positions.

What if we just made it `spreadRequirement = longLegRequirement + (movedPartnerRight - movedRight)*SCR)`? That way the collateral requirement for risk-partnered spreads is asymptotic to the non-risk-partnered requirement as the strikes approach the ends of the tickrange and there's no weird calculus about whether to use the risk partner system for your spread or not.

dyedm1: >Small nitpick: it is better for readability if the base requirement is added before line 1701

Agreed

PR #31, opened 2024-02-16
feat: make `settleLongPremium` feature for 1 account+tokenId

guil-lambert:
Simplify the `settleLongPremium` to only accept a single owner and a single tokenId, build on https://github.com/panoptic-labs/panoptic-v1-core-private/pull/26


dyedm1 (changes requested):
While testing this I realized that we don't confirm
- the validity of the tokenId
- the validity of the leg index
- whether the user even owns it 

which could lead to some very dangerous exploits (!) 

We also don't do this for force exercises, but that executes a burn with the user's position balance (which will be 0 if any of the above are false, causing the SFPM to revert)

The cheapest way to rectify this issue is taking the tokenId at the end of the list (like what we do when minting an option), so I will do that. Otherwise, you have to either read the position balance again or check if the positionIdList contains an odd number of the tokenId, both of which are expensive.

PR #32, opened 2024-02-19
feat: Store tickSpacing in tokenId

guil-lambert:
It turns out that the only place where `tickSpacing` is needed is when calling `tokenId.asTicks` and in `CollateralTracker.exerciseCost`

Storing it in `PanopticPool` and `CollateralTracker` can be made unnecessary if the `tickSpacing` value is stored in the tokenId's `poolId`

I decided to store the tickSpacing in the lower 12 bits of the poolId. tickSpacing is int24, meaning that the max `tickSpacing` allowed in tokenId is `<4096`

I also had to refactor the getFinalPoolId a bit, but that's all very manageable.

This saves several `SLOAD` and bytecode+gas because tickSpacing doesn't need to be transmitted around anymore.

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -155,8 +155,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @dev Amount of assets moved from the Panoptic Pool to the AMM.
     uint128 internal s_inAMM;
 
-    /// @dev The tick spacing in the Uniswap pool.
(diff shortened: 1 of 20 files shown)


dyedm1:
We removed the IR compilation upstream, so ok that it fails.

dyedm1 (approved):
LGTM! (pending tests pass)

dyedm1:
I think the runner is stuck, they passed locally

PR #33, opened 2024-02-20
feat: round up commission fees

dyedm1:
This prevents avoidance of commission fees on low-decimal pools by making small positions, and by extension makes most premium rounding exploits unprofitable to execute.

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -824,31 +824,13 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // note: we HAVE to start with a negative number as the base exercise cost because when shifting a negative number right by n bits,
             // the result is rounded DOWN and NOT toward zero
             // this divergence is observed when n (the number of half ranges) is > 10 (ensuring the floor is not zero, but -1 = 1bps at that point)
-            uint256 fee = uint256(int256(s_exerciseCost >> maxNumRangesFromStrike)); // exponential decay of fee based on number of half ranges away from the price
+            int256 fee = (s_exerciseCost >> maxNumRangesFromStrike); // exponential decay of fee based on number of half ranges away from the price
 
             // store the exercise fees in the exerciseFees variable
             // rounding up, can use unsafe because numerator never overflows
             exerciseFees = exerciseFees
-                .toRightSlot(
-                    int128(
-                        uint128(
-                            Math.unsafeDivRoundingUp(
-                                uint256(uint128(longAmounts.rightSlot())) * fee,
-                                DECIMALS
-                            )
-                        )
-                    )
-                )
-                .toLeftSlot(
-                    int128(
-                        uint128(
-                            Math.unsafeDivRoundingUp(
-                                uint256(uint128(longAmounts.leftSlot())) * fee,
-                                DECIMALS
-                            )
-                        )
-                    )
-                );
+                .toRightSlot(int128((longAmounts.rightSlot() * fee) / DECIMALS_128))
+                .toLeftSlot(int128((longAmounts.leftSlot() * fee) / DECIMALS_128));
         }
     }
 
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -21,7 +21,6 @@ import {TickStateCallContext} from "@types/TickStateCallContext.sol";
 import {LeftRight} from "@types/LeftRight.sol";
 import {LiquidityChunk} from "@types/LiquidityChunk.sol";
 import {TokenId} from "@types/TokenId.sol";
-import "forge-std/Test.sol";
 
 /// @title Collateral Tracking System / Margin Accounting used in conjunction with a Panoptic Pool.
 /// @author Axicon Labs Limited
@@ -1142,11 +1141,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                     totalSupply,
                     totalAssets()
                 );
-                console2.log(
-                    "balanceOf, shares",
-                    balanceOf[tickStateCallContext.caller()],
-                    sharesToBurn
-                );
+
                 _burn(tickStateCallContext.caller(), sharesToBurn);
             } else if (tokenToPay < 0) {
                 // if user must receive tokens, mint them
@@ -1264,7 +1259,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                     s_ITMSpreadFee * uint256(Math.abs(intrinsicValue)),
                     DECIMALS
                 );
-                console2.log("swapCommission", swapCommission);
+
                 // set the exchanged amount to the sum of the intrinsic value and swapCommission
                 exchangedAmount = intrinsicValue + int256(swapCommission);
             }
@@ -1276,7 +1271,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                     DECIMALS
                 )
             );
-            console2.log("exchangedAmount", exchangedAmount);
         }
     }
 
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -21,6 +21,7 @@ import {TickStateCallContext} from "@types/TickStateCallContext.sol";
 import {LeftRight} from "@types/LeftRight.sol";
 import {LiquidityChunk} from "@types/LiquidityChunk.sol";
 import {TokenId} from "@types/TokenId.sol";
+import "forge-std/Test.sol";
 
 /// @title Collateral Tracking System / Margin Accounting used in conjunction with a Panoptic Pool.
 /// @author Axicon Labs Limited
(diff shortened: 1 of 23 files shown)


guil-lambert:
We need to round up to a bunch of other places too, especially when dealing with collateral requirement:

- ~~In `_getRequiredCollateralAtUtilization`: overestimate requirement for [short](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/67e1595d6e168fddc421fd97e7f3507cf597f9a2/contracts/CollateralTracker.sol#L1635) and [long](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/67e1595d6e168fddc421fd97e7f3507cf597f9a2/contracts/CollateralTracker.sol#L1647)~~ 
  - commit: 40e5152f62e59fabcfcdaff198dbb5b09ec3a8fd
- ~~In `_getRequiredCollateralSingleLegNoPartner`: overestimate requirement for for [c2](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/67e1595d6e168fddc421fd97e7f3507cf597f9a2/contracts/CollateralTracker.sol#L1550) and [c3](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/67e1595d6e168fddc421fd97e7f3507cf597f9a2/contracts/CollateralTracker.sol#L1558)~~ 
  - commit: 40e5152f62e59fabcfcdaff198dbb5b09ec3a8fd
- ~~In `_computeSpread`: overestimate requirement for the final calculation for [spreadRequirement](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/67e1595d6e168fddc421fd97e7f3507cf597f9a2/contracts/CollateralTracker.sol#L1722-L1723)~~
  - commit: 3438b3883afc784d75e5741bd647794d4c12b8d4
- ~~In `exerciseCost`: overestimate when computing the returned [exerciseFees](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/67e1595d6e168fddc421fd97e7f3507cf597f9a2/contracts/CollateralTracker.sol#L831-L832)~~
  - commit: 3438b3883afc784d75e5741bd647794d4c12b8d4
- ~~In `_getSolvencyBalances`: overestimate the required amounts by rounding up [here](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/67e1595d6e168fddc421fd97e7f3507cf597f9a2/contracts/PanopticPool.sol#L1387-L1389)~~
  - commit: 7caf9306a31def34867a14faa549967259552ac2
- ~~In `_checkSolvency`: also round up the required balances [here](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/67e1595d6e168fddc421fd97e7f3507cf597f9a2/contracts/PanopticPool.sol#L1366)~~
  - commit: 7caf9306a31def34867a14faa549967259552ac2

And also we should round up whenever we compute long leg premia:
- ~~I would find a way to use the `unsafeDivRoundingUp` when `isLong = 1` [here](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/67e1595d6e168fddc421fd97e7f3507cf597f9a2/contracts/PanopticPool.sol#L1662-L1696)~~
  - commit: 725395baf6891f67a5d1ef76a10eaa031b74543e

And whenever we update the accumulator for the owed premia:
- ~~In `_getPremiaDeltas`: use round up when computing `deltaPremiumOwed` [here](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/67e1595d6e168fddc421fd97e7f3507cf597f9a2/contracts/SemiFungiblePositionManager.sol#L1343-L1350)~~
  - commit: 4f933e18b65be3f99a5043ab6e00161bc3c44427
- I guess we should still round down for `deltaPremiumGross`, right?
  - added rounding up in _base and _owed calculations: 2ce066375e64099c246218714de33ba9d4a4b6a0

I'd rather be cautious and always overestimate whenever calculating requirements or amounts to be paid, and round down for balances and amounts to be received, but let me know if any of those don't make sense

NEW roundings:

- Rounding up in PanopticFactory when computing full-range liquidities: 4a29a3254ea47f778eaed7d3b3a86882aa725971
- Did a few optimizations and used `Math.unsafeDivRoundingUp` where appropriate: 34373bb0fc1b3fe5ca26841832ac565631ce3e5e and 50bec54590d5c517d2302133aa0ec44211457b8a

contracts/CollateralTracker.sol
```diff
@@ -831,9 +827,10 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             int256 fee = (s_exerciseCost >> maxNumRangesFromStrike); // exponential decay of fee based on number of half ranges away from the price
 
             // store the exercise fees in the exerciseFees variable
+            // rounding up, can use unsafe because numerator never overflows
```
dyedm1: I disagree with rounding up the exercise fees. As long as long positions cannot be exercised in-range, I don't think it's necessary that we ensure the exercise fees are nonzero. Worst case, a small long position gets exercised a little earlier, which is not really a problem (honestly, it's debatable whether we even need the exercise fee past the cliff, which is currently the position range).

All this does is incentivize option buyers to split up long positions to make their portfolio more expensive to exercise (on certain pools). Rounding is always a tradeoff between two non-ideal outcomes (which only sometimes have a security impact), and the latter is definitely less ideal.

(also, we did not actually round them up here, so the comment is wrong lol)

contracts/CollateralTracker.sol
```diff
@@ -1714,9 +1721,10 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                     contracts = movedRight;
                 }
                 // the required amount is the amount of contracts multiplied by (notional1 - notional2)/min(notional1, notional2)
+                // can use unsafe because numerator never overflows
```
dyedm1: fine, but realized comment on L1709-1710 is wrong/outdated

contracts/PanopticPool.sol
```diff
@@ -1363,7 +1363,8 @@ contract PanopticPool is ERC1155Holder, Multicall {
             Math.getSqrtRatioAtTick(atTick)
         );
 
-        return balanceCross > (thresholdCross * buffer) / 10_000;
+        // compare balance and required tokens, can use unsafe div because thresholdCross is always smaller than 2**192
```
dyedm1: btw the "unsafe div" doesn't have anything to do with over/underflow, it means that if you div by 0 it will return 0 instead of panicking. So this is actually checked and will revert if it overflows.

dyedm1: More accurately it is always smaller than 2**224 (* buffer which is like 13_333), so perhaps we should uncheck this.

contracts/PanopticPool.sol
```diff
@@ -1663,7 +1664,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
                         .toRightSlot(
                             int128(
                                 int256(
-                                    (uint256(
+                                    Math.mulDiv64(
```
dyedm1: I also don't agree with rounding up the long premium here (provided that we keep the `settleLongPremium` function). It does have a benefit (short option holders are guaranteed to receive the full amount of premium owed), but if the round can be triggered at any time (through `settleLongPremium`) option buyers can be stolen from by repeatedly accumulating ε premia and calling `settleLongPremium` on an option buyer to force them to pay 1 premia. I don't think that tradeoff makes sense. 

The effect on security if we *don't* round up is very minor — option buyers may receive *slightly* less premium *in the future*, but they retain anything they earned given the conditions they minted and can leave at any time. This is a small price to pay, given that if we do round up, any user who buys an option can have their entire account balance zeroed.

contracts/SemiFungiblePositionManager.sol
```diff
@@ -1340,13 +1341,15 @@ contract SemiFungiblePositionManager is ERC1155, Multicall {
...
                 {
                     // compute the owed premium (from Eqn 3)
-                    uint256 numerator = netLiquidity + (removedLiquidity / 2 ** VEGOID);
+                    // round up to prevent no vegoid term for small removedLiquidity
+                    uint256 numerator = netLiquidity +
+                        Math.unsafeDivRoundingUp(removedLiquidity, 2 ** VEGOID);
 
                     premium0X64_owed = uint128(
-                        Math.mulDiv(premium0X64_base, numerator, totalLiquidity)
+                        Math.mulDivRoundingUp(premium0X64_base, numerator, totalLiquidity)
```
dyedm1: See above - definitely not safe *if* we have the `settleLongPremium` function.

Even if not, I'm hesitant to round up here
- if we round up base, the gross premium may end up diverging from Uniswap 
- Rounding up owed here means that it occurs *every time the chunk is touched*, so there is still arbitrary opportunity to accumulate ε premia and mint a tiny short/long in the same chunk or otherwise touch it to increase the owed premium by L/2^64 (which could be more than 1 (!))

dyedm1 (commented):
I agree for the most part on the collateral changes, but have concerns regarding the exercise cost and premium rounding

dyedm1:
lgtm

PR #34, opened 2024-02-21
feat: remove internal median oracle and use Uniswap observations instead

dyedm1:
During our trading competitions, we've observed that users tend to interact with Panoptic infrequently (much less frequently than Uniswap), with sometimes hours-long periods between interactions. The internal median oracle can only be updated on Panoptic interactions, so the pricing data in there is at times very stale. It is also vulnerable to manipulation because it snapshots prices set in the current block.

We introduced the median oracle because our Uniswap TWAP oracle reads 20 observations plus some more during the search (to find the observation most closely corresponding to the secondsAgo we request), which ends up being ~300k gas (far too much to expect users to pay during a mint). 

At the time, we only used it to ensure that the position a user was minting did not make them insolvent. The consequence of a manipulated price there is rather low - a user could mint more options than their buying power could accommodate, but they would be immediately liquidated after with a low chance of protocol loss.

However, we now use the median oracle as the price for the post-burn collateral check, which is significantly more consequential. Provided the cost of swapping to manipulate the pool price is not too high, a motivated user could mint a short chunk from one account in a relatively empty liquidity area, then use another account to mint a short option with a high intrinsic value and buy the long option from the first account. After they set that up, they could swap to accumulate owed premium (to their first account) on the long option and make their second account insolvent. If they were liquidated at this point, all their premium would be haircut and there would be no protocol loss, which is why it is important that the second account not be allowed to burn the long option in its insolvent state. If they are able to manipulate the median oracle, however, they could move the price such that their short position becomes OTM, making their account solvent. If that happens, they can burn their long option, paying all the premium to their first account, then liquidate their short option at a steep loss (pocketing the intrinsic value of the short option + the liquidation bonus)).

The characteristics of the internal median oracle are as follows:
Manipulation Time: 4 minutes
Manipulation Cost: 8x swap fees from manipulated price/back
Gas: 1x cold sstore + warm sload + overhead (~5100 gas). Note that this cost is incurred on every interaction including collateral deposits/withdrawals 

This PR removes the internal median oracle and replaces it with a function that reads the last three observed prices from the Uniswap pool. The Uniswap pool observes the closing price of block n - 1 at block n, so in order to manipulate the price snapshot the manipulated price must be held cross-block.

Manipulation Time: Blocktime * 2 (24 seconds on Ethereum) (This can be increased significantly at no additional cost if we need)
Manipulation Cost: 4x swap fees from manipulated price/back + 2x slippage from holding manipulated price across block (Either from someone else closing the arb or bribe(some significant percentage paid to get the swapback transaction included - this will be much larger than the swap fees) (If some entity gets multiple blocks in a row they can bypass the large slippage cost in certain cases, we may want to increase the number of prices we retrieve depending on the chain)
Gas: 4x observations array reads (~2650 each, +2650 for each additional price) (~10600). Note that this is only on mints/burns since we don't need to poke on every interaction.

This improves the reliability of the oracle because the price snapshots are taken on each block containing a Uniswap swap for a pool, which happens much more frequently on average than interactions with Panoptic. It also adds a significant level of technical complexity and cost to manipulation since it can't be done atomically. If we need, we can increase the time required for manipulation at no cost (at the expense of a staler price output) We can also increase the amount of prices for ~2650 gas per mint/burn for each additional price, which adds to the manipulation time, and adds 2x swap fees + slippage to the manipulation cost. 

**NOTE: THIS PR DESCRIPTION IS SOMEWHAT OUTDATED BUT WE TALKED ABOUT IT


guil-lambert:
I do not agree with this proposed PR

dyedm1:
> I do not agree with this proposed PR

What specifically do you disagree with? Do you have an alternative proposal for making `medianTick` more difficult to manipulate?

BTW, tweet on why Uniswap observations are expensive to manipulate: https://x.com/orenyomtov/status/1765381415800074266?s=20

contracts/PanopticPool.sol
```diff
@@ -51,13 +49,11 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @param tokenId TokenId of the liquidated position.
     /// @param exerciseFee LeftRight encoding for the cost paid by the exercisor to force the exercise of the token.
     /// The token0 fee is in the right slot, and token1 fee is in the left slot.
-    /// @param tickAt Tick at which the position was exercised.
     event ForcedExercised(
         address indexed exercisor,
         address indexed user,
         uint256 indexed tokenId,
-        int256 exerciseFee,
-        int24 tickAt
```
guil-lambert, line 60: Don't we want to emit the tick for all events? I think it's better if we can know the tick where anything happened directly from the emitted event, although I can be OK with it if I understand why not

contracts/PanopticPool.sol
```diff
@@ -146,12 +138,37 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+
     // The maximum allowed delta between the currentTick and the Uniswap TWAP tick during a liquidation (~5% down, ~5.26% up)
     // Prevents manipulation of the currentTick to liquidate positions at a less favorable price
     int256 internal constant MAX_TWAP_DELTA_LIQUIDATION = 513;
 
-    // The minimum amount of time, in seconds, permitted between mini/median TWAP updates.
-    uint256 internal constant MEDIAN_PERIOD = 60;
+    /// The maximum allowed delta between the fast and slow oracle ticks at burn
+    /// Prevents burning positions during extremely volatile periods/price manipulation (to ensure the account is always solvent)
+    int256 internal constant MAX_TWAP_DELTA_BURN = 1115;
```
guil-lambert: Very low, would lead to reverts at the most inopportune time, etc. Set to at least 1800 (20% move) for now.

contracts/PanopticPool.sol
```diff
@@ -295,7 +310,19 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
 
         // Store the univ3Pool variable
-        s_univ3pool = IUniswapV3Pool(univ3pool);
+        s_univ3pool = IUniswapV3Pool(_univ3pool);
+
+        (, , uint16 observationIndex, uint16 observationCardinality, , , ) = IUniswapV3Pool(
+            _univ3pool
+        ).slot0();
+
+        int24 slowOracleTick = PanopticMath.computeMedianObservedPrice(
```
guil-lambert, line 316: What happens if the number of observations is less than `SLOW_ORACLE_CARDINALITY` --ie. when the uniswap pool is brand new? Does it return 0?

We could be liberal and use slot0 here and at least match the current state of the Uniswap pool.

contracts/PanopticPool.sol
```diff
@@ -324,6 +351,19 @@ contract PanopticPool is ERC1155Holder, Multicall {
                              QUERY HELPERS
     //////////////////////////////////////////////////////////////*/
 
+    /// @notice Reverts if current Uniswap price is not within the provided bounds.
+    /// @dev Can be used for composable slippage checks with `multicall` (such as for a force exercise or liquidation)
+    /// @dev Can also be used for more granular subtick precision on slippage checks
+    /// @param sqrtLowerBound The lower bound of the acceptable open interval for `currentSqrtPriceX96`
+    /// @param sqrtUpperBound The upper bound of the acceptable open interval for `currentSqrtPriceX96`
+    function assertPriceWithinBounds(uint160 sqrtLowerBound, uint160 sqrtUpperBound) external view {
+        (uint160 currentSqrtPriceX96, , , , , , ) = s_univ3pool.slot0();
```
guil-lambert, line 357: Combine this slot0 read with the one that fetches the cardinality eg. line 656.

ie. 
```
    function assertPriceWithinBounds(
        uint160 sqrtLowerBound, 
        uint160 sqrtUpperBound, 
        uint16 observationIndex, 
        uint16 observationCardinality) external view {

        (uint160 currentSqrtPriceX96, , uint16 observationIndex, uint16 observationCardinality, , , ) = s_univ3pool.slot0();
```

contracts/PanopticPool.sol
```diff
@@ -645,23 +713,19 @@ contract PanopticPool is ERC1155Holder, Multicall {
             .toLeftSlot(poolUtilizations)
             .toRightSlot(positionSize);
 
+        (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
```
guil-lambert: This duplicated the slot0 reads in mintTokenizedPosition, can't we bring back `(uint128 poolUtilizations, **int24 newTick**) = _mintInSFPMAndUpdateCollateral(` to avoid the extra gas?

contracts/PanopticPool.sol
dyedm1: I didn't think it was important to expend the extra effort to expose the tick from our end when you can easily query the tick from the last swap event when indexing. Generally it's not necessary to emit something unless there's a state change on your end. Fun fact: Even the Uniswap mint/burn events don't emit the tick. 

Although now that I'm thinking of it, we emit the SFPM mint/burn events before creating the position but the PanopticPool events are emitted afterward. It would be a little trickier to get the PanopticPool events because the state at the time the event is emitted doesn't necessarily match the token distribution in the position. We might want emit the PanopticPool events *first*.

The liquidation tickAt is also kind of nonsensical because it encompasses multiple positions, and the forceExercise one was just redundant (we emit the option mint/burn events with the tick irrespective of what function the option was burnt through), so not sure why we were keeping those.

contracts/PanopticPool.sol
```diff
@@ -679,25 +743,27 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+            tickLimitHigh
+        );
 
         // update premium settlement info
         _updateSettlementPostMint(tokenId, collectedByLeg, positionSize);
 
         // pay commission based on total moved amount (long + short)
         // write data about inAMM in collateralBase
         poolUtilizations = _payCommissionAndWriteData(
-            tickStateCallContext.updateCurrentTick(newTick),
```
guil-lambert, line 694: Not sure about this, it should be newTick and not slow_tick in `_payCommissionAndWriteData`. Here's why:

Immediately before a mint, the price is pushed way down using a flash loan. In that case, the price is 10 while the SLOW and FAST oracles both show 3500. 

Hence, since: `(fastOracleTick - slowOracleTick) < TICK_DEVIATION`, then the utilization is NOT `10001` and the position can be minted with ~20% collateral requirement only. 

The user deposits 0.201 ETH as collateral and mints 1 call at the 15 strike. The `_checkSolvency` uses FAST_ORACLE as the reference, so the user can mint that option with the 20% collateral requirement.

The attacker moves the price back, buys 1 ETH for 15, netting 2788 in profits: Sell 1000 ETH, sell 1 ETH call @ 15, buy back 1 ETH on the up-swap, the 1 ETH call gets liquidated, lose 0.2ETH. Net profit = 0.8*(3500-15) = 2788

IF the requirement was `10001`, then the user could only mint fully collateralized calls, meaning they buy they can only buy 0.2 ETH during the swap step, making it impossible to profit: Sell 1000 ETH, sell 0.2 ETH call @ 15, buy back 0.2 ETH on the up-swap, the 0.2 ETH call gets liquidated, lose 0.2ETH, make 0.2ETH. Net profit = 0

contracts/PanopticPool.sol
dyedm1: No it just reverts. I did this to avoid setting the medianTick to a weird value if the initial price was wrong.

I suppose using slot0 allows you to create a pool, pay for the observations, and wait for it to fill, so maybe we should use that instead for now. You would be able to do the same in Uniswap mode, so perhaps it's a little strict to do it this way.

contracts/PanopticPool.sol
dyedm1: Not sure what you mean by this? The `assertPriceWithinBounds` function is only supposed to be used at the top-level through `multicall` in liquidations/force exercises.

contracts/PanopticPool.sol
```diff
@@ -902,6 +951,59 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
     }
 
+    /// @notice Validates an account's state after burning an option, ensuring their solvency and checking the validity of the price oracles.
+    /// @param user The account to validate.
+    /// @param newPositionIdList The new positionIdList without the token(s) being burnt.
+    function _validateSolvencyBurn(
+        address user,
+        uint256[] calldata newPositionIdList
+    ) internal view {
+        IUniswapV3Pool _univ3pool = s_univ3pool;
```
guil-lambert, line 905: Do not re-fetch, use values from `_getSlippageLimits` or from `(collectedByLeg, totalSwapped, newTick) = SFPM.burnTokenizedPosition(`

contracts/PanopticPool.sol
```diff
@@ -902,6 +951,59 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+        (
+            ,
+            int24 currentTick,
+            uint16 observationIndex,
+            uint16 observationCardinality,
+            ,
+            ,
+
+        ) = _univ3pool.slot0();
+        int24 fastOracleTick = PanopticMath.computeMedianObservedPrice(
```
guil-lambert, line 915: Can't you re-use the same logic as during the oracle calculation at mint? ie. put everything in a function that outputs `(slowOracleTick, fastOracleTick) = getOracleTicks` or something and also call it when setting up `tickStateCallContext` on lines ~650

contracts/PanopticPool.sol
```diff
@@ -978,8 +1060,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
             );
             paidAmounts = paidAmounts.toLeftSlot(paid1);
         }
-
-        return (realizedPremia, premiaByLeg, newTick, paidAmounts);
```
guil-lambert, line 982: I'm confused why sometimes we explicitly return the variables and sometimes we don't --even though they are initialized in the `) internal returns (` part of the function initialization in each case.

contracts/PanopticPool.sol
```diff
@@ -1061,14 +1141,16 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-            (, finalTick, netExchanged, premiasByLeg) = _burnAllOptionsFrom(
+            (netExchanged, premiasByLeg) = _burnAllOptionsFrom(
                 liquidatee,
                 Constants.MIN_V3POOL_TICK,
                 Constants.MAX_V3POOL_TICK,
                 DONOT_COMMIT_LONG_SETTLED,
                 positionIdList
             );
 
+            (, finalTick, , , , , ) = s_univ3pool.slot0();
```
guil-lambert, line 1103: Do no call again

guil-lambert: Or maybe we do? I don't care as much since this is in `liquidate` anyways

contracts/CollateralTracker.sol
```diff
@@ -1057,12 +1045,13 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
-            // get the current Panoptic pool utilization
-            // Check if current tick is too far away from median, set to utilization to 10,001 if it is
-            int24 currentTick = tickStateCallContext.currentTick();
-            int24 medianTick = tickStateCallContext.medianTick();
-            // if the distance between the current and mini-median tick is more than the accepted tick deviation, default to 100% collateral requirement
-            if (Math.abs(currentTick - medianTick) > int256(TICK_DEVIATION)) {
+            // if the distance between the current and median tick is more than the accepted tick deviation, default to 100% collateral requirement
+            if (
+                Math.abs(
+                    int256(tickStateCallContext.fastOracleTick()) -
```
guil-lambert: Look at my comment below in `PanopticPool.sol`, no using `finalTick` here is dangerous

contracts/PanopticPool.sol
dyedm1: Not sure this is true :thinking: 
If you mint 1 call at the 15 strike, you borrow 1 ETH and move it to the Uniswap pool. Then on the way back you sell that ETH at 15 USDC/ETH to the attacker, which is not great...

But the collateral requirement of your call is not 20%. In fact, it would only be 20% if we used the `finalTick` as our oracle price. The collateral requirement is 20% (0.201 ETH) plus the intrinisic value at the `finalTick` (3500). At a price of 3500, we owe 1 ETH to PLPs but our position only has 15 USDC in it, so the intrinsic value is (3500-15=3485). Final CR = 3485 + 0.2*3500=4185 (before mint buffer etc). If you mint the position with the required collateral requirement, the attacker's account isn't liquidated when the price moves back.

contracts/PanopticPool.sol
dyedm1: Hmm yeah we can, it's only necessary for mints though.

For context I originally removed this when I was switching away from using `finalTick` as an oracle tick, but had to add it back at the end.

contracts/PanopticPool.sol
dyedm1: > Or maybe we do? I don't care as much since this is in liquidate anyways

Yeah I figured there was no point in that gas optimization since we don't need it in `burnOptions` anyway. Liquidations are very infrequent and very expensive, and this is only once per position list...

contracts/PanopticPool.sol
dyedm1: The difference is purely semantic; the variables are automatically returned at the end either way.
We haven't been enforcing a consistent styleguide rule regarding this, I'm actually doing a bunch of semantics/consistency-related tweaks in another branch before we submit to C4 and this is one of the things I'm trying to standardize.

contracts/PanopticPool.sol
dyedm1: Maybe, do we also want to poke at burn?

contracts/PanopticPool.sol
dyedm1: Not sure, but if you're talking about the univ3pool value here I'd rather not cache it because the gas savings are minimal once it's warm and it makes the code more complex by having to pass around that value (e.g. we don't cache the SFPM or that value from storage either) and eats into the savings. Slot0 values are more important to cache because there's a lot more overhead to actually retrieve them, but internal ones are cheap.

PR #47, opened 2024-04-15
fix: sign flip in `settleLongPremium` causes payments to flow the wrong way

dyedm1:
The wrong convention was used for the `realizedPremia`  values passed into `exercise` during `settleLongPremium` -- they are never flipped negative (as they are for long positions during the `burnOptions` flow), causing premium payments to be credited instead of debited from buyers.

We have an extensive test covering this logic, but a separate issue prevented it from catching this. Token delta calculations for both buyer *and* seller accounts used the same format: `ctN.convertToAssets(ctN.balanceOf(account)) - assetsBeforeN`, but those terms need to be reversed for buyers because their deltas have the opposite sign.

So, effectively, this issue was caused by sign flips present in both the actual logic and the tests. In the future, we need to conduct more careful review of the assertions we make in tests. The code here is not visibly incorrect (that convention could have been easily been the reverse), so in situations like these we have to rely on the test assertions being accurate.

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1634,8 +1634,8 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 .toLeftSlot(int128(int256((accumulatedPremium.leftSlot() * liquidity) / 2 ** 64)));
 
             // deduct the paid premium tokens from the owner's balance and add them to the cumulative settled token delta
-            s_collateralToken0.exercise(owner, 0, 0, 0, realizedPremia.rightSlot());
-            s_collateralToken1.exercise(owner, 0, 0, 0, realizedPremia.leftSlot());
+            s_collateralToken0.exercise(owner, 0, 0, 0, -realizedPremia.rightSlot());
+            s_collateralToken1.exercise(owner, 0, 0, 0, -realizedPremia.leftSlot());
 
             bytes32 chunkKey = keccak256(
                 abi.encodePacked(
--- a/test/foundry/core/Misc.t.sol
+++ b/test/foundry/core/Misc.t.sol
@@ -900,13 +900,13 @@ contract Misctest is Test, PositionUtils {
         }
 
         assertEq(
-            ct0.convertToAssets(ct0.balanceOf(Buyers[0])) - assetsBefore0,
+            assetsBefore0 - ct0.convertToAssets(ct0.balanceOf(Buyers[0])),
             33_342,
             "Incorrect Buyer 1 1st Collect 0"
         );
 
         assertEq(
-            ct1.convertToAssets(ct1.balanceOf(Buyers[0])) - assetsBefore1,
+            assetsBefore1 - ct1.convertToAssets(ct1.balanceOf(Buyers[0])),
             33_343_452,
             "Incorrect Buyer 1 1st Collect 1"
         );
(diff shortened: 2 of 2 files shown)


PeakyCryptos (approved):

contracts/PanopticPool.sol
```diff
@@ -1634,8 +1634,8 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 .toLeftSlot(int128(int256((accumulatedPremium.leftSlot() * liquidity) / 2 ** 64)));
```
flanagansteve, line 1634: I was reading the docstring on this method and saw:

> Called by sellers on buyers of their chunk to increase the available premium for withdrawal (before closing their position).

Out of curiosity, have we considered a convenience method that would first call this method and then call withdraw?

Or is that unnecessary because of our multicall functionality? I suppose there'd be a slight gas saving in providing the convenience method, though

flanagansteve, line 1634: I'm trying to trace how `_validatePositionList` works at the top of this function. I think it will fail if the user only passes a subset in of their open positions. Is this OK? Would users ever desire to only collect premia on some but not all of their positions?

Also, is there a term i can search to understand this validate-list-via-hashes and XORs thing we're doing in `_validatePositionList`?

flanagansteve, line 1634: Above, we seem to say:

```
        if (tokenId.isLong(legIndex) == 0 || legIndex > 3) revert Errors.NotALongLeg();
```

Why would the leg index being greater than 3 cause this error?

flanagansteve, line 1634: Why do we define liquidity above only using  the rightslot, not the leftslot?

```
        uint256 liquidity = PanopticMath
            .getLiquidityChunk(tokenId, legIndex, s_positionBalance[owner][tokenId].rightSlot())
            .liquidity();
```

flanagansteve (approved):
Your changes make sense to me. I left some questions that I can also investigate on my own, but figured I would ask here if they're easy to answer, for posterity and to accelerate my exploration here

contracts/PanopticPool.sol
dyedm1: I already implemented automatic `multicall`-based support for this in the UI. There might be a way to reduce the gas cost slightly, but the tradeoff in complexity/attack surface doesn't make sense. In fact, this function used to accept multiple positions, but we simplified it to only accept one (in a slightly more consequential gas/complexity tradeoff). Sellers often need to call this function multiple times for different owners/tokenIds either way.

dyedm1: >I'm trying to trace how _validatePositionList works at the top of this function. I think it will fail if the user only passes a subset in of their open positions. Is this OK? Would users ever desire to only collect premia on some but not all of their positions?

You only settle one position at a time (the last index), but all the other positions must be included in the input for a collateral check.

>Also, is there a term i can search to understand this validate-list-via-hashes and XORs thing we're doing in _validatePositionList?

Nothing specific OTOH, although it shares some underlying concepts with cuckoo/xor filters (the difference being this is an absolute inclusion proof). The key point is that XOR is symmetrical, so XORing numbers together always yields the same result regardless of the order you put them in (and a number can be removed from a compounded XOR result by simply XORing it again). Using this principle, you can create a robust mutable "fingerprint" of a set of numbers (w/ simple adds/removes) that can be easily verified, without knowing the original order.

dyedm1: Legs greater than 3 do not exist, and thus are not long.

dyedm1: The position size is only in the right slot -- other slot is used to pack the pool utilizations at mint.

PR #48, opened 2024-04-18
fix: remove deprecated `convertNotional` function

dyedm1:


PeakyCryptos (approved):

flanagansteve (approved):
✅ Grepped for `convertNotional` and you got all instances
✅ It doesn't appear that the documentation for convertNotional contained any info one couldn't find elsewhere

PR #49, opened 2024-04-18
feat: expose oracle tick calculations in `PanopticHelper`

dyedm1:


PeakyCryptos (approved):

periphery/PanopticHelper.sol
```diff
@@ -72,6 +73,72 @@ contract PanopticHelper {
...
+
+    /// @notice Returns the median of the last `cardinality` average prices over `period` observations from `univ3pool`.
+    /// @dev Used when we need a manipulation-resistant TWAP price.
+    /// @dev Uniswap observations snapshot the closing price of the last block before the first interaction of a given block.
+    /// @dev The maximum frequency of observations is 1 per block, but there is no guarantee that the pool will be observed at every block.
+    /// @dev Each period has a minimum length of blocktime * period, but may be longer if the Uniswap pool is relatively inactive.
+    /// @dev The final price used in the array (of length `cardinality`) is the average of all observations comprising `period` (which is itself a number of observations).
+    /// @dev Thus, the minimum total time window is `cardinality` * `period` * `blocktime`.
+    /// @param univ3pool The Uniswap pool to get the median observation from
+    /// @param cardinality The number of `periods` to in the median price array, should be odd.
```
flanagansteve, line 88: Is `cardinality` the best word here? I know its accurate, but `observations` or `numObservations` might be more digestible

edit: ah, i notice you're matching the existing arg named `cardinality` in `computeMedianObservedPrice`. could be a pain to change, then, but wanted to throw it out there

periphery/PanopticHelper.sol
```diff
@@ -72,6 +73,72 @@ contract PanopticHelper {
...
+    /// @param period The number of observations to average to compute one entry in the median price array
+    /// @return The median of `cardinality` observations spaced by `period` in the Uniswap pool
+    function computeMedianObservedPrice(
+        IUniswapV3Pool univ3pool,
+        uint256 cardinality,
```
flanagansteve, line 93: Since `cardinality` can be any number in this method, the comment in `computeMedianObservedPrice` that we call is now inaccurate (I think):

```
// get the median of the 3 calculated ticks
            return int24(Math.sort(ticks)[cardinality / 2]);
```

should be `get the median of the calculated ticks`

periphery/PanopticHelper.sol
flanagansteve, line 88: I don't think `computeMedianObservedPrice` enforces a check that this value will be odd. Would it be worth enforcing that in this method? I suppose the counterargument is that consumers of this helper contract might prefer the gas savings and handling any oddness checks on their end

flanagansteve (approved):

periphery/PanopticHelper.sol
dyedm1: Uniswap also uses the word `cardinality` in their implementation, so I thought it made sense to keep that convention.

periphery/PanopticHelper.sol
dyedm1: Yeah that comment is just inaccurate in general -- we call it with `cardinality` values other than 3 elsewhere. There are a few anachronistic comments like this (the original version of this method was hardcoded to return the fast oracle price) that I'll get in a final docs/overall review pass.

periphery/PanopticHelper.sol
dyedm1: Don't think so. This function is intended to help our interface retrieve the fast/slow oracle prices our smart contracts use; the parameters will be the same values hardcoded into the contracts at launch.

PR #50, opened 2024-04-19
Uniswap Migrator

flanagansteve:
- [ ✅ ] first pass at the contract
- [  ] nat spec on contract not there yet
- [  ] not tested - spent a while setting up the test, understanding how we init world, but got into trouble setting up the dependencies properly and have to go sadly

Hoping someone can take this across the finish line.

Left some TODOs about implementation choices too

### DIFF
--- a/periphery/PeripheryErrors.sol
+++ b/periphery/PeripheryErrors.sol
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+/// @title Custom Errors library.
+/// @author Axicon Labs Limited
+/// @notice Contains all custom error messages used in Panoptic's periphery.
+library PeripheryErrors {
+    /// @notice Caller does not own the NFPM token being migrated
+    error UnauthorizedMigration();
+}
--- a/periphery/PeripheryErrors.sol
+++ b/periphery/PeripheryErrors.sol
@@ -3,7 +3,7 @@ pragma solidity ^0.8.0;
 
 /// @title Custom Errors library.
 /// @author Axicon Labs Limited
-/// @notice Contains all custom error messages used in Panoptic's periphery.
+/// @notice Contains all custom error messages used in Panoptic's periphery contracts.
 library PeripheryErrors {
     /// @notice Caller does not own the NFPM token being migrated
     error UnauthorizedMigration();
--- a/periphery/UniswapMigrator.sol
+++ b/periphery/UniswapMigrator.sol
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: UNLICENSED
+// SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.18;
 
 // Interfaces
--- a/periphery/UniswapMigrator.sol
+++ b/periphery/UniswapMigrator.sol
@@ -12,12 +12,15 @@ import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 /// @title Facilitates the migration from Uniswap LPing to PLPing.
 contract UniswapMigrator {
-
     PanopticFactory public panopticFactory;
     INonfungiblePositionManager public univ3NFPM;
     IUniswapV3Factory public univ3Factory;
 
-    constructor(PanopticFactory panopticFactory_, INonfungiblePositionManager univ3NFPM_, IUniswapV3Factory univ3Factory_) {
+    constructor(
+        PanopticFactory panopticFactory_,
+        INonfungiblePositionManager univ3NFPM_,
+        IUniswapV3Factory univ3Factory_
+    ) {
         panopticFactory = panopticFactory_;
         univ3NFPM = univ3NFPM_;
         univ3Factory = univ3Factory_;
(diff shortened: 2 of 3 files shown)


dyedm1:
I refactored the migrator a bit and wrote a test suite for it. Few things to note:
- Generally, everything that can be natspecc'd should be (incl. functions, state variables (even non-public ones), structs, etc.)
- We always capitalize natspec, use periods on function/contract notice/dev/title tags, and no periods elsewhere
- Note that dev tags are italic and notice/param tags are not. Natspec also supports markdown -- (``) is nice for referencing params/vars. 
- Try out `forge doc --serve` if you want to see what the natspec output looks like. We are still in the process of enforcing a consistent styleguide across all the contracts, but new code should follow that natspec convention at the very least.
- You missed an important check -- the caller MUST own all of the tokenIds they migrate. If that assertion is not done, an attacker could steal all LP tokens approved to the migrator (!)
- Using `liquidity` for the amounts to collect is nonsensical (`liquidity` is not an amount of tokens) and could cause some tokens not to be collected from the position (forcing the user to figure that out and go and collect the tokens themselves + potentially causing the migration to fail if not enough collateral is migrated)
- Burning the token doesn't have any purpose aside from vanity and adds extra gas costs
- I decided to do away with the non-amount0min function signature because we shouldn't be encouraging people to forego slippage checks
- I also had the interface specify the collateral vaults they deposit into -- this is the most direct/intuitive/gas-efficient way
- We've been doing this recently, but int he future we will be taking a barbell approach to testing complex systems -- primarily writing unit tests after each PR and developing an Echidna/stateful fuzzing suite on the final product. Our current stateless fuzz test suite is comprehensive, but only exists due to poor management decisions and has wasted an immense quantity of the team's time across multiple dimensions -- writing them takes much longer than unit tests, feedback loop is slow due to long runtime (even with a custom CI runner colocated with a mainnet node running on a server costing $400+/mo), and debugging the tests is a task of nightmarish complexity (and most of the time the hours spent digging into the rabbit holes created by assertion failures in the tests reveal an inconsequential precision difference or a problem otherwise caused by the testing methodology instead of an actual issue). With stateless fuzz tests you get the worst of both worlds -- the difficulty of debugging small discrepancies in fuzzed assertions (multiplied by the number of test configurations) combined with the limitations of the team's imagination (stateless fuzz test are just unit tests with random values -- ultimately the team still chooses how exactly the test is run. An issue being present on only a particular value or range of values is far less common than an issue present on a certain branch that could be triggered by a unit test or an issue that can only be triggered by a certain sequence of calls + multiple agents).

PeakyCryptos (approved):

flanagansteve:
> I refactored the migrator a bit and wrote a test suite for it. Few things to note:
> 
>     * Generally, everything that can be natspecc'd should be (incl. functions, state variables (even non-public ones), structs, etc.)
> 
>     * We always capitalize natspec, use periods on function/contract notice/dev/title tags, and no periods elsewhere
> 
>     * Note that dev tags are italic and notice/param tags are not. Natspec also supports markdown -- (``) is nice for referencing params/vars.
> 
>     * Try out `forge doc --serve` if you want to see what the natspec output looks like. We are still in the process of enforcing a consistent styleguide across all the contracts, but new code should follow that natspec convention at the very least.
> 
>     * You missed an important check -- the caller MUST own all of the tokenIds they migrate. If that assertion is not done, an attacker could steal all LP tokens approved to the migrator (!)
> 
>     * Using `liquidity` for the amounts to collect is nonsensical (`liquidity` is not an amount of tokens) and could cause some tokens not to be collected from the position (forcing the user to figure that out and go and collect the tokens themselves + potentially causing the migration to fail if not enough collateral is migrated)
> 
>     * Burning the token doesn't have any purpose aside from vanity and adds extra gas costs
> 
>     * I decided to do away with the non-amount0min function signature because we shouldn't be encouraging people to forego slippage checks
> 
>     * I also had the interface specify the collateral vaults they deposit into -- this is the most direct/intuitive/gas-efficient way
> 
>     * We've been doing this recently, but int he future we will be taking a barbell approach to testing complex systems -- primarily writing unit tests after each PR and developing an Echidna/stateful fuzzing suite on the final product. Our current stateless fuzz test suite is comprehensive, but only exists due to poor management decisions and has wasted an immense quantity of the team's time across multiple dimensions -- writing them takes much longer than unit tests, feedback loop is slow due to long runtime (even with a custom CI runner colocated with a mainnet node running on a server costing $400+/mo), and debugging the tests is a task of nightmarish complexity (and most of the time the hours spent digging into the rabbit holes created by assertion failures in the tests reveal an inconsequential precision difference or a problem otherwise caused by the testing methodology instead of an actual issue). With stateless fuzz tests you get the worst of both worlds -- the difficulty of debugging small discrepancies in fuzzed assertions (multiplied by the number of test configurations) combined with the limitations of the team's imagination (stateless fuzz test are just unit tests with random values -- ultimately the team still chooses how exactly the test is run. An issue being present on only a particular value or range of values is far less common than an issue present on a certain branch that could be triggered by a unit test or an issue that can only be triggered by a certain sequence of calls + multiple agents).

Replied on Slack - agree with all this and thank you for pointing them out - should consider adding linter rules to enforce these commenting practices

PR #52, opened 2024-05-01
fix: update settlement at burn when legPremia = 0

dyedm1:
The`s_grossPremiumLast` update in `_updateSettlementPostBurn` is gated by a `legPremia != 0` check.  While this may appear logical because that check necessarily implies there is no change in the overall amount of premium owed to sellers for a given chunk, `s_grossPremiumLast` needs to be updated regardless because it is a *per-liquidity* accumulator. When sellers close their positions, the total premium owed according to the previous `s_grossPremiumLast` decreases by `currentGross - grossPremiumLast * closedLiq`. This occurs regardless of the the amount of premium owed to the seller, because the total amount of liquidity sold decreases. In the `_updateSettlementPostBurn` calculation, the premium owed to a seller is subtracted from the total premium, and `s_grossPremiumLast` is adjusted to reflect the remaining amounts of premium and liquidity in that chunk. In the case where a seller is not owed any premium, that calculation will still result in a lower `s_grossPremiumLast` to account for the liquidity removed by that seller at burn (which mirrors the increase in `s_grossPremiumLast` at mint). Thus, we cannot skip that calculation --  if positions are minted and burned in a loop and earn no premium, the calculated premium total will gradually decrease (`s_grossPremiumLast` increases at mint with no corresponding increase when the position is burned), breaking the invariant that sellers can only withdraw `max(owedSeller, owedSeller*settledTotal/owedTotal)` tokens of premium.

To rectify this, we simply remove that `legPremia != 0` check. There are no significant gas consequences/side effects to doing this.

This was not caught by our existing test suite because the only way to observe the effects of that internal state change is by minting and burning multiple positions with no premium, then attempting to withdraw from that chunk while a long position owes a significant portion of the premium. Our Echidna tests should be able to catch this soon -- the required sequence of transactions can be generated rather easily and we can check the invariants directly on the internal settlement state.

### DIFF
--- a/test/foundry/core/Misc.t.sol
+++ b/test/foundry/core/Misc.t.sol
@@ -480,6 +480,105 @@ contract Misctest is Test, PositionUtils {
         pp.burnOptions($posIdList[0], new TokenId[](0), 0, 0);
     }
 
+    // total owed/grossPremiumLast should not change when positions with 0 premia are minted/burnt
+    function test_settledtracking_premia0() public {
+        swapperc = new SwapperC();
+        vm.startPrank(Swapper);
+        token0.mint(Swapper, type(uint128).max);
+        token1.mint(Swapper, type(uint128).max);
+        token0.approve(address(swapperc), type(uint128).max);
+        token1.approve(address(swapperc), type(uint128).max);
+
+        // mint OTM position
+        $posIdList.push(
+            TokenId.wrap(0).addPoolId(PanopticMath.getPoolId(address(uniPool))).addLeg(
+                0,
+                1,
+                1,
+                0,
+                0,
+                0,
+                15,
+                1
+            )
+        );
+
+        $tempIdList.push(
+            TokenId.wrap(0).addPoolId(PanopticMath.getPoolId(address(uniPool))).addLeg(
+                0,
+                1,
+                1,
+                0,
+                0,
+                0,
+                15,
+                1
+            )
+        );
+
+        $tempIdList.push(
+            TokenId.wrap(0).addPoolId(PanopticMath.getPoolId(address(uniPool))).addLeg(
+                0,
+                1,
+                1,
+                1,
+                0,
+                0,
+                15,
+                1
+            )
+        );
+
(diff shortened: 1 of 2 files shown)


PeakyCryptos (approved):

test/foundry/core/Misc.t.sol
```diff
@@ -480,6 +480,105 @@ contract Misctest is Test, PositionUtils {
...
+            uniPool.liquidity() - 1,
+            1_000_000_000_000_000_000_000,
+            1_000_000_000_000
+        );
+        swapperc.swapTo(uniPool, 2 ** 96);
+
+        uint256 snap = vm.snapshot();
+        vm.startPrank(Charlie);
+
+        for (uint256 i = 0; i < 10; i++) {
```
flanagansteve, line 673: Did you pick 10 for a particular reason here (IE maybe the scale of options minted to the posIdList and the tempIdList require the mint-then-burn-er to do this exactly 10 times?)

If not, is there merit to making this as small a number of loops as possible such that it _would_ cause a problem before your changes in this PR to PanopticPool.sol, to make this test more sensitive and more likely to catch a small change in our math?

Conversely, is it worth making a paired test with this one that uses a very large number of mint-then-burn cycles, so we cover both edgecases (someone doing a few of these cycles and someone doing a bunch)?

flanagansteve (approved):

test/foundry/core/Misc.t.sol
dyedm1: 10 is just an arbitrary number that demonstrates a significant premium difference. This error is caused by an incorrect branch as opposed to a math error, so the only difference in the *amount* of cycles you do would be the magnitude of the error.

We used to fuzz positions/numbers in our regular tests, but we found that stateless fuzz testing makes the tests significantly more complicated to write and rarely, if ever, provides more value than unit tests (the vast majority of errors are not caused by specific inputs, and even if they were, fuzzing only provides a way to fuzz over different *ranges* of numbers (an error found in a small range of inputs can only practically be caught with manual review/symbolic execution because of the breadth of the search space). 

Echidna should be able to cover some different scenarios related to this once we expand it a little.

PR #57, opened 2024-05-01
fix: add support for <1bps pools

dyedm1:
The available Uniswap V3 fee tiers are limited to 1bps, 5bps, 30bps, and 100bps. However, there are proposals for even lower fee tiers -- 0.1bps or 0.01bps. Also, in Uniswap V4, users will be able to select from the full set of possible fee tiers when creating a pool.

Fee tiers in Uniswap are stored in hundredths of basis points (e.g. 5bps would be represented as `500`), but we convert the fee to basis points internally. This means that any fee tier below 1bps would be stored as `0`, which results in a token name with a fee tier of `"0bps"` and an ITM spread fee of 0. 

To rectify this, we need to store the fee tiers in hundredths of basis points, adjust the ITM spread fee denominator accordingly, and add decimal support to the token name calculation. This allows the ITM spread fee to be computed accurately on pools with <1bps and ensure token names on different fee tiers are differentiated (i.e. "0.03bps" or "0.49bps" instead of both "0bps").

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -115,7 +115,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     uint128 internal s_inAMM;
 
     /// @notice additional risk premium charged on intrinsic value of ITM positions,
-    /// defined in basis points as a multiple of the pool fee / 10_000.
+    /// defined in hundredths of basis points.
     /// @dev The result of the calculation is stored instead of the multiple to save gas during usage.
     /// When the fee is set, the multiple is calculated and stored
     uint128 internal s_ITMSpreadFee;
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -223,12 +223,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         // store the Panoptic pool for this collateral token
         s_panopticPool = panopticPool;
 
-        // cache the pool fee in basis points
-        uint24 _poolFee;
-        unchecked {
-            _poolFee = fee / 100;
-        }
-        s_poolFee = _poolFee;
+        // cache the pool fee in hundredths of basis points
+        s_poolFee = fee;
 
         // Stores the addresses of the underlying tracked tokens.
         s_univ3token0 = token0;
@@ -239,7 +235,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
 
         // Additional risk premium charged on intrinsic value of ITM positions
         unchecked {
-            s_ITMSpreadFee = uint128((ITM_SPREAD_MULTIPLIER * _poolFee) / DECIMALS);
+            s_ITMSpreadFee = uint128((ITM_SPREAD_MULTIPLIER * fee) / DECIMALS);
         }
     }
 
@@ -1088,7 +1084,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                 // the swap commission is paid on the intrinsic value, and it is always positive
                 uint256 swapCommission = Math.unsafeDivRoundingUp(
                     s_ITMSpreadFee * uint256(Math.abs(intrinsicValue)),
-                    DECIMALS
+                    DECIMALS * 100
                 );
 
                 // set the exchanged amount to the sum of the intrinsic value and swapCommission
--- a/contracts/libraries/InteractionHelper.sol
+++ b/contracts/libraries/InteractionHelper.sol
@@ -42,7 +42,7 @@ library InteractionHelper {
     /// @param token0 The token0 in the Uniswap Pool
     /// @param token1 The token1 in the Uniswap Pool
     /// @param isToken0 Whether the collateral token computing the name is for token0 or token1
-    /// @param fee The fee of the Uniswap pool in basis points
+    /// @param fee The fee of the Uniswap pool in hundredths of basis points
     /// @param prefix A constant string appended to the start of the token name
     /// @return The complete name of the collateral token calling this function
     function computeName(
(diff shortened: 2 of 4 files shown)


PeakyCryptos (approved):

contracts/CollateralTracker.sol
```diff
@@ -239,7 +235,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
 
         // Additional risk premium charged on intrinsic value of ITM positions
         unchecked {
-            s_ITMSpreadFee = uint128((ITM_SPREAD_MULTIPLIER * _poolFee) / DECIMALS);
+            s_ITMSpreadFee = uint128((ITM_SPREAD_MULTIPLIER * fee) / DECIMALS);
```
flanagansteve, line 238: Is our description of `ITM_SPREAD_MULTIPLIER` still accurate?

```
    /// @notice multiplier, ***in basis points***, to the pool fee that is charged on the intrinsic value of ITM positions.
    /// e.g. ITM_SPREAD_MULTIPLIER = 20_000, s_ITMSpreadFee = 2 * -->s_poolFee<--
    uint256 immutable ITM_SPREAD_MULTIPLIER;
```

(emphasis on ITM_SPREAD_MULTIPLIER being in basis points, and the described calculation using `s_ITMSpreadFee = 2 * s_poolFee` when we seem to have started setting s_ITMSpreadFee as a product of `fee`)

does this PR's changes mean that in the future we'll pass in a `_ITMSpreadMultiplier` (which inits `ITM_SPREAD_MULTIPLIER`) in basis points to the CollateralTracker `constructor`?

contracts/CollateralTracker.sol
```diff
@@ -1088,7 +1084,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                 // the swap commission is paid on the intrinsic value, and it is always positive
                 uint256 swapCommission = Math.unsafeDivRoundingUp(
                     s_ITMSpreadFee * uint256(Math.abs(intrinsicValue)),
-                    DECIMALS
+                    DECIMALS * 100
```
flanagansteve, line 1132: It seems like the only changes we need to make are places using `s_ITMSpreadFee`, which you have covered. Confirming that the other usages of DECIMAL such as in previewDeposit / maxMint / previewMint, or internal methods, don't need any changes?

contracts/libraries/InteractionHelper.sol
```diff
@@ -78,7 +78,14 @@ library InteractionHelper {
                     "/",
                     symbol1,
                     " ",
-                    Strings.toString(fee),
+                    Strings.toString(fee / 100),
```
flanagansteve, line 81: Sanity checking myself - it seems like you anticipated this upcoming change in a comment in CollateralTracker.t.sol?

```
    function test_Success_name(uint256 x) public {
        _initWorld(x);

        // string memory expectedName =
        //     string.concat(
        //             "POPT-V1",
        //             " ",
        //             IERC20Metadata(s_univ3token0).symbol(),
        //             " LP on ",
        //             symbol0,
        //             "/",
        //             symbol1,
        //             " ",
        //             fee % 100 == 0
        //                 ? Strings.toString(fee / 100)
        //                 : string.concat(Strings.toString(fee / 100), ".", Strings.toString(fee % 100)),
        //             "bps"
        //         );

        string memory returnedName = collateralToken0.name();
        console2.log(returnedName);
    }
```

Because it doesn't show up in the diff for this PR, but it matches the code you have here

test/foundry/core/PanopticPool.t.sol
```diff
@@ -1551,12 +1551,12 @@ contract PanopticPoolTest is PositionUtils {
...
+            bytes32(uint256((2 * uint256(fee)) + (uint256(fee) << 128)))
         ); // ITMSpreadFee + poolFee
 
         assertEq(
             vm.load(address(ct1), bytes32(uint256(8))),
-            bytes32(uint256((2 * uint256(fee)) / 100 + (uint256(fee / 100) << 128)))
+            bytes32(uint256((2 * uint256(fee)) + (uint256(fee) << 128)))
         ); // ITMSpreadFee + poolFee
 
         assertEq(vm.load(address(ct0), bytes32(uint256(9))), bytes32(uint256(0))); // 0
```
flanagansteve, line 1560: Am i understanding correctly that values in this test file such as ITMSpread in `test_Success_mintOptions_ITMShortCall` should be unaffected by our changes? They are continuing to multiply by 100 which was a bit surprising to me given the changes we made to s_ITMSpreadFee:

```
int256 ITMSpread = notionalVal > 0
  ? (notionalVal * int24(2 * (fee / 100))) / 10_000
  : -(notionalVal * int24(2 * (fee / 100))) / 10_000;
```

flanagansteve (approved):

contracts/CollateralTracker.sol
dyedm1: Yes. ITM_SPREAD_MULTIPLIER is a multiplier in basis points to the pool fee. You could replace the pool fee with any arbitrary quantity -- the fact that it is in hundredths of basis points has no bearing on the relative effect of ITM_SPREAD_MULTIPLIER. 

(we already pass in that multiplier to this constructor -- this PR does not alter how that value is defined)

contracts/libraries/InteractionHelper.sol
dyedm1: That (previous) implementation wasn't necessarily to support <1bps pools but to support >1bps pools with fractional basis points. It was removed because we deemed it unnecessary given that no such pools actually exist.

test/foundry/core/PanopticPool.t.sol
dyedm1: Yes, this implementation will work exactly the same for pools with integer bp fees.

contracts/CollateralTracker.sol
dyedm1: Yes, the pool fee is the only quantity not represented in basis points

PR #51, opened 2024-04-21
Mint ERC721 from the Panoptic Factory when creating new pools

guil-lambert:
Issue the Panoptic NFTs as ERC721 when calling `deployNewPool`

Will likely be merged with: https://github.com/panoptic-labs/panoptic-NFTs

PR #61, opened 2024-05-01
feat: add slippage checks to factory

dyedm1:
This was brought up in the C4 audit. I initially disputed it but upon further thought it's a good point. 

If a user does infinite or very large approvals to the Panoptic Factory (our app should not do this by default, but some wallets like MetaMask detect approval transactions and allow the user to change the amount requested by the application anyway), that user could be frontrun for the entire approved value of (one) of their deposit tokens if someone manipulates the price before their `deployNewPool` transaction.

I added `amount0Max` and `amount1Max` parameters through which users can specify the maximum amount of tokens they would be willing to pay when deploying a pool. Our frontend can compute the values to set for these parameters from the current price and add a small buffer.

### DIFF
--- a/contracts/PanopticFactory.sol
+++ b/contracts/PanopticFactory.sol
@@ -269,7 +269,7 @@ contract PanopticFactory is Multicall {
         // The SFPM will `safeTransferFrom` tokens from the donor during the mint callback
         (uint256 amount0, uint256 amount1) = _mintFullRange(v3Pool, token0, token1, fee);
 
-        if (amount0 >= amount0Max || amount1 >= amount1Max) revert Errors.PriceBoundFail();
+        if (amount0 > amount0Max || amount1 > amount1Max) revert Errors.PriceBoundFail();
 
         emit PoolDeployed(
             newPoolContract,
--- a/contracts/PanopticFactory.sol
+++ b/contracts/PanopticFactory.sol
@@ -206,12 +206,16 @@ contract PanopticFactory is Multicall {
     /// @param token1 Address of token1 for the underlying Uniswap v3 pool
     /// @param fee The fee tier of the underlying Uniswap v3 pool, denominated in hundredths of bips
     /// @param salt User-defined salt used in CREATE2 for the PanopticPool (must contain caller addr as first 20 bytes)
+    /// @param amount0Max The maximum amount of token0 to spend on the full-range deployment, which serves as a slippage check
+    /// @param amount1Max The maximum amount of token1 to spend on the full-range deployment, which serves as a slippage check
     /// @return newPoolContract The address of the newly deployed Panoptic pool
     function deployNewPool(
         address token0,
         address token1,
         uint24 fee,
-        bytes32 salt
+        bytes32 salt,
+        uint256 amount0Max,
+        uint256 amount1Max
     ) external returns (PanopticPool newPoolContract) {
         // sort the tokens, if necessary:
         (token0, token1) = token0 < token1 ? (token0, token1) : (token1, token0);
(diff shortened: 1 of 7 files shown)


contracts/PanopticFactory.sol
```diff
@@ -257,13 +261,15 @@ contract PanopticFactory is Multicall {
         // When that happens, there will be a period of time where the PanopticPool is deployed, but not (safely) usable
         v3Pool.increaseObservationCardinalityNext(CARDINALITY_INCREASE);
 
+        // Issue reward NFT to donor
```
flanagansteve, line 218: Did you meant to commit this in this PR or should we add NFT issuance for donor in a dedicated PR?

also, we should likely have a test to see successful NFT issuance

dyedm1: > Did you meant to commit this in this PR or should we add NFT issuance for donor in a dedicated PR?

I just moved it, it was already present below

> also, we should likely have a test to see successful NFT issuance

Yes, we had one at one point but I changed this last-minute before the audit so our half-implemented reward NFT scheme wasn't distracting to auditors (the actual NFT issuance part isn't in the scope of that audit). Guillaume is changing this in PR #51 so tests for the final impl will be committed there.

contracts/PanopticFactory.sol
```diff
@@ -257,13 +261,15 @@ contract PanopticFactory is Multicall {
...
+        DONOR_NFT.issueNFT(msg.sender, newPoolContract, token0, token1, fee);
+
         // Mints the full-range initial deposit
         // which is why the deployer becomes also a "donor" of full-range liquidity
         // The SFPM will `safeTransferFrom` tokens from the donor during the mint callback
         (uint256 amount0, uint256 amount1) = _mintFullRange(v3Pool, token0, token1, fee);
 
-        // Issue reward NFT to donor
-        DONOR_NFT.issueNFT(msg.sender, newPoolContract, token0, token1, fee);
+        if (amount0 > amount0Max || amount1 > amount1Max) revert Errors.PriceBoundFail();
```
PeakyCryptos, line 226: PriceBoundFail is used to assert price/tick limits elsewhere. Maybe Add a new error type? Otherwise all LGTM

PeakyCryptos (approved):

contracts/PanopticFactory.sol
dyedm1: This is functionally a price slippage check, so I think `PriceBoundFail` is appropriate (& should be interpreted similarly in the UI).

scripts/DeployTestPool.s.sol
```diff
@@ -44,7 +44,9 @@ contract DeployTestPool is Script {
             address(token0),
             address(token1),
             500,
-            bytes32(uint256(uint160(vm.addr(DEPLOYER_PRIVATE_KEY))) << 96)
+            bytes32(uint256(uint160(vm.addr(DEPLOYER_PRIVATE_KEY))) << 96),
+            type(uint256).max,
```
flanagansteve, line 48: I think there's an argument to make `amount0max` and `amount1max` env vars for this script, too, instead of hard coding the uint256 max. I could see us wanting to change things up without going through a code change if we notice people frontrunning our deployments.

flanagansteve (approved):
Nice!

scripts/DeployTestPool.s.sol
dyedm1: If we deploy pools on mainnet through a script (we'll probably just use the UI), we'll compute the slippage values on the fly rather than passing them in. This script is just for playing around on testnet.

PR #62, opened 2024-05-02
feat: rm ownership/gated launch from factory

dyedm1:
We're no longer doing a permissioned launch with a few pools before enabling permissionless pool deployment, so doesn't make sense to have this extra surface area.

### DIFF
--- a/contracts/PanopticFactory.sol
+++ b/contracts/PanopticFactory.sol
@@ -28,11 +28,6 @@ contract PanopticFactory is Multicall {
                                  EVENTS
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice Emitted when the ownership of the factory is transferred.
-    /// @param oldOwner The previous owner of the factory
-    /// @param newOwner The new owner of the factory
-    event OwnershipTransferred(address indexed oldOwner, address indexed newOwner);
-
     /// @notice Emitted when a Panoptic Pool is created.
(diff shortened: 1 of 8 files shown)


contracts/PanopticFactory.sol
```diff
@@ -92,12 +92,6 @@ contract PanopticFactory is Multicall {
                                 STORAGE
```
flanagansteve, line 87: I think you can also remove the `event OwnershipTransferred` definition above on line 34

test/foundry/core/PanopticFactory.t.sol
```diff
@@ -394,17 +295,6 @@ contract PanopticFactoryTest is Test {
...
     }
 
-    function test_Fail_permissioned(uint256 x, uint96 nonce) public {
-        _initWorld(x);
-
-        vm.startPrank(Deployer);
-
-        uint256 salt = uint256(nonce) + (uint256(uint160(Deployer)) << 96);
-
-        vm.expectRevert(Errors.NotOwner.selector);
```
flanagansteve, line 404: You can also remove the definition of this NotOwner error in `contracts/libraries/Errors.sol`, as its no longer used. if you'd like to stay aligned with some other standard library we could of course keep it too.

flanagansteve (approved):

PeakyCryptos (approved):

PR #63, opened 2024-05-08
feat: optional ITM swaps

dyedm1:
This PR
- removes support for the undocumented `tickLimitLow==tickLimitHigh` flag for disabling slippage checks
- reinstates `tickLimitHigh < tickLimitLow` flag support in the PanopticPool, allowing users to skip ITM swaps during mints and burns (will document this more clearly in a follow up later)
- disables ITM swapping during force exercises and liquidations

ITM swaps (in theory) improve trading UX by allowing users to denominate their profits and losses due to intrinsic value changes in a single token instead of paying to exercise their option and receiving the proceeds directly.

In traditional finance, options contracts are exchanged for USD but denominated in stock units. This means unwinding completely requires an extra stock leg: to exercise an ITM put you must purchase the stock on another exchange, and to exercise an ITM call you must sell the stocks received after exercising the options.

That's not so bad. In Panoptic, however, things get even more complicated. Opening or closing ITM positions requires either providing a token not borrowed/previously removed from the Uniswap pool or paying back a loan/funding a strategy in a different token (short/long). Without ITM swaps, users would need far more tokens available to create and settle options (ex: If a user wants to sell a (1-tick-wide) 1000 USDC put at s=1000, but the price is 999 USDC, they would need to provide 999 USDC worth of ETH to create the position and hold the borrowed 1000 USDC in their account, effectively limiting their leverage to ~1x even though the capital efficiency remains the same post-mint. If ITM swaps are enabled, they just swap the 1000 USDC they borrowed for ETH instead).

That being said, there's nothing inherently insecure about options traders settling manually, and because we have fairly robust handling of token deltas already, it doesn't make sense to arbitrarily force ITM swaps. Many (or even most) users will probably want ITM swaps in some form for convenience, but swapping through Uniswap directly can get expensive and some users (especially those migrating from Uniswap) have a lower tolerance for execution costs. At some point (maybe in a future version of the protocol) ITM swaps will probably occur in separate frames instead of being enshrined.

For force exercises and liquidations, ITM swaps increase the attack surface significantly and create incentives for price manipulation (as the swapping account has no control over slippage). The impact on liquidations is somewhat limited because we have an enshrined slippage check against an oracle price, but that check can lead to liquidation delays in certain situations and still allows some degree of manipulation (ideally, we would remove that check in a future PR after swaps are disabled).

Force exercisors compensate exercisees for unfavorable token deltas in long legs between the current and oracle prices, but that does not cover price manipulation outside the range of the position. (ex: Alice purchases a 1 ETH (1-tick-wide) call at S=1000. The price moves organically to 2000 (so Alice's profit should be ~1000 USDC). Before Alice can exercise her position, Bob manipulates the price to 1001 USDC/ETH and force exercises Alice. At that price, Alice still owes 1000 USDC to the option seller, but is now forced to sell her ETH at a steep discount to Bob and is left with a profit (or even a loss!) of <1 USDC. If ITM swaps are disabled, settlement at a price of 1001 proceeds exactly the same as it would at 2000 (Alice pays 1000 USDC to recreate the LP position and receives the 1 ETH she was holding, which will be worth 2000 USDC in the open market regardless of what the pool price says at that point)).

We already have mechanisms for liquidators and exercisors to exchange tokens required to settle the position with surplus tokens, so disabling the ITM swaps there to solve these issues doesn't have any new side effects (besides being a potential inconvenience for force exercisees)

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1031,18 +1031,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // current available assets belonging to PLPs (updated after settlement) excluding any premium paid
             int256 updatedAssets = int256(uint256(s_poolAssets)) - swappedAmount;
 
-            // add premium to be paid/collected on position close
-            int256 tokenToPay = -realizedPremium;
-
-            // if burning ITM and swap occurred, compute tokens to be paid through exercise and add swap fees
-            int256 intrinsicValue = swappedAmount - (longAmount - shortAmount);
-
-            if ((intrinsicValue != 0) && ((shortAmount != 0) || (longAmount != 0))) {
-                // intrinsic value is the amount that need to be exchanged due to burning in-the-money
-
-                // add the intrinsic value to the tokenToPay
-                tokenToPay += intrinsicValue;
-            }
+            // add premium and token deltas not covered by swap to be paid/collected on position close
+            int256 tokenToPay = swappedAmount - (longAmount - shortAmount) - realizedPremium;
 
             if (tokenToPay > 0) {
                 // if user must pay tokens, burn them from user balance (revert if balance too small)
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -489,29 +489,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
         return (portfolioPremium, balances);
     }
 
-    /// @notice Disable slippage checks if tickLimitLow == tickLimitHigh and reverses ticks if given in correct order to enable ITM swaps
-    /// @param tickLimitLow The lower slippage limit on the tick.
-    /// @param tickLimitHigh The upper slippage limit on the tick.
-    /// @return tickLimitLow Adjusted value for the lower tick limit.
-    /// @return tickLimitHigh Adjusted value for the upper tick limit.
-    function _getSlippageLimits(
-        int24 tickLimitLow,
-        int24 tickLimitHigh
-    ) internal pure returns (int24, int24) {
-        // disable slippage checks if tickLimitLow == tickLimitHigh
-        if (tickLimitLow == tickLimitHigh) {
-            // note the reversed order of the ticks
-            return (MAX_SWAP_TICK, MIN_SWAP_TICK);
-        }
-
-        // ensure tick limits are reversed (the SFPM uses low > high as a flag to do ITM swaps, which we need)
-        if (tickLimitLow < tickLimitHigh) {
-            return (tickLimitHigh, tickLimitLow);
-        }
-
-        return (tickLimitLow, tickLimitHigh);
-    }
-
     /*//////////////////////////////////////////////////////////////
                           ONBOARD MEDIAN TWAP
     //////////////////////////////////////////////////////////////*/
@@ -624,8 +601,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
         // do duplicate checks and the checks related to minting and positions
         _validatePositionList(msg.sender, positionIdList, 1);
-
-        (tickLimitLow, tickLimitHigh) = _getSlippageLimits(tickLimitLow, tickLimitHigh);
 
         // make sure the tokenId is for this Panoptic pool
         if (tokenId.poolId() != SFPM.getPoolId(address(s_univ3pool)))
@@ -828,9 +803,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
         int24 tickLimitLow,
         int24 tickLimitHigh
     ) internal returns (LeftRightSigned paidAmounts, LeftRightSigned[4] memory premiaByLeg) {
-        // Ensure that the current price is within the tick limits
-        (tickLimitLow, tickLimitHigh) = _getSlippageLimits(tickLimitLow, tickLimitHigh);
-
         uint128 positionSize = s_positionBalance[owner][tokenId].rightSlot();
 
         LeftRightSigned premiaOwed;
@@ -1083,8 +1055,8 @@ contract PanopticPool is ERC1155Holder, Multicall {
             // Note: tick limits are not applied here since it is not the liquidator's position being liquidated
             (netExchanged, premiasByLeg) = _burnAllOptionsFrom(
                 liquidatee,
-                0,
-                0,
+                MIN_SWAP_TICK,
+                MAX_SWAP_TICK,
                 DONOT_COMMIT_LONG_SETTLED,
                 positionIdList
             );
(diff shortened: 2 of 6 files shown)


guil-lambert:
This looks similar to one of my earlier comment on slack about allow users to formally exercise their option (ie. buy/sell through token exchanges instead of swapping):
image


But I'm not sure enforcing it for forceExercises and liquidations leads to a better outcome. It comes to `potential for less profits` vs `potential for larger losses` in the force exercise scenario (I'm still exploring how the liquidations come into play here).

What about this reverse situation:

- Price is 1000
- User purchased a 2000 call. 
- The exercisor/attacker moves the price to 2001, forces the user to exercise their option at 2000 with no ITM swap, then moves back the price to 1000. 
- **No ITM swaps**: The user now purchased 1 ETH for 2000 when the "real" price was 1000, `net = (-2000USDC, +1ETH)`, with a net loss of `-1000USDC=-1ETH`. 
- **ITM swaps**: In the ITM swap scenario, the user net balance for the user would be `net = (0, +0.0005ETH)`. 
- **Exercisor could force user to execute at the strike price even if initially OTM.**


Or, in case of liquidations (harder to pull off though due to multi-block price manipulation, although possible in low volume tokens):
- Price is 1000
- User has sold a 2000 ETH call. 
- The liquidator/exercisor moves the price to 5000, making the account liquidatable. 
- Waits long enough/pokes the median/does swaps/etc so that the liquidation can proceed
- Liquidation forces the user to sell 1 ETH for 2000USDC with no ITM swaps, then moves the price back to 1000. 
- **No ITM swaps**: The net balance for the user is `net = (+2000USDC, -1ETH)`, which is a net gain of `1000USDC=1ETH`. 
- **ITM swaps**: the net balance delta would be `net = (0, -0.6ETH)`, which is a loss of `600USDC = -0.6ETH`. 
- **Users would want to liquidate themselves.**

dyedm1:
Wow, I had completely forgotten about that conversation 6 months ago lol! Maybe it was sitting somewhere in my subconscious.

The conclusions from there were similar (it could be beneficial for some users, but most option buyers will be trading on leverage and will therefore lack the required collateral to formally exercise their option). I think it's much more compelling for option *sellers*, many of whom are more sophisticated and will be migrating from Uniswap. Those users have higher expectations for execution quality and are more likely to have the funds available to formally open/close their positions.

> The exercisor/attacker moves the price to 2001, forces the user to exercise their option at 2000 with no ITM swap, then moves back the price to 1000.

That's correct. This exact situation is what the loss compensation component of the force exercise fee is for (if somebody tried this, the oracle price would be hovering somewhere around 1000 and the exercisor would have to provide the difference in position composition between those two prices, so the exercisee would get the 2000 USDC they paid back in exchange for their ETH (leaving them with a ~0 pnl). (position comp near 1000 - 1 ETH 0 USDC, position comp at 2001 - 2000 USDC 0 ETH, so +2000 USDC -1ETH )

> Or, in case of liquidations (harder to pull off though due to multi-block price manipulation, although possible in low volume tokens):

Yeah, this is why I haven't yet removed the slippage check on liquidations in this PR. You can't really do too much additional damage by moving the price 5% (as is also true with ITM swap manipulation), but there may be a better way to mitigate this type of manipulation (perhaps by deducting the position deltas from the bonus as we do in force exercises, so the worst thing that can happen is the liquidation being unprofitable for a period of time instead of being DoSed when the oracle price is stale).

dyedm1:
> But I'm not sure enforcing it for forceExercises and liquidations leads to a better outcome. It comes to potential for less profits vs potential for larger losses in the force exercise scenario (I'm still exploring how the liquidations come into play here).

It's also important to note that if you purchase ITM options, there is not necessarily a difference between the two.

dyedm1:
> Yeah, this is why I haven't yet removed the slippage check on liquidations in this PR. You can't really do too much additional damage by moving the price 5% (as is also true with ITM swap manipulation), but there may be a better way to mitigate this type of manipulation (perhaps by deducting the position deltas from the bonus as we do in force exercises, so the worst thing that can happen is the liquidation being unprofitable for a period of time instead of being DoSed when the oracle price is stale).

Changed my mind on this. I was testing an implementation where the force exercises and liquidations share a `getLossCompensationDelta` function they use to determine the amount of tokens to settle with the liquidator/force exercisor (based on the delta), but it turns out that the required changes are somewhat more complex than replacing the stale TWAP checks with that loss compensation function.

Namely, we often rely on `sqrtPriceX96Final` as a token conversion price. Removing the stale TWAP restriction prevents us from having any kind of confidence as to the verity of that price, so we would have to retool all the calculations to use an oracle price. 

Given the stage we are at now (essential bug fixes/patches only) I don't think it makes sense to include a change this serious in our diff audit, especially for an ambiguous increase in liquidation efficiency/some level of MEV reduction.

contracts/CollateralTracker.sol
```diff
@@ -1072,18 +1072,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // current available assets belonging to PLPs (updated after settlement) excluding any premium paid
             int256 updatedAssets = int256(uint256(s_poolAssets)) - swappedAmount;
 
-            // add premium to be paid/collected on position close
-            int256 tokenToPay = -realizedPremium;
-
-            // if burning ITM and swap occurred, compute tokens to be paid through exercise and add swap fees
-            int256 intrinsicValue = swappedAmount - (longAmount - shortAmount);
-
-            if ((intrinsicValue != 0) && ((shortAmount != 0) || (longAmount != 0))) {
```
flanagansteve, line 1081: For my own understanding: I think you're removing this condition because:

- it existed to let callers of `.exercise` pass in a special 0 argument for both `shortAmount` and `longAmount`
- when passed, this method skipped over increasing `tokenToPay` by `swappedAmount - (longAmount - shortAmount)`
  - (which is really just `swappedAmount` when the longAmount and shortAmount are 0.)
- Skipping over that addition to `tokenToPay` was, in effect, creating an ITM swap (somehow?).
- You want to remove that conditional skip now because you're removing the ITM swap from liquidations and force exercises. You could have accomplished that by just changing the hardcoded 0 args in PanopticPool.sol (which your PR also does), but then this conditional skip wasn't being used, so we might as well remove that too. 

Am i understanding right? If so, how is it that increasing tokenToPay by swappedAmount caused the ITM swap in the first place?

I expected your diff here to look more like removing an actual `.swap()` call somewhere, or inverting `tickLimitHigh` and `tickLimitLow`, but of course I'm still learning how we handle various exercise scenarios

flanagansteve (approved):

contracts/CollateralTracker.sol
PeakyCryptos: https://github.com/panoptic-labs/panoptic-v1-core-private/blob/352685cfdc2856686c86864189fde48f54ae60fa/contracts/SemiFungiblePositionManager.sol#L715

A swap occurs if min swap tick > max swap tick. Flipping the order of how these ticks are passed into the SFPM is what triggers a swap to occur

flanagansteve: Right that I understand. But trying to understand why this conditional was removed. It only seems to affect `tokenToPay`'s value and trying to see whether that relates to ITM swaps at all or not

dyedm1: > Right that I understand. But trying to understand why this conditional was removed. It only seems to affect tokenToPay's value and trying to see whether that relates to ITM swaps at all or not

That conditional prevents the required swap deltas from being added for ITM positions where all legs have a single tokenType. Previously, this was fine because the ITM swap ensured that delta was always cancelled for non-tokenType tokens, but if there is no swap or an incomplete swap the accounting would wrong.

Ex: Price is 2000, Bob mints a 1 ETH short call @ s=1000 on ETH with ITM swaps disabled. In theory, he should pay 1000 USDC to create the position and receive the 1 ETH he borrowed, but due to this condition (longAmount = shortAmount = 0 for USDC collateralTracker), the 1000 USDC is never actually deducted.

flanagansteve: Aha! That makes sense, thanks

PeakyCryptos (approved):

PR #64, opened 2024-05-13
feat: withdrawals for accounts with open positions

dyedm1:
Added a `withdraw(uint256, address, address, TokenId[])` overload to the standard 4626 interface for users that want to withdraw collateral before closing all their positions. I don't think there's any need for a `redeem` counterpart -- we only use `withdraw(uint256, address, address)` in the UI as of right now.

To accomplish this, I exposed a function, `validateCollateralWithdrawable`, on the PanopticPool. This function wraps a standard internal solvency check (fast oracle tick + slow fallback) with `BP_DECREASE_BUFFER`.

The solvency checking logic could have been moved to a library directly accessible from `CollateralTracker`, but that would add significant gas/call overhead given that the position size/accumulator data is stored in the `PanopticPool`. The former method is cheaper and results in a significantly smaller diff than the latter.


contracts/PanopticPool.sol
```diff
@@ -341,6 +341,19 @@ contract PanopticPool is ERC1155Holder, Multicall {
         }
     }
 
+    /// @notice Determines if account is eligible to withdraw or transfer collateral.
+    /// @dev Checks whether account is solvent with `BP_DECREASE_BUFFER` according to `_validateSolvency`.
+    /// @dev Prevents insolvent and near-insolvent accounts from withdrawing collateral before they are liquidated.
+    /// @dev Reverts if account is not solvent with `BP_DECREASE_BUFFER`.
+    /// @param user The account to check for collateral withdrawal eligibility
+    /// @param positionIdList The list of all option positions held by `user`
+    function validateCollateralWithdrawable(
```
flanagansteve, line 350: Should we consider just making `_validateSolvency` a public method if all this method does is call it and insert in the constant `BP_DECREASE_BUFFER`?

contracts/CollateralTracker.sol
```diff
@@ -545,6 +545,53 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+
+            if (allowed != type(uint256).max) allowance[owner][msg.sender] = allowed - shares;
+        }
+
+        // burn collateral shares of the Panoptic Pool funds (this ERC20 token)
+        _burn(owner, shares);
+
+        // update tracked asset balance
+        unchecked {
+            s_poolAssets -= uint128(assets);
```
flanagansteve, line 576: I am likely missing something here, but: why are you able to safely cast `assets` down to uint128 if it comes in as a uint256 argument?

contracts/PanopticPool.sol
dyedm1: I wanted to separate concerns here a little to leave the door open for potentially using a different buffer for this down the line. This way, `CollateralTracker` has a standard interface it can use to determine if an account is allowed to withdraw. That interface remains the same even if the underlying mechanism changes.

PeakyCryptos (changes requested):
Remember to remove the original withdraw function L#511

contracts/CollateralTracker.sol
dyedm1: If a user is able to accumulate `>2^128-1` assets worth of shares, the protocol cannot function on that pool either way.

dyedm1:
> Remember to remove the original withdraw function L#511

Nah, we still need that one because `withdraw(uint256, address, address, uint256[])` is not ERC4626-compliant.

PeakyCryptos (approved):

contracts/PanopticPool.sol
flanagansteve: Gotcha ok - makes sense!

flanagansteve (approved):

PR #65, opened 2024-05-16
feat: modular metadata for NFTs

dyedm1:
The previous implementation of the NFT constructed it from hardcoded (and sometimes compressed) component strings in various libraries. This is not a scalable approach due to contract size limitations (Spurious Dragon is `24.576 KB` while `Art.sol` was almost `100 KB`).

To resolve this, I have developed a modular metadata system that can be configured through a JSON file. 

A bunjs script ( `metadata/compiler.js`) parses `metadata/FactoryNFT.json` and compiles all data contiguously into chunks with a maximum size of `24.576 KB` which can deployed as contracts. That script writes the bytecode of the contracts to deploy, the property names, the indexes of all the values under each property name, and pointers to the contract slices that contain the values to `metadata/out/MetadataPackage.json`. 

`MetadataPackage.json` is read/parsed with Foundry cheatcodes in a Forge script. Information from that file is used to deploy the data contracts and provide constructor arguments containing the indexed pointers to `PanopticFactory` which stores them at construction time and uses the pointers to retrieve the information stored in the data contracts when `tokenURI` is called.

Before deploying a contract or running tests, `bun run ./metadata/compiler.js` must be run to generate the metadata package.

 `FactoryNFT.json` can be updated with new or edited parameters and the information will transfer seamlessly once `compiler.js` is run to generate the new metadata package.


contracts/base/FactoryNFT.sol
```diff
@@ -0,0 +1,334 @@
...
+    /// @param properties An array of identifiers for different categories of metadata
+    /// @param indices A nested array of keys for K-V metadata pairs for each property in `properties`
+    /// @param pointers Contains pointers to the metadata values stored in contract data slices for each index in `indices`
+    constructor(
+        bytes32[] memory properties,
+        uint256[][] memory indices,
+        Pointer[][] memory pointers
+    )
+        MetadataStore(properties, indices, pointers)
+        ERC721("Panoptic Factory Deployer NFTs", "PANOPTIC-NFT")
```
flanagansteve, line 32: It seems possible we make different kinds of NFTs in the future - therefore, there's maybe an argument for making this ticker "PANOPTIC-DEPL-NFT" to specify a deployer NFT. That's a pretty long ticker, however

contracts/base/FactoryNFT.sol
```diff
@@ -0,0 +1,334 @@
...
+                panopticPool,
+                PanopticMath.safeERC20Symbol(PanopticPool(panopticPool).univ3pool().token0()),
+                PanopticMath.safeERC20Symbol(PanopticPool(panopticPool).univ3pool().token1()),
+                PanopticPool(panopticPool).univ3pool().fee()
+            );
+    }
+
+    /// @notice Returns the metadata URI for a given set of characteristics.
+    /// @param panopticPool The displayed address used to determine the rarity (leading zeros) and lastCharVal (last 4 bits)
+    /// @param symbol0 The symbol of the second token in the pool
```
flanagansteve: Do you have these descriptions swapped? Or is symbol0 the second and symbol1 the first? I would expect the opposite

contracts/base/FactoryNFT.sol
```diff
@@ -0,0 +1,334 @@
...
+        return
+            svgIn
+                .replace("<!-- RARITY -->", write(LibString.toString(rarity)))
+                .replace("<!-- SYMBOL0 -->", write(symbol0, maxSymbolWidth(rarity)))
+                .replace("<!-- SYMBOL1 -->", write(symbol1, maxSymbolWidth(rarity)));
+    }
+
+    /// @notice Get the name of the current chain.
+    /// @return The name of the current chain, or the chain ID if not recognized
+    function getChainName() internal view returns (string memory) {
```
flanagansteve, line 182: Do we wish to add any testnets here? Maybe "Ethereum Sepolia Testnet" for 11155111 and Polygon Mumbai Testnet for 80001?

periphery/DonorNFT.sol
```diff
@@ -1,30 +0,0 @@
-// SPDX-License-Identifier: UNLICENSED
```
flanagansteve, line 1: We still have a contract that references this and the interface - `DeployProtocol.s.sol`. I assume this is some deprecated way of deploying our contracts that we should drop from the codebase now?

flanagansteve (approved):
My only other thought is that it might be wortwhile to add the info in your PR description into our README - IE, something like:

> - To deploy with our NFTs or to run tests, install [bun](https://bun.sh/) and run `bun run ./metadata/compiler.js` before proceeding.

contracts/base/FactoryNFT.sol
dyedm1: The NFT ticker is rarely used for anything, and PANOPTIC-NFT is a rather uninspired ticker anyway, so I doubt this will create conflicts in the future. We can change later if Guillaume thinks of another name.

contracts/base/FactoryNFT.sol
dyedm1: You could make an argument for either "order" but really there's no specific precedence beyond the internal identifiers of 0 and 1. I'll call these token0 and token1 to make it a little less ambiguous.

contracts/base/FactoryNFT.sol
dyedm1: Don't see any reason to -- the stringified chainId should be enough to identify the chain for development purposes.

periphery/DonorNFT.sol
dyedm1: We generally update scripts on an as-needed basis (and keep the previous versions around until a new deployment required). Either way, script for mainnet will not resemble any of the current scripts because we'll be doing CREATE2 deployments.

flanagansteve (approved):
Awesome!

PR #68, opened 2024-05-29
feat: add optional `assertMinCollateralValues` slippage check

dyedm1:
Force exercisors (and in particular, liquidators) can use this more granular slippage check to prevent sandwich attacks/unwanted token deltas without having to trade through a custom contract.

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -328,17 +328,17 @@ contract PanopticPool is ERC1155Holder, Multicall {
                              QUERY HELPERS
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice Reverts if current Uniswap price is not within the provided bounds.
+    /// @notice Reverts if the caller has a lower collateral balance than required to meet the provided `minValue0` and `minValue1`.
     /// @dev Can be used for composable slippage checks with `multicall` (such as for a force exercise or liquidation)
-    /// @dev Can also be used for more granular subtick precision on slippage checks
-    /// @param sqrtLowerBound The lower bound of the acceptable open interval for `currentSqrtPriceX96`
-    /// @param sqrtUpperBound The upper bound of the acceptable open interval for `currentSqrtPriceX96`
-    function assertPriceWithinBounds(uint160 sqrtLowerBound, uint160 sqrtUpperBound) external view {
-        (uint160 currentSqrtPriceX96, , , , , , ) = s_univ3pool.slot0();
-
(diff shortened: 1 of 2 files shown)


PeakyCryptos (approved):

contracts/PanopticPool.sol
```diff
@@ -328,17 +328,17 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-        (uint160 currentSqrtPriceX96, , , , , , ) = s_univ3pool.slot0();
-
-        if (currentSqrtPriceX96 <= sqrtLowerBound || currentSqrtPriceX96 >= sqrtUpperBound) {
-            revert Errors.PriceBoundFail();
-        }
+    /// @notice Reverts if the caller has a lower collateral balance than required to meet the provided `minValue0` and `minValue1`.
+    /// @dev Can be used for composable slippage checks with `multicall` (such as for a force exercise or liquidation).
+    /// @param minValue0 The minimum acceptable `token0` value of collateral
+    /// @param minValue1 The minimum acceptable `token1` value of collateral
+    function assertMinCollateralValues(uint256 minValue0, uint256 minValue1) external view {
```
flanagansteve, line 335: I _think_ that this changes still allow you to do the same check you could previously do with `assertPriceWithinBounds` with `assertMinCollateralValues` - you just have to do the math to convert price to the collateral values, but the collateral values are what you care about anyways

But just double checking that? If not, could it be worth leaving both methods on the contract and just having `assertMinCollateralValues` as an alternative to rather than a replacement of `assertPriceWithinBounds`?

flanagansteve (approved):

contracts/PanopticPool.sol
dyedm1: It's not *quite* the same check, but it does a better job of accomplishing the intended purpose for liquidators and force exercisors. The price is a decent heuristic that will circumscribe gross price manipulation in most cases, but doesn't rule out unexpected results caused by a change in the available liquidity or other factors. This is a more direct approach that allows liquidators/exercisors to assert the properties they actually care about instead of relying on a second-order heuristic.

Price makes a little more sense for opening/closing positions since the composition of your position is directly attributable to the current price.

flanagansteve: Gotcha, and you don't think keeping this older (inferior) `assertPriceWithinBounds` is worth the extra deployment cost?

dyedm1: Yes -- don't see a usecase for it if `assertMinCollateralValues` is available.

PR #69, opened 2024-05-30
Add `calculatePortfolioValue` back

guil-lambert:
This was removed here: https://github.com/panoptic-labs/panoptic-v1-core-private/commit/61be5f8ca994fe32e48fd2da207aaee1355ffb60

Adding it back, this is the only place where one can compute the value of a portfolio using accurate, onchain position balance values


dyedm1:
That should be fine, let me see if it still compiles under size once the final changes are included

dyedm1 (approved):
a little close, but still compiles

PR #70, opened 2024-05-31
fix: miscellaneous low-sev/documentation fixes

dyedm1:
This *should* be the last change before we start the C4 audit. Includes:
- `twapFilter` should return the *actual* median and not the 10th index
- NatSpec formatting should be based on what we want the md-docs to look like -- can be previewed with `forge doc`
- fixes for other significant errors/anachronisms in our documentation/comments

The first point is the only logical change, so review that and also let me know if there's any egregious documentation errors I missed that stick out. I'm sending out the diff today, but we have until Tuesday to make last-minute changes if absolutely necessary.

If anyone wants to review the overall [diff](https://github.com/panoptic-labs/panoptic-v1-core-private/compare/f59ed0f71d70729fb4331de336bc7cdca91abffe..feat/mit-review-diff) between the version that went out to C4 and this commit, I have a gist with a preliminary version of the [audit report](https://gist.github.com/dyedm1/49389c9e94754a5df213d410609eb3cd) to reference.

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -34,7 +34,6 @@ import {TokenId} from "@types/TokenId.sol";
 //
 /// @notice 2) get any gain in capital that results from buying an option that becomes in-the-money.
 contract CollateralTracker is ERC20Minimal, Multicall {
-    // Used for safecasting
     using Math for uint256;
 
     /*//////////////////////////////////////////////////////////////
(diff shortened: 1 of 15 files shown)


contracts/PanopticPool.sol
```diff
@@ -1404,32 +1387,32 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
         return s_collateralToken1;
     }
 
-    /// @notice get the number of positions for an account
-    /// @param user the account to get the positions hash of
-    /// @return _numberOfPositions number of positions in the account
+    /// @notice Get the current number of open positions for an account
+    /// @param user The account to query
+    /// @return _numberOfPositions Number of open positions for `user`
     function numberOfPositions(address user) public view returns (uint256 _numberOfPositions) {
```
flanagansteve, line 1410: Thought: Why do these two view-only helpers, `numberOfPositions` and `getUniV3TWAP`, have named return variables when the others just return unnamed ones? I feel like the names are unnecessary given the name of the function already describing what they are, and would be nice to have consistency among these set of view helpers

May be outside the scope of this PR though / a change we don't want to bother inserting ahead of the diff audit

contracts/CollateralTracker.sol
```diff
@@ -261,7 +269,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     }
 
     /// @notice Returns name of token composed of underlying token symbol and pool data.
-    /// @return name The name of the token.
+    /// @return name The name of the token
```
flanagansteve: Technically the returned value does not have a varname - this doc string seems to say it has the varname `name`

Same goes for the docstring on the returned values in `symbol` and `decimal`

contracts/types/TokenId.sol
```diff
@@ -450,7 +450,7 @@ library TokenIdLibrary {
 
     /// @notice Clear a leg in an option position with index `i`.
     /// @dev set bits of the leg to zero. Also sets the optionRatio and asset to zero of that leg.
-    /// @dev NOTE it's important that the caller fills in the leg details after.
+    /// @dev NOTE: it's important that the caller fills in the leg details after.
```
flanagansteve, line 453: Some of our `NOTE:`s are all-lowercase `note:`s - such as line 105 in the FeesCalc library, line 913 in the PanopticMath library, and line 628 in PanopticPool, as well as several in the SFPM. We also have a `Note:` on line 1053 of PanopticPool. Would be good to make this consistently upcased.

Also, we have an open TODO in contracts/base/FactoryNFT.sol, but i assume thats for a future cleanup as its not part of the C4 audit

contracts/PanopticPool.sol
```diff
@@ -215,77 +206,80 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
     mapping(address account => uint256 positionsHash) internal s_positionsHash;
 
     /*//////////////////////////////////////////////////////////////
                              INITIALIZATION
     //////////////////////////////////////////////////////////////*/
 
-    /// @notice During construction: sets the address of the panoptic factory smart contract and the SemiFungiblePositionMananger (SFPM).
-    /// @param _sfpm The address of the SemiFungiblePositionManager (SFPM) contract.
+    /// @notice Store the address of the canonical SemiFungiblePositionManager (SFPM) contract.
+    /// @param _sfpm The address of the SemiFungiblePositionManager (SFPM) contract
```
flanagansteve: I think you can just use SFPM on this line, or just SemiFungiblePositionManager, rather than SemiFungiblePositionManager (SFPM) since you just defined the abbreviation on the line prior

flanagansteve (approved):
✅ confirmed all functions have docstrings now with appropriately labelled arg descriptions and return value descriptions (aside from the return values i noted in collateral tracker)
✅ changed logic for median in twapFilter makes sense

I didn't read through every docstring you touched but scanned for things i thought could be improved. If you want a more thorough full read of each docstring's content let me know.

contracts/PanopticPool.sol
dyedm1: Named return vars are not just semantic (in that case we would name every one) -- they actually define a variable and therefore have an effect on the compiled code. We didn't settle on a consistent convention for this until later on (generally it should just be "wherever needed" which usually means a cumulative return value that is accumulated or read throughout the function.)

It could be a little more consistent, but I don't think it's worth it at this point to go through every function and decide whether to use named return variables or not. It's just extra changes for our auditors to review for a small visual/consistency improvement.

contracts/CollateralTracker.sol
dyedm1: Yeah I noticed this when reading through the compiled docs. It does look slightly weird so I'll remove that first varname

contracts/types/TokenId.sol
dyedm1: Sure why not.

>Also, we have an open TODO in contracts/base/FactoryNFT.sol, but i assume thats for a future cleanup as its not part of the C4 audit

It is *technically* part of the C4 audit (though any NFT-related issues will not be in scope). Also, the idea was to create a helper function that converts a pool fee to a string with support for hundredth-bip precision, so that would affect the CollateralTracker as well. I'm doing this one in a separate PR so it's easier to review.

contracts/PanopticPool.sol
dyedm1: Sure why not

contracts/PanopticPool.sol
flanagansteve: > It's just extra changes for our auditors to review for a small visual/consistency improvement.

Gotcha, that makes sense.

> Named return vars are not just semantic (in that case we would name every one) -- they actually define a variable and therefore have an effect on the compiled code. 

This is what i initially believed too - and when i first suggested it i also thought there might be a gas saving from removing the unnecessary variable (and therefore variable assignment). But from further reading / the below manual test, it seems like the solidity compiler is smart enough to treat both the same - this was my proof of concept:

```
// SPDX-License-Identifier: GPL-3.0
pragma solidity ^0.8.13;

contract GasTest {
    uint value = 6;

    /*
    // 2410 gas units
    function getValue() public view returns(uint) {
        return value;
    }
    */


    // also 2410 gas units
    function getValue() public view returns(uint someValue) {
        someValue = value;
    }
}
```

PR #73, opened 2024-05-31
Refactor: SFPM call optimization

guil-lambert:
It turns out the `_updateSettlementPostMint` and `_updateSettlementPostBurn` functions do calls to sfpm that are done again in the same function.

We also loop over the legs twice there too: once in update settlement and once again in `addUserOption` and `_updatePositionDataBurn` for mints/burns.

This PR merges the logic of `_addUserOption` into `_updateSettlementPostMint` and the logic of `_updatePositionDataBurn` into `_updateSettlementPostBurn` to consolidate the calls to sfpm

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -657,7 +657,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
         (LeftRightUnsigned[4] memory collectedByLeg, LeftRightSigned totalSwapped) = SFPM
             .mintTokenizedPosition(tokenId, positionSize, tickLimitLow, tickLimitHigh);
 
-
         // update premium settlement info
         _updateSettlementPostMint(
             tokenId,
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1732,7 +1732,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 leg,
                 liquidityChunk.tickLower(),
                 liquidityChunk.tickUpper(),
-                uint64(Math.min(effectiveLiquidityLimitX32, MAX_SPREAD))
+                isLong == 0 ? MAX_SPREAD : uint64(Math.min(effectiveLiquidityLimitX32, MAX_SPREAD))
             );
 
             // if position is short, update gross premia accumulator
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -788,30 +788,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
         emit OptionBurnt(owner, positionSize, tokenId, premiaOwed);
     }
 
-    /// @notice Update the internal tracking of the owner's position data upon burning a position.
-    /// @param owner The owner of the option position.
-    /// @param tokenId The option position to burn.
-    function _updatePositionDataBurn(address owner, TokenId tokenId) internal {
-        // reset balances and delete stored option data
-        s_positionBalance[owner][tokenId] = LeftRightUnsigned.wrap(0);
-
-        uint256 numLegs = tokenId.countLegs();
-        for (uint256 leg = 0; leg < numLegs; ) {
-            if (tokenId.isLong(leg) == 0) {
-                // Check the liquidity spread, make sure that closing the option does not exceed the MAX_SPREAD allowed
-                (int24 tickLower, int24 tickUpper) = tokenId.asTicks(leg);
-                _checkLiquiditySpread(tokenId, leg, tickLower, tickUpper, MAX_SPREAD);
-            }
-            s_options[owner][tokenId][leg] = LeftRightUnsigned.wrap(0);
-            unchecked {
-                ++leg;
-            }
-        }
-
-        // Update the position list hash (hash = XOR of all keccak256(tokenId)). Remove hash by XOR'ing again
-        _updatePositionsHash(owner, tokenId, !ADD);
-    }
-
     /// @notice Validates the solvency of `user` at the fast oracle tick.
     /// @notice Falls back to the more conservative tick if the delta between the fast and slow oracle exceeds `MAX_SLOW_FAST_DELTA`.
     /// @dev Effectively, this means that the users must be solvent at both the fast and slow oracle ticks if one of them is stale to mint or burn options.
@@ -1534,19 +1510,23 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
         if (tokenId.isLong(legIndex) == 0 || legIndex > 3) revert Errors.NotALongLeg();
 
+        LiquidityChunk liquidityChunk = PanopticMath.getLiquidityChunk(
+            tokenId,
+            legIndex,
+            s_positionBalance[owner][tokenId].rightSlot()
+        );
+
         (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
 
         LeftRightUnsigned accumulatedPremium;
         {
-            (int24 tickLower, int24 tickUpper) = tokenId.asTicks(legIndex);
-
             uint256 tokenType = tokenId.tokenType(legIndex);
             (uint128 premiumAccumulator0, uint128 premiumAccumulator1) = SFPM.getAccountPremium(
                 address(s_univ3pool),
                 address(this),
                 tokenType,
-                tickLower,
-                tickUpper,
+                liquidityChunk.tickLower(),
+                liquidityChunk.tickUpper(),
                 currentTick,
                 1
             );
@@ -1563,11 +1543,9 @@ contract PanopticPool is ERC1155Holder, Multicall {
             accumulatedPremium = accumulatedPremium.sub(premiumAccumulatorsLast);
         }
 
-        uint256 liquidity = PanopticMath
-            .getLiquidityChunk(tokenId, legIndex, s_positionBalance[owner][tokenId].rightSlot())
-            .liquidity();
-
         unchecked {
+            uint256 liquidity = liquidityChunk.liquidity();
+
             // update the realized premia
             LeftRightSigned realizedPremia = LeftRightSigned
                 .wrap(0)
@@ -1650,6 +1628,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
             uint256 grossCurrent1;
             {
                 uint256 tokenType = tokenId.tokenType(leg);
+                // can use (type(int24).max flag because premia accumulators were updated during the mintTokenizedPosition step.
                 (grossCurrent0, grossCurrent1) = SFPM.getAccountPremium(
                     address(s_univ3pool),
                     address(this),
@@ -1667,8 +1646,8 @@ contract PanopticPool is ERC1155Holder, Multicall {
                         .toLeftSlot(uint128(grossCurrent1));
                 }
             }
-            // verify base Liquidity limit only if new position is long
 
+            // verify base Liquidity limit only if new position is long
             // if position is long, ensure that new mints don't deplete strike beyond MAX_SPREAD
             // new totalLiquidity (total sold) = removedLiquidity + netLiquidity (R + N)
             uint256 totalLiquidity = _checkLiquiditySpread(
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -468,8 +468,9 @@ contract PanopticPool is ERC1155Holder, Multicall {
                         )
                     );
 
+                    (uint256 totalLiquidity, , ) = _getLiquidities(tokenId, leg);
                     LeftRightUnsigned availablePremium = _getAvailablePremium(
-                        _getTotalLiquidity(tokenId, leg),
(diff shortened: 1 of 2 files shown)


contracts/PanopticPool.sol
```diff
@@ -1725,61 +1662,91 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+
+            uint256 grossCurrent0;
+            uint256 grossCurrent1;
+            {
+                uint256 tokenType = tokenId.tokenType(leg);
+                // can use (type(int24).max flag because premia accumulators were updated during the mintTokenizedPosition step.
+                (grossCurrent0, grossCurrent1) = SFPM.getAccountPremium(
                     address(s_univ3pool),
                     address(this),
-                    tokenId.tokenType(leg),
```
dyedm1, line 1776: Would be nice if we could inline this, not sure whether it will cause a stack-too-deep or not. If it does, could try moving it before the `chunkKey` declaration maybe

contracts/PanopticPool.sol
```diff
@@ -1725,61 +1662,91 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-                //
-                // T * (C - L) = (T + R) * (C - Ln)
-                // (TC - TL) / (T + R) = C - Ln
-                // Ln = C - (TC - TL)/(T + R)
-                // Ln = (CT + CR - TC + TL)/(T+R)
-                // Ln = (CR + TL)/(T+R)
-
-                uint256[2] memory grossCurrent;
-                (grossCurrent[0], grossCurrent[1]) = SFPM.getAccountPremium(
+            // We need to adjust the grossPremiumLast value such that the result of
```
dyedm1: Might want to move this explanation block down below L1749 where the actual math occurs, it's a bit early here

contracts/PanopticPool.sol
```diff
@@ -1939,12 +1906,26 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+                        tokenId,
+                        leg,
+                        positionSize
+                    );
+                    positionLiquidity = liquidityChunk.liquidity();
 
-                // new totalLiquidity (total sold) = removedLiquidity + netLiquidity (T - R)
-                uint256 totalLiquidity = _getTotalLiquidity(tokenId, leg);
+                    // new totalLiquidity (total sold) = removedLiquidity + netLiquidity (T - R)
+                    //uint256 totalLiquidity = _getTotalLiquidity(tokenId, leg);
```
```suggestion

```

contracts/PanopticPool.sol
```diff
@@ -2034,9 +2015,18 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+
             unchecked {
                 ++leg;
             }
         }
+
+        // reset balances and delete stored option data
+        s_positionBalance[owner][tokenId] = LeftRightUnsigned.wrap(0);
+
+        // Update the position list hash (hash = XOR of all keccak256(tokenId)). Remove hash by XOR'ing again
```
dyedm1: This comment makes it sound like it's adding the position to the hash, we should make it clear that this particular XOR removes an existing position from the hash and decreases the position count.

contracts/PanopticPool.sol
```diff
@@ -1725,61 +1662,91 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+            // new totalLiquidity (total sold) = removedLiquidity + netLiquidity (R + N)
+            uint256 totalLiquidity = _checkLiquiditySpread(
+                tokenId,
+                leg,
+                liquidityChunk.tickLower(),
+                liquidityChunk.tickUpper(),
+                isLong == 0 ? MAX_SPREAD : uint64(Math.min(effectiveLiquidityLimitX32, MAX_SPREAD))
+            );
 
+            // if position is short, update gross premia accumulator
```
```suggestion
            // if position is short, adjust `grossPremiumLast` upward to account for the increase in short liquidity
```
It's not really an accumulator per se

contracts/PanopticPool.sol
```diff
@@ -1725,61 +1662,91 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+                    // update the premium accumulators
+                    s_options[msg.sender][tokenId][leg] = LeftRightUnsigned
+                        .wrap(0)
+                        .toRightSlot(uint128(grossCurrent0))
+                        .toLeftSlot(uint128(grossCurrent1));
+                }
+            }
+
+            // verify base Liquidity limit only if new position is long
+            // if position is long, ensure that new mints don't deplete strike beyond MAX_SPREAD
```
```suggestion
            // if position is long, ensure that removed liquidity does not deplete strike beyond min(MAX_SPREAD, user-provided effectiveLiquidityLimit)
```

contracts/PanopticPool.sol
```diff
@@ -1725,61 +1662,91 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+            // new totalLiquidity (total sold) = removedLiquidity + netLiquidity (R + N)
+            uint256 totalLiquidity = _checkLiquiditySpread(
+                tokenId,
+                leg,
+                liquidityChunk.tickLower(),
+                liquidityChunk.tickUpper(),
+                isLong == 0 ? MAX_SPREAD : uint64(Math.min(effectiveLiquidityLimitX32, MAX_SPREAD))
```
dyedm1, line 1710: If you move the `SFPM.getAccountLiquidity` call out of `_checkLiquiditySpread`, we would avoid the extra overhead of checking the spread on minting short legs and burning long legs

contracts/PanopticPool.sol
```diff
@@ -1725,61 +1662,91 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-                    0
+                    isLong
                 );
+                {
+                    // update the premium accumulators
+                    s_options[msg.sender][tokenId][leg] = LeftRightUnsigned
+                        .wrap(0)
+                        .toRightSlot(uint128(grossCurrent0))
+                        .toLeftSlot(uint128(grossCurrent1));
+                }
```
```suggestion
                
                s_options[msg.sender][tokenId][leg] = LeftRightUnsigned
                    .wrap(0)
                    .toRightSlot(uint128(grossCurrent0))
                    .toLeftSlot(uint128(grossCurrent1));                
```

contracts/PanopticPool.sol
```diff
@@ -1725,61 +1662,91 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-    /// @notice Adds collected tokens to settled accumulator and adjusts grossPremiumLast for any liquidity added.
-    /// @param tokenId The option position that was minted
-    /// @param collectedByLeg The amount of tokens collected in the corresponding chunk for each leg of the position
-    /// @param positionSize The size of the position, expressed in terms of the asset
+    /// @notice Adds collected tokens to settled accumulator and adjusts grossPremiumLast for any liquidity added
+    /// @dev Always called after `mintTokenizedPosition`
+    /// @param tokenId The option position that was minted.
+    /// @param collectedByLeg The amount of tokens collected in the corresponding chunk for each leg of the position.
+    /// @param positionSize The size of the position, expressed in terms of the asset.
+    /// @param effectiveLiquidityLimitX32 Maximum amount of "spread" defined as removedLiquidity/netLiquidity for a new position.
```
```suggestion
    /// @notice Adds collected tokens to `s_settledTokens` and adjusts `s_grossPremiumLast` for any liquidity added.
    /// @dev Always called after `mintTokenizedPosition`.
    /// @param tokenId The option position that was minted
    /// @param collectedByLeg The amount of tokens collected in the corresponding chunk for each leg of the position
    /// @param positionSize The size of the position, expressed in terms of the asset
    /// @param effectiveLiquidityLimitX32 Maximum amount of "spread" defined as `removedLiquidity/netLiquidity`
```

contracts/PanopticPool.sol
```diff
@@ -1559,6 +1491,9 @@ contract PanopticPool is ERC1155Holder, Multicall {
         // the effective liquidity measures how many times more the newly added liquidity is compared to the existing/base liquidity
         if (effectiveLiquidityFactorX32 > uint256(effectiveLiquidityLimitX32))
             revert Errors.EffectiveLiquidityAboveThreshold();
+
+        // returns the totalLiquidity
+        return totalLiquidity;
```
```suggestion
```

A return statement is not required for named return variables (though, if we move out `SFPM.getAccountLiquidity` as suggested in an earlier comment, this is a moot point)

contracts/PanopticPool.sol
```diff
@@ -1536,7 +1467,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
         int24 tickLower,
         int24 tickUpper,
         uint64 effectiveLiquidityLimitX32
-    ) internal view {
+    ) internal view returns (uint256 totalLiquidity) {
```
dyedm1, line 1466: natspec (not needed if we move the liquidity query out)

contracts/PanopticPool.sol
```diff
@@ -824,36 +781,10 @@ contract PanopticPool is ERC1155Holder, Multicall {
             owner
         );
 
-        // erase position data
-        _updatePositionDataBurn(owner, tokenId);
-
+        // emit event
```
```suggestion
```

contracts/PanopticPool.sol
```diff
@@ -665,7 +665,13 @@ contract PanopticPool is ERC1155Holder, Multicall {
         (LeftRightUnsigned[4] memory collectedByLeg, LeftRightSigned totalSwapped) = SFPM
             .mintTokenizedPosition(tokenId, positionSize, tickLimitLow, tickLimitHigh);
 
-        _updateSettlementPostMint(tokenId, collectedByLeg, positionSize);
+        // update premium settlement info
```
```suggestion
```

contracts/PanopticPool.sol
```diff
@@ -644,13 +642,15 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @notice Move all the required liquidity to/from the AMM and settle any required collateral deltas.
     /// @param tokenId The option position to be minted
     /// @param positionSize The size of the position, expressed in terms of the asset
-    /// @param tickLimitLow The lower bound of an acceptable open interval for the ending price
-    /// @param tickLimitHigh The upper bound of an acceptable open interval for the ending price
-    /// @return Packing of the pool utilization (how much funds are in the Panoptic pool versus the AMM pool) at the time of minting,
+    /// @param effectiveLiquidityLimitX32 Maximum amount of "spread" defined as removedLiquidity/netLiquidity for a new position.
+    /// @param tickLimitLow The lower slippage limit on the tick.
+    /// @param tickLimitHigh The upper slippage limit on the tick.
+    /// @return poolUtilizations Packing of the pool utilization (how much funds are in the Panoptic pool versus the AMM pool) at the time of minting,
```
```suggestion
    /// @param effectiveLiquidityLimitX32 Maximum amount of "spread" defined as `removedLiquidity/netLiquidity` 
    /// @param tickLimitLow The lower bound of an acceptable open interval for the ending price
    /// @param tickLimitHigh The upper bound of an acceptable open interval for the ending price
    /// @return Packing of the pool utilization (how much funds are in the Panoptic pool versus the AMM pool) at the time of minting,
```

contracts/PanopticPool.sol
```diff
@@ -1939,12 +1906,26 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+                {
+                    LiquidityChunk liquidityChunk = PanopticMath.getLiquidityChunk(
+                        tokenId,
+                        leg,
+                        positionSize
+                    );
+                    positionLiquidity = liquidityChunk.liquidity();
 
-                // new totalLiquidity (total sold) = removedLiquidity + netLiquidity (T - R)
-                uint256 totalLiquidity = _getTotalLiquidity(tokenId, leg);
```
dyedm1, line 1947: The now unused `_getTotalLiquidity` function can be removed (but we can use it if moving `SFPM.getAccountLiquidity` back out of `_checkLiquiditySpread`).

dyedm1 (dismissed):
LGTM; few suggestions and nits

contracts/PanopticPool.sol
guil-lambert: it does cause stack-too-deep :(

contracts/PanopticPool.sol
guil-lambert: I might keep this in. That extra overhead for the spread calculation is just two math operations: 
- `effectiveLiquidityFactorX32 = (uint256(removedLiquidity) * 2 ** 32) / netLiquidity;`
- `if (effectiveLiquidityFactorX32 > uint256(effectiveLiquidityLimitX32))`

It also acts as a safety check in case the effective liquidity factor is indeed larger than `MAX_SPREAD`, not sure if that's possible but that could be seen as a type of consistency+sanity check. 

But I did refactor `_checkLiquiditySpread`, so now we're only calling getAccountLiquidity from only one location in `_getLiquidities` (commit 5499428b7fd714993d5b589c02b4f13f277c5766)

contracts/PanopticPool.sol
guil-lambert: I did refactor `_checkLiquiditySpread`, so now we're only calling getAccountLiquidity from only one location in `_getLiquidities` (commit 5499428b7fd714993d5b589c02b4f13f277c5766)

PR #75, opened 2024-06-03
Feat: expand echidna tests on the CollateralTracker

flanagansteve:
- test CT.deposit and mint:
  - Balances of underlying asset change appropriately - depositor has less of the underlying asset, pool has more
  - Depositor receives correct amount of shares (aligning with previewDeposit)
  - Can never deposit greater than 2^104 of the underlying asset, nor mint > 2^104 shares
  - Can never deposit nor mint an amount greater than what you actually own in the underlying asset
- test CT.withdraw (both versions) and redeem:
  - Balances of underlying asset change appropriately - withdrawer has more of the underlying asset, pool has less
  - Depositor is deducted the correct amount of shares (aligning with previewWithdraw)
  - Cannot withdraw or redeem when you have open positions, unless through the designated overloaded withdraw for use with open positions
  - When withdrawing with open positions, withdrawer is solvent before and after the withdrawal, if withdrawing an amount such that your balance in the CT won't fall below the margin call threshold. Reverts are strictly only due to failing a solvency check after withdrawing the supplied `amountToWithdraw`.
- test CT.transfer and transferFrom:
  - Cannot transfer your shares when you have open positions
  - Balances of shares change appropriately - sender is deducted and receiver is accredited the correct amount
- test invariants about the state:
  - The CollateralTrackers internal counter of the pool assets never exceeds the contract's actual balance + 1
  - No user can ever withdraw greater than the CollateralTracker's internally-tracked poolAssets
  - No user can ever withdraw/redeem/transfer more than the balance the CT accredits them
  - pool utilisation never exceeds 100%


contracts/fuzz/FuzzDeployments.sol
```diff
@@ -908,44 +908,137 @@ contract FuzzDeployments is FuzzHelpers {
...
+            // fuzz a fraction of the total balance to try and withdraw
+            if (fuzzNumerator > fuzzDenominator)
+                (fuzzNumerator, fuzzDenominator) = (fuzzDenominator, fuzzNumerator);
+
+            // attempt a full withdrawal every 3rd block, and a self-withdrawal every 5th block, to ensure we're testing those cases
+            if (block.number % 3 == 0) fuzzNumerator = fuzzDenominator;
+            if (block.number % 5 == 0) recipient = msg.sender;
+
+            // TODO: i don't think we have to care about overflows here with how bal0 / bal1 get generated, but check that.
+            uint256 fuzzedSharesToRedeem0 = (shareBal0 * fuzzNumerator) / fuzzDenominator;
```
flanagansteve, line 947: This is how i've been fuzzing fractions of a user's balance to collateral-remove. We could use `bound` instead, or a single fuzzed x64Multiplier instead of a fuzzed numerator + denominator like i use here (which requires slightly more manipulation - see my swap such that the greater value is the denominator on line 940, for example). Let me know if you'd like me to switch to that (and likely also switch `invariant_collateral_removal_via_transfer` in the same way)

edit: i ended up switching to `bound` below as i felt like there wasn't a big advantage in proportionality for the other tests. Let me know what you think - would be plenty easy to switch this to bound or something like that.

contracts/fuzz/FuzzDeployments.sol
```diff
@@ -1003,6 +1139,406 @@ contract FuzzDeployments is FuzzHelpers {
...
+            } catch {
+                if (
+                    collToken.convertToShares(ct_s_poolAssets + amount_over) >
+                    collToken.balanceOf(owner)
+                ) {
+                    emit LogString(
+                        "invariant_no_withdrawal_gt_pool_assets succeeded because user didnt have enough shares to attempt overwithdrawal"
+                    );
+                } else {
+                    // NOTE: we could add a deal of the collToken.asset() if we wanted to ensure we hit this case more often
```
flanagansteve, line 1235: Let me know if you think we should do this - we could also grep the logs after running it for a while to see how often we successfully prove that we enforce limits in withdraw/deposit etc, compared to just generating withdraw/deposit calls that exceed user balance

contracts/fuzz/FuzzDeployments.sol
```diff
@@ -1003,6 +1134,402 @@ contract FuzzDeployments is FuzzHelpers {
...
+                    emit LogString(
+                        "invariant_no_withdrawal_gt_pool_assets succeeded, possibly because we correctly enforced a max withdrawal of ct_s_poolAssets"
+                    );
+                }
+            }
+        } else {
+            TokenId[] memory withdrawersOpenPositions = userPositions[owner];
+            try
+                collToken.withdraw(
```
flanagansteve, line 1264: When i run this locally, i had just pasted in the overloaded withdraw on my local copy of CollateralTracker.sol. Depending on our merge strategy, i could either cherrypick the commit that added that method and insert it into this branch, or i could just merge my branch into someone else's that already has it

dyedm1: It's fine for now, I think we'll just end up overwriting whatever version of the contracts we have in the tests when we merge it all together

contracts/fuzz/FuzzDeployments.sol
```diff
@@ -908,44 +908,175 @@ contract FuzzDeployments is FuzzHelpers {
...
+            if (block.number % 5 == 0) recipient = msg.sender;
+
+            uint256 fuzzedSharesToRedeem0 = (shareBal0 * fuzzNumerator) / fuzzDenominator;
+            uint256 fuzzedSharesToRedeem1 = (shareBal1 * fuzzNumerator) / fuzzDenominator;
+
+            uint256 fuzzedAssetsToWithdraw0 = (assetBal0 * fuzzNumerator) / fuzzDenominator;
+            uint256 fuzzedAssetsToWithdraw1 = (assetBal1 * fuzzNumerator) / fuzzDenominator;
+
+            if (fuzzedAssetsToWithdraw0 > 0) {
+                try collToken0.redeem(fuzzedSharesToRedeem0, recipient, msg.sender) {
```
dyedm1, line 955: I think you need to be pranking for these calls, otherwise they come from the contract (so this test will always work). Did you put an invariant in the catch case to ensure it was being hit with legit reasons?

contracts/fuzz/FuzzDeployments.sol
dyedm1: AFAIK there's no way to get logs (it would be a huge amount of logs...). You would have to add a temp invariant there. Why don't we add a function that randomly sends tokens to the panoptic pool just to make sure there's a possibilit.

contracts/fuzz/FuzzDeployments.sol
```diff
@@ -1003,6 +1134,402 @@ contract FuzzDeployments is FuzzHelpers {
...
+        address recipient,
+        uint256 amountOver
+    ) internal {
+        (uint256 ct_s_poolAssets, , ) = collToken.getPoolData();
+        amountOver = bound(amountOver, 1, type(uint256).max - ct_s_poolAssets);
+
+        uint256 numOfPositions = panopticPool.numberOfPositions(owner);
+        if (numOfPositions == 0) {
+            hevm.prank(owner);
+            // every other attempt, make it a non-owner call:
```
dyedm1: btw, the block.number is advanced randomly and independently from other calls, so this doesn't really mean "every other attempt". It's probably fine but I've been using input params instead

contracts/fuzz/FuzzDeployments.sol
```diff
@@ -1050,113 +1577,313 @@ contract FuzzDeployments is FuzzHelpers {
...
+        (int128 premium0, int128 premium1, uint256[2][] memory positions) = panopticPool
+            .calculateAccumulatedFeesBatch(withdrawer, false, userPositions[withdrawer]);
+        LeftRightUnsigned tokenData = collToken.getAccountMarginDetails(
+            withdrawer,
+            curTick,
+            positions,
+            isToken0 ? premium0 : premium1
+        );
+        uint256 marginCallThreshold = tokenData.leftSlot();
+        return withdrawerAsssetsInCT - marginCallThreshold;
```
dyedm1: I'm a little confused as to how this works. The validation is done on the cross-collateral req+balance, but this seems to only use one collateral token + one side of the collateral req to determine solvency?

Getting the amount withdrawable could be a bit tricky, so it might be easier just to fuzz the amount between 0 and their balance, use the validate function to check that, and have a separate invariant for the validate function

contracts/fuzz/FuzzDeployments.sol
```diff
@@ -1050,113 +1577,313 @@ contract FuzzDeployments is FuzzHelpers {
...
         } else {
-            amount = bound(amount, 1, collToken1.balanceOf(msg.sender));
-            hevm.prank(msg.sender);
-            collToken1.redeem(amount, msg.sender, msg.sender);
+            emit LogString("Attempting to transfer/transferFrom token1");
+            _transfer_and_check(collToken1, shares, sender, recipient, transferFromSender);
+        }
+    }
+
+    function _transfer_and_check(
```
dyedm1, line 2044: can we extend this to transfers that fail for other reasons (i.e. not having enough shares etc.?

dyedm1 (changes requested):
Looks solid overall. I had a few comments on things that I think might not be working correctly. There is a rather large issue that I'll bring up here -- you don't appear to be pranking correctly before sending calls, which would cause a lot of them to revert unexpectedly or for a different reason than expected (because the calls come from the fuzzDeployments contract and not the account you're using for all your calculations). 

This might be a bit counterintuitive, but the `hevm.prank()` function *only lasts for one call*. This *includes* view/static/query calls.There is an analogue we use in Foundry/Anvil (`vm.startPrank()`) that lasts until otherwise specified, but the standard `vm.prank()` function there does the same thing as well.

contracts/fuzz/FuzzDeployments.sol
flanagansteve: Aha, my bad. I think this was also an issue in the ToB implementation for the original version of this method. I added the prank now and will add a temp invariant to ensure we hit the revert as expected.

contracts/fuzz/FuzzDeployments.sol
flanagansteve: We ensure that users cannot transfer more than their own balance of shares in `invariant_never_allow_overremoval` - it calls the helper `_attempt_overtransfer`, in which we purposely attempt to transfer a fuzzed `amountOver` over a user's share balance.

Is that what you're looking for? Or do you specifically prefer to widen the possible amounts to transfer in `transfer_ct_shares` and just handle reverts due to transfer amounts greater than balance?

contracts/fuzz/FuzzDeployments.sol
flanagansteve: Confirmed reverted as expected!

contracts/fuzz/FuzzDeployments.sol
dyedm1: > Is that what you're looking for

Yep - looks like the negative cases are covered pretty well elsewhere :)

> Or do you specifically prefer to widen the possible amounts to transfer in transfer_ct_shares and just handle reverts due to transfer amounts greater than balance?

I think this would probably be easier, but either way works!

contracts/fuzz/FuzzDeployments.sol
flanagansteve: I was going in to make this change. It was easy to add deal_USDC or deal_WETH as appropriate to ensure depositors/minters had enough of the underlying asset to attempt a too-large mint or deposit. I just added a commit that adds those `deal`s and removes the NOTEs that were there before.

However, there's still 4 places we may end up "succeeding" on withdrawal/redemption invariants such as not being able to withdraw > pool assets, but not because of our logic succeeding necessarily - more likely, just because the user doesn't have enough deposited to make the withdrawal (which makes sense):

```
if (
        collToken.convertToShares(ct_s_poolAssets + amountOver) >
        collToken.balanceOf(owner)
) {
    emit LogString(
        "invariant_no_withdrawal_gt_pool_assets succeeded because user didnt have enough shares to attempt overwithdrawal"
    );
}
```

I think your recommendation for this is:

> Why don't we add a function that randomly sends tokens to the panoptic pool just to make sure there's a possibilit.

Which i don't quite understand - are you saying try to deposit some amount > `ct_s_poolAssets` at random then withdraw after? wouldn't pool assets just (appropriately) increase when you deposit?

My approach was going to be to use `deal_Generic` like we do in `editCollateral`, as that seems to use `hevm.store` directly and could go around any accounting we do that increments `s_poolAssets`

contracts/fuzz/FuzzDeployments.sol
flanagansteve: Input params are definitely better. I replaced every place that i do `block.number % 2` with a fuzzed bool, and the other ones - % 3, % 5, and % 4 - with a modulo on a fuzzed uint256 input parameter instead of block number.

contracts/fuzz/FuzzDeployments.sol
flanagansteve: awesome!

contracts/fuzz/FuzzDeployments.sol
flanagansteve: After discussion in standup - the approach in this PR is now to:

- in the `catch` case after calling withdraw-with-open-positions, check if the attempted withdrawal would indeed have caused insolvency; if so, we reverted for a good reason and no invariants have been violated
- in the inverse test that proves we revert when you try to withdraw too much (via  the overloaded withdraw) - we just try to withdraw the user's full balance

contracts/fuzz/FuzzDeployments.sol
flanagansteve: After discussion in standup - i added a `deal_to_contracts` method which should ensure we get sequences where the contracts are randomly dealt some extra tokens, to see if this breaks things like the protection against withdrawing greater than the s_poolAssets

contracts/fuzz/FuzzDeployments.sol
flanagansteve: ... and now the `catch` case that checks for solvency correctly deducts your `balanceThreshold` by the amount & type of asset you're withdrawing!

PR #76, opened 2024-06-05
[Draft] Hardened solvency checks

guil-lambert:
# Hardened Solvency Checks

This PR addresses one UX issue and another potential security issue.

## StaleTWAP()
First, we are currently performing a staleness check on the `twapTick` vs `currentTick` during liquidations and revert if those two diverge by more than 5%:

https://github.com/panoptic-labs/panoptic-v1-core-private/blob/53c45f1f18365069a21dde23b1bcd7a9c0b640da/contracts/PanopticPool.sol#L1003-L1004

This potentially impacts UX because an account could be unliquidatable under choppy price moves (ie. continuously up/down by 5% in a 10mins window), and liquidation would only proceed if the price happens to be within 5% of the TWAP, which could lead to execution at an arbitrary price and not the current one.

This PR solves this by ensuring that an additional solvency check at the `currentTick` is performed when the price diverges by more than 5%:

https://github.com/panoptic-labs/panoptic-v1-core-private/blob/357b46e139642649344efcdc98d3becee19df890/contracts/PanopticPool.sol#L1059-L1068

This ensure that accounts could still be liquidated even of there is a recent 5% price move AND it is also liquidatable at the TWAP tick and the current Tick

## Covered Position Mandate

This PR also address a potential issue where a rapid price move larger than 20% leads to users trading many options using the few blocks of "free" collateral.

That is, create positions when `fastOracleTick` hasn't quite caught up to `currentTick`. In this scenario, users could mint options with ITM amounts, receive tokens due to the intrinsic value, and then withdraw those funds because the collateral check is done at the `fastOracleTick` and those positions appear OTM

To solve this, we mandate that all newly minted options to be `covered` when the recent price deviated by more than 20%. This means selling and/or buying ITM options would require the user to own the itm amounts in their account.

This is implemented here, where the `tickLimit` order is changed to small-to-large if the price deviates too much from the `slowOracleTick`, hardcoded to be median tick.

https://github.com/panoptic-labs/panoptic-v1-core-private/blob/357b46e139642649344efcdc98d3becee19df890/contracts/PanopticPool.sol#L659-L673

## 100% Collateralization Mandate

This one is further hardening the solvency of the protocol after large/sudden price moves that exceed 20%. We do this by mandating that, in addition to the covered minting, we also change the collateral requirement of all short options to 100% collateralized. 

Increasing the collateral requirement to 100% does not change the funds required inside the account for the covered mint mandate --all transactions that have enough funds will still succeed.

The additional collateral requirement criteria (ie. `safeMode = true`) means that those positions will also be fully collateralized and will never lead to any protocol loss. 

This would mean users that want to capitalize on sudden price moves could still mint options, but those options would not add any extra risk to the protocol. If the price stabilizes at a later time, those users could close+redeploy their position at a lower collateral requirement provides that the >20% move condition is not satisfied.

This full collateralization requirement is implemented by passing a `safeMode=true` flag in `collateralTracker.takeCommissionAddData`:

https://github.com/panoptic-labs/panoptic-v1-core-private/blob/357b46e139642649344efcdc98d3becee19df890/contracts/CollateralTracker.sol#L1052

## Review conditions

This is changing the code after our intended code freeze... The additional checks should make the collateral requirements and options minting _safer_ by adding more checks + requirements, but that'd still need to be verified before merging into the frozen codebase. 

## Alternatives

An alternative check in `liquidate` for tick staleness could be based on the vector displacement instead of the individual deviations. Could be safer, but I haven't looked at it too deeply:

```        
if (
    ((int256(fastOracleTick) - slowOracleTick)**2 + 
    (int256(fastOracleTick) - currentTick)**2 + 
    (currentTick - slowOracleTick)**2) >  MAX_SLOW_FAST_DELTA**2
) {
```

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1425,6 +1425,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                         // position is in-the-money, collateral requirement = amountMoved*(1-SRC)*(scaleFactor-ratio)/(scaleFactor+1) + SCR*amountMoved
                         required += c3;
                     }
+                    // cannot require more than the notional value of that position
+                    required = Math.max(required, amountMoved);
                 }
             }
         }
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1010,12 +1010,14 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param longAmount The amount of longs
     /// @param shortAmount The amount of shorts
     /// @param swappedAmount The amount of tokens moved during creation of the option position
+    /// @param safeMode whether to mandate 100% collateralization
     /// @return utilization The final utilization of the collateral vault
     function takeCommissionAddData(
         address optionOwner,
         int128 longAmount,
         int128 shortAmount,
-        int128 swappedAmount
+        int128 swappedAmount,
+        bool safeMode
     ) external onlyPanopticPool returns (int256 utilization) {
         unchecked {
             // current available assets belonging to PLPs (updated after settlement) excluding any premium paid
@@ -1047,7 +1049,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             s_poolAssets = uint128(uint256(updatedAssets));
             s_inAMM = uint128(uint256(int256(uint256(s_inAMM)) + (shortAmount - longAmount)));
 
-            utilization = _poolUtilization();
+            utilization = safeMode ? DECIMALS_128 : _poolUtilization();
         }
     }
 
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1426,7 +1426,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                         required += c3;
                     }
                     // cannot require more than the notional value of that position
-                    required = Math.max(required, amountMoved);
+                    required = Math.min(required, amountMoved);
                 }
             }
         }
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1074,7 +1074,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
         int256 liquidationBonus0;
         int256 liquidationBonus1;
-        int24 finalTick;
         {
             LeftRightSigned netExchanged;
             LeftRightSigned[4][] memory premiasByLeg;
@@ -1091,8 +1090,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 positionIdList
             );
 
-            (, finalTick, , , , , ) = s_univ3pool.slot0();
-
             LeftRightSigned collateralRemaining;
             // compute bonus amounts using latest tick data
             (liquidationBonus0, liquidationBonus1, collateralRemaining) = PanopticMath
@@ -1100,7 +1097,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
                     tokenData0,
                     tokenData1,
                     Math.getSqrtRatioAtTick(twapTick),
-                    Math.getSqrtRatioAtTick(finalTick),
+                    Math.getSqrtRatioAtTick(currentTick),
                     netExchanged,
                     premia
                 );
@@ -1116,7 +1113,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
             // reusing variables to save stack space; netExchanged = deltaBonus0, premia = deltaBonus1
             address _liquidatee = liquidatee;
             TokenId[] memory _positionIdList = positionIdList;
-            int24 _finalTick = finalTick;
             int256 deltaBonus0;
             int256 deltaBonus1;
             (deltaBonus0, deltaBonus1) = PanopticMath.haircutPremia(
@@ -1126,7 +1122,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 collateralRemaining,
                 s_collateralToken0,
                 s_collateralToken1,
-                Math.getSqrtRatioAtTick(_finalTick),
+                Math.getSqrtRatioAtTick(currentTick),
                 s_settledTokens
             );
 
@@ -1156,8 +1152,8 @@ contract PanopticPool is ERC1155Holder, Multicall {
             !_checkSolvencyAtTick(
                 msg.sender,
                 positionIdListLiquidator,
-                finalTick,
-                finalTick,
+                currentTick,
+                currentTick,
                 BP_DECREASE_BUFFER
             )
         ) revert Errors.NotEnoughCollateral();
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1307,6 +1307,13 @@ contract PanopticPool is ERC1155Holder, Multicall {
         return _checkCrossBalances(account, atTick, positionBalanceArray, portfolioPremium, buffer);
     }
 
+    /// @notice Check whether a the balances of an account is solvent at a given `atTick` with a collateral requirement of `buffer`/10_000 multiplied by the requirement of `positionBalanceArray`.
+    /// @param account The account to check solvency for
+    /// @param atTick The tick to check solvency at
+    /// @param positionBalanceArray A list of balances and pool utilization for each position, of the form [[tokenId0, balances0], [tokenId1, balances1], ...]
+    /// @param portfolioPremium The computed premia of the user's positions, where premia contains the accumulated premia for token0 in the right slot and for token1 in the left slot
+    /// @param buffer The buffer to apply to the collateral requirement
+    /// @return Whether the account is solvent at the given tick
     function _checkCrossBalances(
         address account,
         int24 atTick,
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -654,6 +654,22 @@ contract PanopticPool is ERC1155Holder, Multicall {
         int24 tickLimitLow,
         int24 tickLimitHigh
     ) internal returns (uint128) {
+        // check if the price has deviated too much recently.
+        {
+            (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
+            uint256 medianData = s_miniMedian;
+            int24 medianTick = (int24(
+                uint24(medianData >> ((uint24(medianData >> (192 + 3 * 3)) % 8) * 24))
+            ) + int24(uint24(medianData >> ((uint24(medianData >> (192 + 3 * 4)) % 8) * 24)))) / 2;
+
+            // If ticks have recently deviated more than +/- 20%, enforce covered mints
+            if (Math.abs(currentTick - medianTick) > MAX_SLOW_FAST_DELTA) {
+                if (tickLimitLow > tickLimitHigh) {
+                    (tickLimitLow, tickLimitHigh) = (tickLimitHigh, tickLimitLow);
+                }
+            }
+        }
+
         (LeftRightUnsigned[4] memory collectedByLeg, LeftRightSigned totalSwapped) = SFPM
             .mintTokenizedPosition(tokenId, positionSize, tickLimitLow, tickLimitHigh);
 
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -654,22 +654,10 @@ contract PanopticPool is ERC1155Holder, Multicall {
         int24 tickLimitLow,
         int24 tickLimitHigh
     ) internal returns (uint128) {
-        bool safeMode;
-        // check if the price has deviated too much recently.
-        {
-            (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
-            uint256 medianData = s_miniMedian;
-            int24 medianTick = (int24(
-                uint24(medianData >> ((uint24(medianData >> (192 + 3 * 3)) % 8) * 24))
-            ) + int24(uint24(medianData >> ((uint24(medianData >> (192 + 3 * 4)) % 8) * 24)))) / 2;
-
-            // If ticks have recently deviated more than +/- 20%, enforce covered mints
-            if (Math.abs(currentTick - medianTick) > MAX_SLOW_FAST_DELTA) {
-                safeMode = true;
-                if (tickLimitLow > tickLimitHigh) {
-                    (tickLimitLow, tickLimitHigh) = (tickLimitHigh, tickLimitLow);
-                }
-            }
+        bool safeMode = isSafeMode();
+        // if safeMode, enforce covered deployment
+        if (safeMode && (tickLimitLow > tickLimitHigh)) {
+            (tickLimitLow, tickLimitHigh) = (tickLimitHigh, tickLimitLow);
         }
 
         (LeftRightUnsigned[4] memory collectedByLeg, LeftRightSigned totalSwapped) = SFPM
@@ -978,6 +966,14 @@ contract PanopticPool is ERC1155Holder, Multicall {
             LeftRightSigned paidAmounts
         )
     {
+        {
+            bool safeMode = isSafeMode();
+            // if safeMode, enforce covered at assignment
+            if (safeMode && (tickLimitLow > tickLimitHigh)) {
+                (tickLimitLow, tickLimitHigh) = (tickLimitHigh, tickLimitLow);
+            }
+        }
+
         (LeftRightUnsigned[4] memory collectedByLeg, LeftRightSigned totalSwapped) = SFPM
             .burnTokenizedPosition(tokenId, positionSize, tickLimitLow, tickLimitHigh);
 
@@ -1071,6 +1067,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
         // Perform the specified delegation from `msg.sender` to `liquidatee`
         // Works like a transfer, so the liquidator must possess all the tokens they are delegating, resulting in no net supply change
         // If not enough tokens are delegated for the positions of `liquidatee` to be closed, the liquidation will fail
+        // TODO: since all options are closed using the covered flag, we could compute the delegation amounts directly here.
         s_collateralToken0.delegate(msg.sender, liquidatee, delegations.rightSlot());
         s_collateralToken1.delegate(msg.sender, liquidatee, delegations.leftSlot());
 
@@ -1365,6 +1362,24 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 Math.mulDivRoundingUp(uint256(tokenData1.leftSlot()), 2 ** 96, sqrtPriceX96) +
                 Math.mulDiv96RoundingUp(tokenData0.leftSlot(), sqrtPriceX96);
         }
+    }
+
+    /// @notice Checks whether the current tick has deviated too much from the slow oracle media tick
+    /// @return safeMode a boolean flag, triggers safe more when true where all newly minted positions must be fully collateralized
+    function isSafeMode() internal returns (bool safeMode) {
+        // check if the price has deviated too much recently.
+        (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
+        uint256 medianData = s_miniMedian;
+        int24 medianTick = (int24(
+            uint24(medianData >> ((uint24(medianData >> (192 + 3 * 3)) % 8) * 24))
+        ) + int24(uint24(medianData >> ((uint24(medianData >> (192 + 3 * 4)) % 8) * 24)))) / 2;
+
+        // If ticks have recently deviated more than +/- 20%, enforce covered mints
+        if (Math.abs(currentTick - medianTick) > MAX_SLOW_FAST_DELTA) {
+            safeMode = true;
+        }
+
+        return (safeMode);
     }
 
     /*//////////////////////////////////////////////////////////////
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -654,6 +654,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
         int24 tickLimitLow,
         int24 tickLimitHigh
     ) internal returns (uint128) {
+        bool safeMode;
         // check if the price has deviated too much recently.
         {
             (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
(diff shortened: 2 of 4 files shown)


dyedm1:
> This ensure that accounts could still be liquidated even of there is a recent 5% price move AND it is also liquidatable at the TWAP tick and the current Tick

This is kind of orthogonal to the stale TWAP restriction. As I mention in a chain of comments on #63: https://github.com/panoptic-labs/panoptic-v1-core-private/pull/63#issuecomment-2136711118, the purpose of this is actually to prevent liquidatees from being forced to exercise long positions at an unfavorable price (i.e. the exploit you brought up in #63 and what the "loss compensation" mechanism covers in force exercises (finalized in #67)). 

However, as discussed in that comment linked above, there is a different path toward removing the stale TWAP restriction altogether. I ultimately decided not to move forward with it given that it is not an *essential* improvement and we wanted to minimize the re-audit diff as much as possible.

> That is, create positions when fastOracleTick hasn't quite caught up to currentTick. In this scenario, users could mint options with ITM amounts, receive tokens due to the intrinsic value, and then withdraw those funds because the collateral check is done at the fastOracleTick and those positions appear OTM

I think we should probably expand on what is *actually* happening in this scenario:

Panoptic is an undercollateralized lending market. Option sellers borrow tokens from PLPs to provide liquidity in Uniswap. Their debts are collateralized with the tokens in the seller's account as well as the LP position itself. 

In any lending market, it is essential to determine accurate values for a user's collateral; otherwise, they could abscond with the borrowed funds, leaving bad debt/protocol loss behind.

Let's run through the scenario:
1. Both the oracle and market prices for ETH begin at 2000
2. The market price moves to 1000 instantaneously (note: it is unimportant whether the current price telegraphs this move; the important part is that the current price moves to the new price in the near future and stays there for at least several minutes)
3. An opportunistic trader borrows 2000 USDC from PLPs under the terms that they will collateralize it with 400 USDC worth of collateral at the oracle price and a 1-tick-wide Uniswap LP position (size = 1ETH/2000 USDC) at a strike price of 2000. At any point, their collateral must be worth at least 120% of their debt.

This is where things get tricky. Ultimately, a similar financial transaction is occurring, but there's a couple different paths to get there depending on the initial state/preconditions and the set restrictions enforced by the protocol. I'll start by running through the variant you're trying to prevent in this PR:
1.  In this variant, the current price has followed or telegraphed other venues and moved to 1000 before any transactions occur.
2. Under the aforementioned terms, the user deposits 0.2 ETH of collateral and borrows the 2000 USDC.
3. They then swap 1000 USDC for 1 ETH and create the specified LP position. 
4. Now, the user's debt of 2000 USDC is collateralized by 1000 USDC, 0.2 ETH(=200USDC), and the LP position (currently worth 1ETH=1000USDC)
5. The oracle price is still 2000, so the protocol views the user as having 3400 USDC of collateral, despite the actual value being 2200 (already insolvent according to the terms of the loan). The collateral requirement is 120%, so the user's account is viewed as having 1000 USDC in surplus/withdrawable collateral.
6. The user withdraws their 1000 USDC collateral surplus. Now, their account is completely insolvent, with 1200 USDC worth of collateral held against a 2000 USDC debt (-800 USDC).

In effect, a user was allowed to borrow 2000 USDC while providing only 1200 USDC worth of tokens to collateralize the position, allowing them to make a profit of 800 USDC by simply allowing their loan to be liquidated. Let's try the same variant with your restrictions added:
1. no change
2. The user is required to be solvent at the current tick. They must now deposit 0.4 ETH (or 400 USDC) to collateralize their position.
3. no change
4. -0.1 ETH +(0.4 ETH | 400 USDC)
5. Collateral requirement = 2200 USDC, Account value (solvency oracle) = 2400 USDC, Account value (market) = 2400 USDC, Surplus collateral = 0 USDC
6. They withdraw 0 USDC in surplus collateral, causing no protocol loss and generating no profits.

Seems like they worked? What if a user tries to disable your restrictions by flash-swapping the current price back to 2000 before minting their position:
1. In this variant, the current price is either lagging other venues and is still at 2000-ε or has been temporarily swapped back to that price by the user. Additionally, that minting user has taken a flash loan or used their own inventory to JIT a great deal of liquidity into the pool around 2000.
2. no change (from original)
3. The user swaps 2000 USDC for 1 ETH (Assuming the position is 1-tick-wide and the current price is 2000-ε (where ε moves the price out of the position's range), adjust position/pool state accordingly for wider positions). Because they own most/all of the liquidity at that strike, they have effectively removed the 2000 USDC they borrowed from PLPs in exchange for 1 ETH.
4. Now, the user's debt of 2000 USDC is collateralized by 0.4 ETH(=200USDC) and the LP position (worth 1ETH=1000 USDC at market).
5. The user is viewed as having 2400 USDC of collateral at the oracle price. Their account is still considered solvent and will not be immediately liquidated, though they have no surplus collateral to withdraw).
6. The user removes their JIT liquidity and swaps the price back down. They also abandon their Panoptic account and leave it to be liquidated in a few minutes once the TWAP price catches up. In total, they have spent ~1200 USDC (acquiring 0.4 ETH collateral for 200 USDC + acquiring 1 ETH for 1000 USDC to provide liquidity for the swap + negligible swap execution/flashloan costs) and received 2000 USDC from the ITM swap, resulting in a profit of 800 USDC.

Even with those restrictions in place, the user was able to easily bypass them by manipulating the current price back and forcing the protocol to perform a swap at an unfavorable price. 

These exploits involve what is essentially the same transaction: The protocol lends the user 2000 USDC in exchange for 1.2 ETH in collateral. That collateral is worth 2400 USDC at the oracle price, but 1200 USDC at market prices. Without a manipulation-resistant indication of what the "market price" is (which is impossible to know without real-time access to all CEX and DEX data for a given pair), I don't see a way to prevent the protocol from underwriting that loan in some form or another. 

Given the ease with which these currentTick-derived restrictions can be bypassed, I don't think it makes any sense to add them to the protocol. Pretty much all lending protocols (excepting Ajna which is a truly oracle-free lending protocol with a novel design and failure modes unique to it) to date are vulnerable to the fundamental scenario described here.

Also not sure it's worth it to remove the stale TWAP restrictions given how short the TWAP period is, though there's a solution to that I'm open to implementing depending on security resources/timeline (which would be what is described in the comment linked above). For those smaller pools, I'm way more worried about the TWAP period being too short and causing liquidation cascades (if you have two consecutive blocks, you can create a manipulated observation in the second block, resulting in a manipulated TWAP price ten minutes later if no one swaps after that point).

guil-lambert:
>An opportunistic trader borrows 2000 USDC from PLPs under the terms that they will collateralize it with 400 USDC worth of collateral at the oracle price and a 1-tick-wide Uniswap LP position (size = 1ETH/2000 USDC) at a strike price of 2000. At any point, their collateral must be worth at least 120% of their debt.

The new restrictions forces the collateral to always be of the same type AND to be 100% collateralized. 

Let go through your example, which is selling a put option at the 2000 strike when the price is 1000 but the slow oracle price is still 2000 with a starting collateral balance of (0USDC, 0.2ETH):

>In this variant, the current price has followed or telegraphed other venues and moved to 1000 before any transactions occur.
>Under the aforementioned terms, the user deposits 0.2 ETH of collateral and borrows the 2000 USDC.
  - If the tick difference between the slow oracle tick and the current tick is more than 20%, then the user has sell covered positions
  - In other words, they need to have at least 1ETH in their account as well because the (tickLimitLow < tickLimitHigh) condition means there is no ITM swap
  - Let's say the collateral for the user is the bare minimum at (0USDC, 1ETH)

>They then swap 1000 USDC for 1 ETH and create the specified LP position.
  - Here, they can't do that: the 1 ETH has to come from their account because is forced to be covered
  - Net balance change for the user: (+2000USDC, -1ETH), and their collateral is (2000USDC, 0ETH)
  - _before, their initial balance was (0USDC, 0.2ETH) and became (+1000USDC, 0.2ETH)_
 
>Now, the user's debt of 2000 USDC is collateralized by 1000 USDC, 0.2 ETH(=200USDC), and the LP position (currently worth 1ETH=1000USDC)
  - The 2000 USDC debt is now collateralized by (2000 USDC, 0ETH) and the LP position

>The oracle price is still 2000, so the protocol views the user as having 3400 USDC of collateral, despite the actual value being 2200 (already insolvent according to the terms of the loan). The collateral requirement is 120%, so the user's account is viewed as having 1000 USDC in surplus/withdrawable collateral.
  - The protocol views the user as having 2000 USDC, it does not depend on price

>The user withdraws their 1000 USDC collateral surplus. Now, their account is completely insolvent, with 1200 USDC worth of collateral held against a 2000 USDC debt (-800 USDC).
  - Withdrawals that leaves the account solved at 2000 and 1000 is allowed, in this case 600 USDC can be withdrawn.

---
BTW, you're doing the collateral requirement in a weird manner. Here's how I see it: if a user sells a put option at the 2000 strike, the collateral requirement at price `S` is
```
requireCollateral = (0.2 + (1-S/K)) * positionSize = 1.2 * K - S (assuming 1 ETH contract)
```

So in the example above:
- collateralBalance = (0, 0.2ETH) or (1, 1ETH)
- required = (1.2*2000 - S, 0)
  - (400, 0) if evaluated at S = 2000
  - (1400, 0) if evaluated at S = 1000  
- Balance after mint:
  - (1000 USDC, 0.2) if not using covered minting
  - (2000 USDC, 0ETH) if using covered minting
- `collateralBalance - required`:
  - Not covered, at 2000: (1000 + 0.2 * 2000) - (400 + 0 * 2000) = 1000 USDC -> collateral in excess, can withdraw
  - Not covered, at 1000: (1000 + 0.2 * 1000) - (1400) = -200 USDC, undercollateralized!
  - Covered, at 2000: (2000) - (400) = 1600 USDC, still collateralized (solvency check here)
  - Covered, at 1000: (2000) - (1400) = 600 USDC, still collateralized (solvency check here)
    - Max withdraw: min(1600,600) = 600 USDC

---
>These exploits involve what is essentially the same transaction: The protocol lends the user 2000 USDC in exchange for 1.2 ETH in collateral. 
>Pretty much all lending protocols (excepting Ajna which is a truly oracle-free lending protocol with a novel design and failure modes unique to it) to date are vulnerable to the fundamental scenario described here.

Users cannot use ETH collateral for puts when the price deviates >20% because we mandate the positions to be covered when there is a large discrepancy between internal & current ticks 

Hence, we only allow USDC to collateralize USDC positions. The price is not even entering the solvency calculations and I'd say we are not vulnerable to that type of scenario anymore. 

---

`safeMode` Gadget: All of this is still further hardened by the safeMode gadget, which mandates that the new positions minted is also at 100% collateral requirement instead of 20% or (1.2 - price/strike)%.
 
So you can think of the redundant solvencyChecks at fast/slow/current ticks as a way to ensure that _previously deployed positions_ are not insolvent, and the new safeMode gadget ensures that the _new position_ is always solvent. 

So, in the example above, we adjust the required collateral:
- collateralBalance = (0, 0.2ETH) or (1, 1ETH)
- required = (1.2*2000 - S, 0)
  - ~~(400, 0)~~ (2000, 0) if evaluated at S = 2000
  - ~~(1400, 0)~~ (2000, 0) if evaluated at S = 1000  
- Balance after mint:
  - (1000 USDC, 0.2) if not using covered minting
  - (2000 USDC, 0ETH) if using covered minting
- `collateralBalance - required`:
  - Not covered, at 2000: (1000 + 0.2 * 2000) - (2000) = -600 USDC, undercollateralized!
  - Not covered, at 1000: (1000 + 0.2 * 1000) - (2000) = -800 USDC, undercollateralized!
  - Covered, at 2000: (2000) - (2000) = 0 USDC, still collateralized but cannot withdraw
  - Covered, at 1000: (2000) - (2000) = 0 USDC, still collateralized but cannot withdraw

dyedm1:
>The new restrictions forces the collateral to always be of the same type AND to be 100% collateralized.
>Let go through your example, which is selling a put option at the 2000 strike when the price is 1000 but the slow oracle price is still 2000 with a starting collateral balance of (0USDC, 0.2ETH):

Sorry, should have been a bit clearer on this. The first example I gave is the variant prevented by your restrictions, so you are correct that this situation is covered. Swaps are actually disabled by your restrictions so the particulars are a little different than the ones I listed (should have given an example without ITM swaps). Either way, your restrictions prevent the attack from occuring *when they are enabled*.

> BTW, you're doing the collateral requirement in a weird manner. Here's how I see it: if a user sells a put option at the 2000 strike, the collateral requirement at price S is

In the analysis I was doing I counted the LP position as part of the user's collateral so we could generalize this a little and compare the mechanisms with other lending protocols. I'm aware we don't actually account it that way -- there is a "basal" token collateral requirement with an assumed maximum (OTM) value for the LP position and we add losses from that OTM value to that collateral requirement for users instead of computing the value of the LP position directly. 

> Users cannot use ETH collateral for puts when the price deviates >20% because we mandate the positions to be covered when there is a large discrepancy between internal & current ticks
> Hence, we only allow USDC to collateralize USDC positions. The price is not even entering the solvency calculations and I'd say we are not vulnerable to that type of scenario anymore.

Only in the first two scenarios where currentTick = 1000. I gave an example where these "hardened" solvency restrictions would actually prevent protocol loss and then presented a modified path a user could take to bypass those restrictions and force the protocol to underwrite a similar loan starting here:

>Seems like they worked? What if a user tries to disable your restrictions by flash-swapping the current price back to 2000 before minting their position:

The key point is that the exploit can *still be executed* even if the current price is equal to 2000 (i.e. an user manipulates it back to 2000 after it falls to 1000 to match the oracle price and disable your restrictions). I gave a method with ITM swaps enabled, but if they are disabled you could also do it by manipulating the price to 2000 depositing 0.2 + 1 ETH to create the position and withdrawing the 2000 USDC you borrow (this is a little bit easier even since you don't have to go through the hassle of adding JIT liquidity to capture the ITM swap). So, unless I'm missing something or there is some way to ascertain the "true" market price (the current price before it was manipulated back to 2000 in this case), this exploit cannot be prevented by adding solvency checks or circuit breakers.

Let me know if that makes sense. If I missed something it would be either:
- The attack cannot in fact be executed in any way (there is some flaw in the paths I detailed for both ITM/non-ITM swapping versions) if the pool price is manipulated back to 2000 USDC, even if the market price remains at 1000 after the transaction
- Your restrictions somehow remain enabled even if the current price, fast oracle, and slow oracle are all sitting around 2000 USDC

guil-lambert:
>So, unless I'm missing something or there is some way to ascertain the "true" market price (the current price before it was manipulated back to 2000 in this case), this exploit cannot be prevented by adding solvency checks or circuit breakers.
>Let me know if that makes sense. If I missed something it would be either:
>The attack cannot in fact be executed in any way (there is some flaw in the paths I detailed for both ITM/non-ITM swapping versions) if the pool price is manipulated back to 2000 USDC, even if the market price remains at 1000 after the transaction

Yeah, this is where I am a bit confused about the last scenario then. Does the user still sell an ITM option when the price was 1000? And then they move the price back up? In that case, I'm not sure what the attack is:

1. Price drops to 1000
2. They sell an ITM put at 2000: 
  a. user balance delta =  (+2000USDC, -1ETH)
  b. Collateral requirement = 2000USDC (100% req)
  c. User balance = (2000USDC, 0), collateralized
3. They cannot withdraw anything (balance = required)
4. They move the price back up to 2000-ε
5. ~~They do some JIT and sell another option at 2000~~
  _a. user balance delta = (+ε, 0)
  b. Collateral requirement = 0.2 * 2000 + 2000 = 2400
  c. User balance = (2000+ε, 0), undercollateralized and cannot mint (?)_
6. They instead close their 2000 option, sell another one at 1000 (?)
  a. user balance delta = (0, 0) because all options is OTM
  b. collateral requirement = 0.2 * 1000 = 200 USDC
  c. User balance = (2000+ε, 0), overcollateralized and can mint (but where's the profit?)

Or if there is no selling at 1000, we get:

1. Price drops to 1000
2. They move the price back up to 2000-ε
3. They do some JIT and sell another option at 2000
  a. user balance delta = (+ε, 0)
  b. Collateral requirement = 0.2 * 2000 = 400
  c. User balance = (+ε, 0.2ETH), collateralized but can only withdraw ε
4. They abandon that account, get liquidated and lose 0.2ETH (where's the profit?)

>Your restrictions somehow remain enabled even if the current price, fast oracle, and slow oracle are all sitting around 2000 USDC

I'd say if all three prices converge to the same value then that's the _correct_ price because there's no way to know what the price history is (and whether this way "real" or not)

I now see the potential scenario where users could atomically move the price within 20% of the stale oracle prices, but it's unclear how it makes them profit because all ITM transfers are based on the currentTick, which is within 20% of the oracle ticks and hence protected in terms of collateral (because `0.8 * 1.33 > 1`)

---

Another naughty change: I also added this condition that the collateral requirement can never be larger than the notional value of the option:

https://github.com/panoptic-labs/panoptic-v1-core-private/blob/597b4d3f39f8526d8e05b94212aaaa8ab450fe9e/contracts/CollateralTracker.sol#L1428-L1429

This one is more about UX: collateral requirement above 100% in unnecessary and limits the number of fully collateralized options that can be minted. 

This is especially damaging with poolUtilization > 90% and the base requirement is 100%, which would result in a collateral requirement for an ITM options (like a covered call) of `(1 + 1 - Price/Strike) = 200% - Price/Strike` ie. 1 ETH of collateral can only mint 0.5ETH of positionSize

dyedm1:
> They abandon that account, get liquidated and lose 0.2ETH (where's the profit?)

In this case profit comes from the JIT liquidity instead of being tokens that are withdrawn. The Uniswap pool fills `1 ETH` at the current tick and gets `2000 + ε USDC`. The user with the JIT liquidity takes the vast majority of that swap, so they effectively get to sell 1 ETH (that can be acquired for 1000 USDC beforehand) to the protocol for 2000 USDC. 

> I now see the potential scenario where users could atomically move the price within 20% of the stale oracle prices, but it's unclear how it makes them profit because all ITM transfers are based on the currentTick, which is within 20% of the oracle ticks and hence protected in terms of collateral (because 0.8 * 1.33 > 1)

This is not necessarily the case -- out-of-range ITM transfers are exactly the same regardless of how far out-of-range they are. Let's go through the scenario without ITM swaps, which is a little less confusing because you don't have to capture the ITM swap with JIT liquidity to profit:

ETH price @ 1000:
1. User deposits 1.4 ETH in collateral (0, 1.4 ETH)
2. User borrows 2000 USDC to mint 2000 put (currently 1 ETH), no ITM swaps so (2000 USDC, 0.4 ETH)
3. CR = 1400 USDC, surplus collateral = 1000 USDC
4. User withdraws 1000 USDC, but still has 1400 USDC worth of collateral and an LP position worth 1000 USDC at market price, so account is solvent and no protocol loss. (1000 USDC, 0.4 ETH)

ETH price @ 2000 (manipulated, market still = 1000):
1. User deposits ~~1.4 ETH~~ 1.2 ETH in collateral (0, ~~1.4 ETH~~ 1.2 ETH)
2. User borrows 2000 USDC to mint 2000 put (still 1 ETH), no ITM swaps so (2000 USDC, ~~0.4 ETH~~ 0.2 ETH)
3. CR = ~~1400 USDC~~ 400 USDC, surplus collateral = ~~1000 USDC~~ 2000 USDC
4. User withdraws ~~1000 USDC~~ 2000 USDC, now only has 200 USDC worth of collateral at market prices and will be liquidated once oracle price catches up (~~1000 USDC~~ 0 USDC, ~~0.4 ETH~~ 0.2 ETH)
5. The user deposited 1.2 ETH worth 1200 USDC and was allowed to withdraw 2000 USDC, so their net profit is +800 USDC

The only difference between these two scenarios is that the user is allowed to deposit less collateral to mint the position and withdraw more collateral post-mint @ 2000. Other than that, the transfers/uniswap interactions to create the position are identical.

dyedm1:
> This one is more about UX: collateral requirement above 100% in unnecessary and limits the number of fully collateralized options that can be minted.

Is it unnecessary if we allow cross-collateralization though? I agree with the premise that ideally you would always be able to collateralize a put with `amountMoved` *in the correct token*.

>ie. 1 ETH of collateral can only mint 0.5ETH of positionSize

Not sure this is entirely accurate. If we ignore the buffer (which you appear to have done in your calculations) then the ITM collateral requirement should match the ITM amount they receive pretty closely:
1. ETH price is 1000, PU > 90%
2. User decides to mint 2000 USDC 1-tick-wide put @ strike=2000. They deposit 2000 USDC as collateral. (2000 USDC, 0)
3. To do this, they borrow 2000 USDC and swap ~1000 (excl execution costs) of it for 1 ETH to create the position (3000 USDC, 0)
4. Base CR = 2000 USDC * 100% = 2000 USDC, ITM CR = 2000 USDC - (2000 USDC *(100% - 1000/2000 = 50%) = 1000 USDC) = 1000 USDC, Total CR = 2000 USDC baseCR + 1000 USDC ITMCR, Effective CR (what user has to deposit) = 3000 USDC totalCR - 1000 USDC ITM amount = 2000 USDC.

So, excluding the mint buffer, they have to deposit ~2000 USDC to borrow 2000 USDC and mint a 2000 USDC put. Seems fair to me. With your condition enabled, they would only have to deposit ~1000 USDC.

This generalizes the deeper ITM positions are, so your condition could allow users to borrow at massive LTVs. For a super deep ITM put worth 1 USDC at the current price they would only need to deposit 1 USDC...
(rationale: borrow 2000 USDC, swap 1 USDC to ETH to create LP position, the rest, 1999 USDC, is the ITM amount added to user's collateral balance, max collateral requirement is amountMoved = 2000 USDC so user only needs to deposit 1 USDC in addition to the ITM amount they received to meet the requirement)

guil-lambert:
>So, excluding the mint buffer, they have to deposit ~2000 USDC to borrow 2000 USDC and mint a 2000 USDC put. Seems fair to me. With your condition enabled, they would only have to deposit ~1000 USDC.
>This generalizes the deeper ITM positions are, so your condition could allow users to borrow at massive LTVs. For a super deep ITM put worth 1 USDC at the current price they would only need to deposit 1 USDC...
(rationale: borrow 2000 USDC, swap 1 USDC to ETH to create LP position, the rest, 1999 USDC, is the ITM amount added to user's collateral balance, max collateral requirement is amountMoved = 2000 USDC so user only needs to deposit 1 USDC in addition to the ITM amount they received to meet the requirement)

I think you're right. My thinking was when users mint covered calls, but it does expose the protocol to quite a level of risk when minting in a "non-covered" manner. 

COVERED:
- Initial balance = (0, 1 ETH), 90% PU, Base collateral requirement = 100%
- Minting a covered call at 2000 strike, requirement = (2*2000 - 1000) = 3000 USDC
- If user can mints 1 ETH worth of calls, new balance is (2000, 0), which is less than collateral requirement
- User can only mint 0.5ETH call, final balance = (1000, 0.5) and they'd get liquidated when price goes below 1000.

We can discuss, but I'll remove that max requirement

PR #79, opened 2024-06-13
fix: round up converted notional amounts

dyedm1:
This prevents:
- positions with zero notional value (0 base CR)
- short positions with insufficient borrowed amounts when OTM (`mint` payment rounded up while (converted) `shortAmount` rounded down)
- long positions with insufficient tokens to collateralize LP debt (`mint` payment rounded up while (converted) `longAmount rounded down)

The second one is largely an inconvenience, but the first and third can be combined for a reversible DoS against sellers relevant in select low-decimal pools:
1. Alice creates a short call at strike price = 1000 (current price = 2000)
2. Bob purchases that call in increments of 999 (wei), receiving token1 with the ability to withdraw immediately (this is quite unprofitable due to the fact that they are forced to pay 2 token0 (=4k token1) in commission fees. They could try to JIT but that would likely take an unrealistic amount of calls to hit the MEV tax profitability threshold).
3. Given that those positions have zero collateral requirement, they cannot be liquidated even if they withdraw their entire stack (assuming the position remains out-of-range -- obviously they could accumulate enough premia to be liquidatable). Even if they were liquidated, the protocol loss would be cancelled out by the earlier gain in commissions.
4. Thus, the only way Alice can get rid of Bob's position and fully close her position is by literally donating tokens to Bob to allow a force exercise to go through (it would not normally given the assumption that the account has sufficient collateral)

### DIFF
--- a/contracts/libraries/PanopticMath.sol
+++ b/contracts/libraries/PanopticMath.sol
@@ -15,6 +15,7 @@ import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
 import {LeftRightUnsigned, LeftRightSigned} from "@types/LeftRight.sol";
 import {LiquidityChunk} from "@types/LiquidityChunk.sol";
 import {TokenId} from "@types/TokenId.sol";
+import "forge-std/Test.sol";
 
 /// @title Compute general math quantities relevant to Panoptic and AMM pool management.
 /// @author Axicon Labs Limited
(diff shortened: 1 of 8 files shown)


dyedm1:
We shouldn't convert notional amounts at the strike price -- on 1bps/low-tickSpacing pools a half-tick can be the true geometric mean.

Can just do `(sqrtpLX96 * sqrtpHX96) / 2^96 = gmPX96` instead.

PeakyCryptos (approved):

contracts/libraries/PanopticMath.sol
```diff
@@ -581,23 +581,26 @@ library PanopticMath {
         uint128 positionSize,
         uint256 legIndex
     ) internal pure returns (LeftRightUnsigned) {
-        // get the tick range for this leg in order to get the strike price (the underlying price)
-        (int24 tickLower, int24 tickUpper) = tokenId.asTicks(legIndex);
-
         uint128 amount0;
         uint128 amount1;
+
+        (int24 tickLower, int24 tickUpper) = tokenId.asTicks(legIndex);
```
flanagansteve, line 587: Am i understanding correctly that these ticks are the boundaries of when this leg of the option is exercisable? based on asTicks doc:

```
    /// @notice Get the option position's nth leg's (index `legIndex`) tick ranges (lower, upper).
    /// @dev NOTE: Does not extract liquidity which is the third piece of information in a LiquidityChunk.
    /// @param self The TokenId to extract the tick range from
    /// @param legIndex The leg index of the position (in {0,1,2,3})
    /// @return legLowerTick The lower tick of the leg/liquidity chunk
    /// @return legUpperTick The upper tick of the leg/liquidity chunk
```

 if so, why is it two ticks instead of one? I would have expected an option's strike price to be one tick and we'd compare the instantaneous price from Uniswap to see if it was exercisable - IE for a call if currentTick >= strike, then its exercisable

flanagansteve (approved):
I tried to follow your math and double check the multiplication / division / bitshifting by 2^96. I think it makes sense, but of course only know a little about that so far, so take it with a grain of salt.

contracts/libraries/PanopticMath.sol
dyedm1: > Am i understanding correctly that these ticks are the boundaries of when this leg of the option is exercisable?

Yes, but (force) exercisability is defined that way because of other properties of that range (namely, that the option only earns premium when the price moves within the range), not the other way around. 

> if so, why is it two ticks instead of one? I would have expected an option's strike price to be one tick and we'd compare the instantaneous price from Uniswap to see if it was exercisable - IE for a call if currentTick >= strike, then its exercisable

Not quite. `strike` *does* represent the effective execution price of the option (although, as mentioned earlier, the user-input `strike` parameter is a half-tick off from the "true" strike or execution price for asymmetrical positions possible in certain pools), but the option cannot be fully executed until the price moves out of the *range* (Uniswap V3 positions are essentially reversible range orders, so Panoptions can be understood to be tokenized range orders). To replicate this instrument in TradFi you would either have to buy an option with some time to expiration (which has a similar ranging payoff curve in the short term) or  purchase many 0DTE options clustered around your range.

PR #80, opened 2024-06-21
Hardening PR0: more solvency checks

guil-lambert:

This first PR checks that the account is solvent at the slow+fast oracle ticks if currentTick diverges too much from oracle ticks

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -332,7 +332,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
         if (
             ct0.convertToAssets(ct0.balanceOf(msg.sender)) < minValue0 ||
             ct1.convertToAssets(ct1.balanceOf(msg.sender)) < minValue1
-        ) revert Errors.NotEnoughCollateral();
+        ) revert Errors.AccountInsolvent();
     }
 
     /// @notice Determines if account is eligible to withdraw or transfer collateral.
@@ -884,7 +884,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
         }
 
         bool solvent = _checkSolvencyAtTicks(user, positionIdList, currentTick, atTicks, buffer);
-        if (!solvent) revert Errors.NotEnoughCollateral();
+        if (!solvent) revert Errors.AccountInsolvent();
     }
 
     function _getOracleTicks()
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -501,7 +501,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     function pokeMedian() external {
         (, , uint16 observationIndex, uint16 observationCardinality, , , ) = s_univ3pool.slot0();
 
-        (, , uint256 medianData) = PanopticMath.computeInternalMedian(
+        (, uint256 medianData) = PanopticMath.computeInternalMedian(
             observationIndex,
             observationCardinality,
             MEDIAN_PERIOD,
@@ -876,11 +876,11 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 int256(currentTick - slowOracleTick) ** 2 >
             MAX_SLOW_FAST_DELTA ** 2
         ) {
-            atTicks = new int24[](3);
+            atTicks = new int24[](4);
             atTicks[0] = fastOracleTick;
             atTicks[1] = slowOracleTick;
             atTicks[2] = lastObservedTick;
-            atTicks[4] = currentTick;
+            atTicks[3] = currentTick;
         } else {
             atTicks = new int24[](1);
             atTicks[0] = fastOracleTick;
@@ -890,6 +890,12 @@ contract PanopticPool is ERC1155Holder, Multicall {
         if (!solvent) revert Errors.AccountInsolvent();
     }
 
+    /// @notice Burns and handles the exercise of options.
+    /// @return currentTick The current tick in the Uniswap pool (as returned in slot0)
+    /// @return fastOracleTick The fast oracle tick computed as the median of the past N observations in the Uniswap Pool
+    /// @return slowOracleTick The slow oracle tick as tracked by `s_miniMedian`
+    /// @return latestObservation The latest observation from the Uniswap pool (price at the end of the last block)
+    /// @return medianData the updated value for `s_miniMedian` (returns 0 if not enough time has passed since last observation)
     function _getOracleTicks()
         internal
         view
@@ -907,7 +913,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
         (, currentTick, observationIndex, observationCardinality, , , ) = _univ3pool.slot0();
 
-        (fastOracleTick, ) = PanopticMath.computeMedianObservedPrice(
+        (fastOracleTick, latestObservation) = PanopticMath.computeMedianObservedPrice(
             _univ3pool,
             observationIndex,
             observationCardinality,
@@ -916,7 +922,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
         );
 
         if (SLOW_ORACLE_UNISWAP_MODE) {
-            (slowOracleTick, latestObservation) = PanopticMath.computeMedianObservedPrice(
+            (slowOracleTick, ) = PanopticMath.computeMedianObservedPrice(
                 _univ3pool,
                 observationIndex,
                 observationCardinality,
@@ -924,7 +930,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 SLOW_ORACLE_PERIOD
             );
         } else {
-            (slowOracleTick, latestObservation, medianData) = PanopticMath.computeInternalMedian(
+            (slowOracleTick, medianData) = PanopticMath.computeInternalMedian(
                 observationIndex,
                 observationCardinality,
                 MEDIAN_PERIOD,
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1011,9 +1011,8 @@ contract PanopticPool is ERC1155Holder, Multicall {
         LeftRightUnsigned tokenData0;
         LeftRightUnsigned tokenData1;
         LeftRightSigned premia;
+        (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
         {
-            (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
-
             // Enforce maximum delta between TWAP and currentTick to prevent extreme price manipulation
             if (Math.abs(currentTick - twapTick) > MAX_TWAP_DELTA_LIQUIDATION)
                 revert Errors.StaleTWAP();
@@ -1057,7 +1056,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
         int256 liquidationBonus0;
         int256 liquidationBonus1;
-        int24 finalTick;
         {
             LeftRightSigned netExchanged;
             LeftRightSigned[4][] memory premiasByLeg;
@@ -1074,8 +1072,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 positionIdList
             );
 
-            (, finalTick, , , , , ) = s_univ3pool.slot0();
-
             LeftRightSigned collateralRemaining;
             // compute bonus amounts using latest tick data
             (liquidationBonus0, liquidationBonus1, collateralRemaining) = PanopticMath
@@ -1083,7 +1079,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
                     tokenData0,
                     tokenData1,
                     Math.getSqrtRatioAtTick(twapTick),
-                    Math.getSqrtRatioAtTick(finalTick),
+                    Math.getSqrtRatioAtTick(currentTick),
                     netExchanged,
                     premia
                 );
@@ -1098,8 +1094,8 @@ contract PanopticPool is ERC1155Holder, Multicall {
             // if premium is haircut from a token that is not in protocol loss, some of the liquidation bonus will be converted into that token
             // reusing variables to save stack space; netExchanged = deltaBonus0, premia = deltaBonus1
             address _liquidatee = liquidatee;
+            int24 _currentTick = currentTick;
             TokenId[] memory _positionIdList = positionIdList;
-            int24 _finalTick = finalTick;
             int256 deltaBonus0;
             int256 deltaBonus1;
             (deltaBonus0, deltaBonus1) = PanopticMath.haircutPremia(
@@ -1109,7 +1105,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 collateralRemaining,
                 s_collateralToken0,
                 s_collateralToken1,
-                Math.getSqrtRatioAtTick(_finalTick),
+                Math.getSqrtRatioAtTick(_currentTick),
                 s_settledTokens
             );
 
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -866,20 +866,25 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
         medianData = _medianData;
 
-        // Check the user's solvency at the fast tick; revert if not solvent
-        bool solventAtFast = _checkSolvencyAtTick(
-            user,
-            positionIdList,
-            currentTick,
-            fastOracleTick,
-            buffer
-        );
-        if (!solventAtFast) revert Errors.NotEnoughCollateral();
+        int24[] memory atTicks;
+        // If one of the ticks is too stale, we fall back to the more conservative tick, i.e,
+        // the user must be solvent at the fast and slow oracle ticks as well as the currentTick.
+        if (
+            Math.abs(int256(fastOracleTick) - slowOracleTick) > MAX_SLOW_FAST_DELTA ||
+            Math.abs(int256(fastOracleTick) - currentTick) > MAX_SLOW_FAST_DELTA ||
+            Math.abs(currentTick - slowOracleTick) > MAX_SLOW_FAST_DELTA
+        ) {
+            atTicks = new int24[](3);
+            atTicks[0] = fastOracleTick;
+            atTicks[1] = slowOracleTick;
+            atTicks[2] = currentTick;
+        } else {
+            atTicks = new int24[](1);
+            atTicks[0] = fastOracleTick;
+        }
 
-        // If one of the ticks is too stale, we fall back to the more conservative tick, i.e, the user must be solvent at both the fast and slow oracle ticks.
-        if (Math.abs(int256(fastOracleTick) - slowOracleTick) > MAX_SLOW_FAST_DELTA)
-            if (!_checkSolvencyAtTick(user, positionIdList, currentTick, slowOracleTick, buffer))
-                revert Errors.NotEnoughCollateral();
+        bool solvent = _checkSolvencyAtTicks(user, positionIdList, currentTick, atTicks, buffer);
+        if (!solvent) revert Errors.NotEnoughCollateral();
     }
 
     function _getOracleTicks()
@@ -1126,26 +1131,16 @@ contract PanopticPool is ERC1155Holder, Multicall {
             liquidatee,
             uint256(int256(uint256(_delegations.leftSlot())) + liquidationBonus1)
         );
+        // ensure the owner is solvent (insolvent accounts are not permitted to pay premium unless they are being liquidated)
+        _validateSolvency(msg.sender, positionIdListLiquidator, BP_DECREASE_BUFFER);
 
-        // check that the provided positionIdList matches the positions in memory
-        _validatePositionList(msg.sender, positionIdListLiquidator, 0);
-
-        if (
-            !_checkSolvencyAtTick(
-                msg.sender,
-                positionIdListLiquidator,
-                finalTick,
-                finalTick,
-                BP_DECREASE_BUFFER
-            )
-        ) revert Errors.NotEnoughCollateral();
-
-        LeftRightSigned bonusAmounts = LeftRightSigned
-            .wrap(0)
-            .toRightSlot(int128(liquidationBonus0))
-            .toLeftSlot(int128(liquidationBonus1));
-
-        emit AccountLiquidated(msg.sender, liquidatee, bonusAmounts);
+        {
+            LeftRightSigned bonusAmounts = LeftRightSigned
+                .wrap(0)
+                .toRightSlot(int128(liquidationBonus0))
+                .toLeftSlot(int128(liquidationBonus1));
+            emit AccountLiquidated(msg.sender, liquidatee, bonusAmounts);
+        }
     }
 
     /// @notice Force the exercise of a single position. Exercisor will have to pay a fee to the force exercisee.
@@ -1262,14 +1257,14 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @param account The account to check solvency for
     /// @param positionIdList The list of positions to check solvency for
     /// @param currentTick The current tick of the Uniswap pool (needed for fee calculations)
-    /// @param atTick The tick to check solvency at
+    /// @param atTicks An array of ticks to check solvency at
     /// @param buffer The buffer to apply to the collateral requirement
     /// @return Whether the account is solvent at the given tick
-    function _checkSolvencyAtTick(
+    function _checkSolvencyAtTicks(
         address account,
         TokenId[] calldata positionIdList,
         int24 currentTick,
-        int24 atTick,
+        int24[] memory atTicks,
         uint256 buffer
     ) internal view returns (bool) {
         (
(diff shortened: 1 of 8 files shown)


dyedm1:
I think this PR and #81 are both attempting to prevent a scenario where a massive (non-manipulated/wick and lasting) spot price move in a single block causes both oracle ticks to become extremely stale, creating an opportunity to extract value by allowing users to withdraw more collateral than they deposited due to inflated position or deposited collateral values. This is something that all oracle-based lending protocols are vulnerable to (even overcollateralized protocols like AAVE/Morpho).

The reasoning here seems to be that this exploit can be prevented by simply detecting if the current price has moved far enough away from the oracle price and incorporating that price into their solvency checks if so. While this does prevent a specific path a user could take to get there (no ITM swaps in any of the following):
1. Price of ETH moves from 2000 to 1000 in a single block (fastOracle and slowOracle = 2000)
1a. User acquires 1.2 ETH for 1200 USDC
2. User deposits 1.2 ETH (0.2 ETH=400 at oracle) (1.2 ETH, 0 USDC)
3. User mints a 2000 USDC put at S=2000 (0.2 ETH, 2000 USDC) 
4. User withdraws 2000 USDC (0.2 ETH, 0)
*this would be valid previous to this PR because that position is OTM at the oracle tick and 0.2 ETH is worth 400 USDC(20%)*
5. User has spent 1200 USDC and received 2000 USDC (-800 protocol loss if the price stays at 1000) *prevented!*

Unfortunately, while this specific PoC has been foiled, these checks do not address the underlying cause of the exploit (that is, a lagging oracle price relative to spot). An attacker who understands this can bypass these restrictions by just adding a swap step (which increases execution costs slightly, but does nothing to prevent protocol loss or foil the overall attack).
1. Price of ETH moves from 2000 to 1000 in a single block (fastOracle and slowOracle = 2000)
1a. User acquires 1.2 ETH for 1200 USDC
**1b. User swaps the pool price back to 2000-ε USDC, where ε is width/2**
2. User deposits 1.2 ETH (0.2 ETH=400 at oracle) (1.2 ETH, 0 USDC)
3. User mints a 2000 USDC put at S=2000 (0.2 ETH, 2000 USDC) 
4. User withdraws 2000 USDC (0.2 ETH, 0)
***this is now valid again because currentTick is close to 2000/fastOracle/slowOracle, so the extra collateral check is disabled (and even if not, it would be 3 checks at the ~same price)***

**5. User swaps the pool price back to 1000, making the cost of their manipulation `poolFee*amountSwapped*2`**
6. User has spent 1200 USDC **(+ swap fees)** and received 2000 USDC (-800 protocol loss if the price stays at 1000)

My opinion is that there's not much point in adding a solvency check at `currentTick` unless it prevents protocol loss in a scenario that can't be trivially bypassed with an atomic swap.

guil-lambert:
>I think this PR and https://github.com/panoptic-labs/panoptic-v1-core-private/pull/81 are both attempting to prevent a scenario where a massive (non-manipulated/wick and lasting) spot price move in a single block...

This PR resolves a failing invariant, which is: no account should be insolvent at the currentTick after performing any operation. 

I discovered this after I added these checks in this commit (0e6ad8a17428a43b2f64d0f3e8e73268b8d40285), and was able to resolve those failing checks by adding the solvency checks here (dd01790ad2829fdccf6e4cddce55efe3b62e022e). 

This PR cleans up dd01790ad2829fdccf6e4cddce55efe3b62e022e and adds an extra condition for the price divergence based on the vector magnitude and not the term-by-term divergence: 

https://github.com/panoptic-labs/panoptic-v1-core-private/blob/b66ced1c3bf0c95eefd7e3f625b9d21fe05ff190/contracts/PanopticPool.sol#L872-L876 

This new requirement means that the ε in 2000-ε USDC will be smaller, although it is not important at the moment.

>My opinion is that there's not much point in adding a solvency check at currentTick unless it prevents protocol loss in a scenario that can't be trivially bypassed with an atomic swap.

Atomic swaps or not, we should prevent the execution of any trade which results in the account being left insolvent at the end of the transaction, which itself resolves other critical issues besides the attack scenario you described. 

These PRs ares hardening the protocol, but they're not a panacea for all oracle-based attacks. We could revisit this atomic swap scenario in a separate PR, but for now let's focus on how this helps resolve the breaking invariant.

dyedm1:
> This PR resolves a failing invariant, which is: no account should be insolvent at the currentTick after performing any operation.

My question is why is this an invariant? I don't necessarily see any reason why that can't be true.

> which itself resolves other critical issues besides the attack scenario you described

Like what? Can you elaborate on this?

> These PRs ares hardening the protocol, but they're not a panacea for all oracle-based attacks. We could revisit this atomic swap scenario in a separate PR, but for now let's focus on how this helps resolve the breaking invariant.

Actually, I have an idea. What if we replace the current tick here with the most recent Uniswap observation? While that technically wouldn't resolve the invariant you set out here, it would have a material "hardening" effect and make it impossible to exploit stale fast/slow oracle prices without manipulating an oracle observation unless you could get it included in the same block (unlikely).

I'm not sure I see another archetype of "critical issue" that occurs when accounts are insolvent at `currentTick` besides the one I described, unless I'm missing something?

guil-lambert:
>My question is why is this an invariant? I don't necessarily see any reason why that can't be true.

We're using the FREI-PI pattern for mint/burn/settle/forceExercise/liquidate/withdraw to make sure the account is always solvent after any operation.  

Using the `fastOracleTick` works most of the time, but the echidna tests highlighted a specific scenario where a large swap occurred right before a mint, which meant `currentTick` and `fastOracleTick` diverged enough such that the account was solvent at `fastOracleTick` but not at `currentTick`. 

Both ticks are important: the `fastOracleTick` provides a "recent" price to be solvent at, but `currentTick` is also the state the pool is left off after the transaction is over. 

If we allow a transaction to leave the user insolvent at `currentTick`, then it is 100% certain that this account will be liquidated within 2+ blocks if the price doesn't move back closer to `fastOracleTick`. 

My question is: why is this a situation we should be OK with? Basically allowing accounts to be left in a liquidatable state?

>>which itself resolves other critical issues besides the attack scenario you described

>Like what? Can you elaborate on this?

If the `fastOracleTick` is always trailing the `finalTick` by 1 block, doesn't it lead to a rather large number of attack scenarios based on moving the price in the same block before performing a transaction? ie.

1. Account 1 moves the price from 2000 to 1000. `fastOracleTick` = 2000
2. In the same block, Account 2 sells 100ETH worth of calls at the 1000.01 strike. Account 2 needs `100*(0.2 - (1-1000/2000))` 70ETH of collateral, so it only deposits 70.01ETH
3. Account 1 swaps back up to 2000. It buys 100 ETH for 1000.01, spending 100,000USDC for 100 ETH now worth 200,000 USDC
4. Account 2 is insolvent, gets liquidated by account 1. Account 2 loses 70 ETH, liquidator (account 1?) gets small bonus.

Account 1 gained 100ETH, account 2 lost 70ETH, net profit = 30ETH = 60,000 USDC

PR80 doesn't quite solve this entirely, but adding PR81 and PR82 means step 2 requires Account 1 to have 100ETH in their account when minting the 1000.01 call, making account 1 always solvent and rendering the attack unprofitable.

> Actually, I have an idea. What if we replace the current tick here with the most recent Uniswap observation? While that technically wouldn't resolve the invariant you set out here, it would have a material "hardening" effect and make it impossible to exploit stale fast/slow oracle prices without manipulating an oracle observation unless you could get it included in the same block (unlikely).

I'm OK with adding it in addition to the fast/slow/current ticks. It's easy enough to the list of ticks [here](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/b66ced1c3bf0c95eefd7e3f625b9d21fe05ff190/contracts/PanopticPool.sol#L878-L881) and returning the 'lastObservedTick' from [computeInternalMedian](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/b66ced1c3bf0c95eefd7e3f625b9d21fe05ff190/contracts/libraries/PanopticMath.sol#L229C21-L229C37)

dyedm1:
> currentTick is also the state the pool is left off after the transaction is over.

Why is this a salient piece of information? Ultimately, liquidations (and, by extension, protocol loss) only happen after a sustained price move that fails to revert to the mean after a period of time.

> then it is 100% certain that this account will be liquidated within 2+ blocks if the price doesn't move back closer to fastOracleTick.

I don't think this is true? The TWAP period (not the fast oracle tick) is used for liquidations, so the price move would have to be sustained for at least 5 minutes (25+ blocks).

>My question is: why is this a situation we should be OK with? Basically allowing accounts to be left in a liquidatable state?

I wouldn't really view it as them being left in a "liquidatable state". Liquidations are not determined by the current tick, so their account wouldn't be liquidatable unless the TWAP was also at that price (which would require it to stay there for some time). All the other price oracles we use have some measure of "durability" (i.e. they indicate that prices were held over multiple blocks against external pressure from arbitrageurs), but the post-mint tick provides very little signal wrt past/future market prices and can be atomically set to whatever number is desired at a very low cost. I don't know of any other protocols that use the current tick as an oracle (except for some hacked ones that used the Curve current price...) and I don't see why we should give it significance unless there's some attack that's prevented specific to our interactions with Uniswap.

> Account 2 is insolvent, gets liquidated by account 1. Account 2 loses 70 ETH, liquidator (account 1?) gets small bonus.

Again, this is not the case unless that price move is "real" and lasts for more than 5 minutes across many blocks. Account 2 does kinda get screwed (they owe 100ETH, position worth 100k=50 ETH so they have to expend/lose 50 additional ETH to exercise the position which the swapper profits), but they're still perfectly solvent and able to exercise the position with no negative consequences to the protocol, the 50 ETH they lose is their problem for failing to set the slippage limits.

periphery/PanopticHelper.sol
```diff
@@ -119,14 +119,14 @@ contract PanopticHelper {
...
 
-        return
-            PanopticMath.computeInternalMedian(
-                observationIndex,
-                observationCardinality,
-                period,
-                medianData,
-                univ3pool
-            );
+        (int24 _medianTick, uint256 _medianData) = PanopticMath.computeInternalMedian(
```
dyedm1, line 122: Were you trying to resolve a stack-too-deep here?

contracts/libraries/PanopticMath.sol
```diff
@@ -187,7 +188,7 @@ library PanopticMath {
             }
 
             // get the median of the `ticks` array (assuming `cardinality` is odd)
-            return int24(Math.sort(ticks)[cardinality / 2]);
+            return (int24(Math.sort(ticks)[cardinality / 2]), int24(ticks[0]));
```
dyedm1, line 191: It occurs to me that the comment on L172 is no longer accurate given that the observations taken by this function are now configurable. We should probably remove or modify it such that it does not refer to a specific number of operations.

contracts/PanopticPool.sol
```diff
@@ -1286,12 +1332,20 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+        LeftRightUnsigned tokenData0,
+        LeftRightUnsigned tokenData1,
+        int24 atTick,
+        uint256 buffer
+    ) internal pure returns (bool) {
         (uint256 balanceCross, uint256 thresholdCross) = _getSolvencyBalances(
             tokenData0,
             tokenData1,
             Math.getSqrtRatioAtTick(atTick)
         );
```
```suggestion
        );
        
```
nit

contracts/PanopticPool.sol
```diff
@@ -1286,12 +1332,20 @@ contract PanopticPool is ERC1155Holder, Multicall {
             portfolioPremium.leftSlot()
         );
 
+        return _checkSolvencyCross(tokenData0, tokenData1, atTick, buffer);
+    }
+
+    function _checkSolvencyCross(
```
dyedm1: natspec? Are `_checkCrossBalances` and `_checkSolvencyCross` separate functions because you ran into a stack-too-deep?

contracts/PanopticPool.sol
```diff
@@ -1116,26 +1142,16 @@ contract PanopticPool is ERC1155Holder, Multicall {
             liquidatee,
             uint256(int256(uint256(_delegations.leftSlot())) + liquidationBonus1)
         );
```
dyedm1, line 1148:
```suggestion
        );
        
```
nit

contracts/PanopticPool.sol
```diff
@@ -1116,26 +1142,16 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-            )
-        ) revert Errors.NotEnoughCollateral();
-
-        LeftRightSigned bonusAmounts = LeftRightSigned
-            .wrap(0)
-            .toRightSlot(int128(liquidationBonus0))
-            .toLeftSlot(int128(liquidationBonus1));
-
-        emit AccountLiquidated(msg.sender, liquidatee, bonusAmounts);
+        {
```
dyedm1: Does this scoping do anything?

contracts/PanopticPool.sol
```diff
@@ -1059,16 +1087,14 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-            (, finalTick, , , , , ) = s_univ3pool.slot0();
-
             LeftRightSigned collateralRemaining;
             // compute bonus amounts using latest tick data
             (liquidationBonus0, liquidationBonus1, collateralRemaining) = PanopticMath
                 .getLiquidationBonus(
                     tokenData0,
                     tokenData1,
                     Math.getSqrtRatioAtTick(twapTick),
-                    Math.getSqrtRatioAtTick(finalTick),
```
dyedm1, line 1071: Did you mean to include this change (doing the swaps between liquidator/liquidatee @ current instead of final) in a different PR (e.g liquidation solvencies)? Doesn't seem to fit with the rest of these changes.

contracts/PanopticPool.sol
```diff
@@ -857,27 +857,72 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+            int24 slowOracleTick,
+            int24 lastObservedTick,
+            uint256 _medianData
+        ) = _getOracleTicks();
+
+        medianData = _medianData;
+
+        int24[] memory atTicks;
+        // If one of the ticks is too stale, we fall back to the more conservative tick, i.e,
+        // the user must be solvent at the fast and slow oracle ticks as well as the currentTick.
```
dyedm1: This comment should probably be updated to explain:
 - all the ticks that are being used (missing lastObservedTick)
 - what the staleness calculation is doing

contracts/PanopticPool.sol
```diff
@@ -857,27 +857,72 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+            atTicks[1] = slowOracleTick;
+            atTicks[2] = lastObservedTick;
+            atTicks[3] = currentTick;
+        } else {
+            atTicks = new int24[](1);
+            atTicks[0] = fastOracleTick;
+        }
+
+        bool solvent = _checkSolvencyAtTicks(user, positionIdList, currentTick, atTicks, buffer);
+        if (!solvent) revert Errors.AccountInsolvent();
```
```suggestion
        if (!_checkSolvencyAtTicks(user, positionIdList, currentTick, atTicks, buffer)) revert Errors.AccountInsolvent();
```
nit, unless this causes stack-too-deep for some reason?

contracts/PanopticPool.sol
```diff
@@ -1116,26 +1142,16 @@ contract PanopticPool is ERC1155Holder, Multicall {
             liquidatee,
             uint256(int256(uint256(_delegations.leftSlot())) + liquidationBonus1)
         );
+        // ensure the owner is solvent (insolvent accounts are not permitted to pay premium unless they are being liquidated)
```
dyedm1: This comment doesn't seem to fit the purpose of `_validateSolvency` on the liquidator (to ensure that the bonus payments don't somehow make them insolvent). I think this may have just been copied from `settleLongPremium` and inadvertently left in?

contracts/PanopticPool.sol
dyedm1, line 1071: Also, unsure how this managed to pass the liquidation test... (not enough fuzz runs? Echidna should eventually find it)

dyedm1 (changes requested):
Looks good minus a few nits and some functionality that was maybe intended for another PR. Even though our echidna tests will have to cover this, given the increasing complexity of the solvency check logic I think it might be a good idea to create a `Misc.t.sol` scenario that demonstrates the behavior of this logic in specific circumstances (i.e. different ticks being stale etc.). Maybe this is something we already have in another one of these hardening PRs? 

Edit: Guillaume will bring in aforementioned test from another PR and the changed liquidation functionality is functionally equivalent (because ITM swaps are disabled).

contracts/PanopticPool.sol
guil-lambert: Not sure why I split them, I merged those two functions in this commit dc2e7d72a75837ffc8405cef69cf9a82d35bb388

contracts/PanopticPool.sol
```diff
@@ -146,9 +146,9 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @dev Mitigates manipulation of the currentTick that causes positions to be liquidated at a less favorable price.
     int256 internal constant MAX_TWAP_DELTA_LIQUIDATION = 513;
 
-    /// @notice The maximum allowed delta between the fast and slow oracle ticks before solvency is evaluated at the slow oracle tick.
-    /// @dev Falls back on the more conservative (less solvent) tick during times of extreme volatility.
-    int256 internal constant MAX_SLOW_FAST_DELTA = 1800;
+    /// @notice The maximum allowed delta between the fast and slow oracle ticks before solvency is evaluated at the more oracle ticks (slow, current, and latest).
```
flanagansteve: How about: "The maximum allowed delta between the fast and slow oracle ticks before solvency is evaluated at the three oracle ticks: slow, current, and latest."

contracts/PanopticPool.sol
```diff
@@ -146,9 +146,9 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+    /// @dev Falls back on the more conservative (less solvent) tick during times of extreme volatility, where the price moves ~10% in <4 minutes.
+    int256 internal constant MAX_SLOW_FAST_DELTA = 953;
```
flanagansteve: Why does 953 delta == ~10% movement in 4 minutes? Might be good to explain how you got that number here

Edit: Ah, i see you explain this as the magnitude of a 3D vector on lines 871 onward. Could maybe just say "Derivation of 953 explained below" or just leave it as-is.

contracts/PanopticPool.sol
```diff
@@ -857,27 +857,76 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+            uint256 _medianData
+        ) = _getOracleTicks();
+
+        medianData = _medianData;
+
+        int24[] memory atTicks;
+        // Fall back to a conservative approach if there's high deviation between internal ticks:
+        // Check solvency at the slowOracleTick, currentTick, and lastObservedTick instead of just the fastOracleTick.
+        // Deviation is measured as the magnitude of a 3D vector:
+        // (fastOracleTick - slowOracleTick, lastObservedTick - slowOracleTick, currentTick - slowOracleTick)
```
flanagansteve, line 874: If the deviation among all _three_ of these measures is what triggers our increased scrutiny, then perhaps the descriptor of MAX_SLOW_FAST_DELTA:

    /// @notice The maximum allowed delta between the fast and slow oracle ticks

should instead be:

    /// @notice The maximum allowed delta between: the fast & slow oracle tick, and the current & slow oracle tick, and the last-observed & slow oracle tick.

and in addition - perhaps we should rename MAX_SLOW_FAST_DELTA to MAX_DELTA_AGAINST_SLOW or just MAX_TICKS_DELTA?

contracts/PanopticPool.sol
```diff
@@ -857,27 +857,76 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+            atTicks[0] = fastOracleTick;
+        }
+
+        if (!_checkSolvencyAtTicks(user, positionIdList, currentTick, atTicks, buffer))
+            revert Errors.AccountInsolvent();
+    }
 
-        ) = _univ3pool.slot0();
-        int24 fastOracleTick = PanopticMath.computeMedianObservedPrice(
+    /// @notice Burns and handles the exercise of options.
```
flanagansteve: >     /// @notice Burns and handles the exercise of options.

This is wrong - should be:

      /// @notice Gets several ticks from Uniswap regarding the underlying pair.

contracts/PanopticPool.sol
```diff
@@ -1252,14 +1271,14 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
     /// @param currentTick The current tick of the Uniswap pool (needed for fee calculations)
-    /// @param atTick The tick to check solvency at
+    /// @param atTicks An array of ticks to check solvency at
     /// @param buffer The buffer to apply to the collateral requirement
     /// @return Whether the account is solvent at the given tick
-    function _checkSolvencyAtTick(
+    function _checkSolvencyAtTicks(
         address account,
         TokenId[] calldata positionIdList,
         int24 currentTick,
```
flanagansteve, line 1280: You could remove the `currentTick` argument entirely now and just include it in `atTicks`. You already do so as `atTicks[3]` when we exceed MAX_SLOW_FAST_DELTA so would just need to add it for the other case.

Alternately, you could _stop_ including it in `atTicks[3]` for the exceeded-MAX_SLOW_FAST_DELTA case for consistency and just use the arg for both.

There may be a gas saving in the former approach over the latter (or v.v.)

contracts/PanopticPool.sol
```diff
@@ -1273,6 +1292,36 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 currentTick
             );
 
+        uint256 numberOfTicks = atTicks.length;
+        bool solvent = true;
+        for (uint256 i; i < numberOfTicks; ++i) {
```
flanagansteve: You could safely move the `++i` to an `unchecked { i += 1 }` for a small gas saving - atTicks will always be length 2 or 4

contracts/PanopticPool.sol
```diff
@@ -1273,6 +1292,36 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 currentTick
             );
 
+        uint256 numberOfTicks = atTicks.length;
```
flanagansteve, line 1295: Also, there may be a small saving in removing this `numberOfTicks` var and just using `atTicks.length` in the loop like:

```
        for (uint256 i; i < atTicks.length; ++i) {
```

contracts/PanopticPool.sol
```diff
@@ -1273,6 +1292,36 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+        bool solvent = true;
```
flanagansteve, line 1296: Also, you could remove this declaration, and instead name the return variable in the function definition:

```
function _checkSolvencyAtTicks(
        address account,
        TokenId[] calldata positionIdList,
        int24 currentTick,
        int24 atTick,
        int24[] memory atTicks,
        uint256 buffer
    ) internal view returns (bool solvent) {
```

and now, you have `solvent` as default-false - so i would invert its name to `notSolvent`, and then invert the logic inside your loop:

```
notSolvent =
                notSolvent ||
                !_checkCrossBalances(
                    account,
                    atTicks[i],
                    positionBalanceArray,
                    portfolioPremium,
                    buffer
                );
```

and i think you may save some gas, and at the least make the method more concise - would remove the line declaring `bool solvent = true` as well as the `return solvent` after the loop

contracts/libraries/Errors.sol
```diff
@@ -51,7 +51,7 @@ library Errors {
     error NoLegsExercisable();
 
     /// @notice PanopticPool: the account is not solvent enough to perform the desired action
-    error NotEnoughCollateral();
```
flanagansteve, line 54: FYI, this is still referenced in `crytic-export/f1abd43cad845cf0a49ace4b99cf26f2.json`, although I don't think that's a problem

contracts/PanopticPool.sol
dyedm1: Not necessarily -- array lengths are stored in memory if not fixed-size so that would add more overhead on each iteration than just pulling from the stack

contracts/PanopticPool.sol
guil-lambert: I had to change it because the 1800 was extremely close to the protocol loss threshold; like ~1750 would lead to 0 protocol loss and ~1820 would. 

NYSE uses between 7% (for ETFs) and 10% (for stocks) price move to trigger a trading halt, so 20% is way above that threshold and may be unnecessarily risky (https://en.wikipedia.org/wiki/Trading_curb)

contracts/PanopticPool.sol
guil-lambert: That would add a bit of extra logic in `_checkSolvencyAtTicks` and gas+memory cost won't really change: https://ethereum.stackexchange.com/a/29905

contracts/PanopticPool.sol
guil-lambert: Good catch! I'm not sure if that's still an issue with the new solidity version, but I'll add it for consistency with the rest of the codebase

contracts/PanopticPool.sol
guil-lambert: Doing an `&&` means that if `solvent` is ever false then it won't call `_checkCrossBalances` in subsequent loops. Using an `||` would result in the whole loop be evaluated at each time.

PR #81, opened 2024-06-21
Hardening PR1: enforcing covered positions

guil-lambert:

This second PR makes all new mints/burns "covered" if the slow+fast oracle ticks diverge too much from currentTick

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -976,8 +976,10 @@ contract PanopticPool is ERC1155Holder, Multicall {
         {
             bool safeMode = _isSafeMode();
             // if safeMode, enforce covered at assignment
-            if (safeMode && (tickLimitLow > tickLimitHigh)) {
-                (tickLimitLow, tickLimitHigh) = (tickLimitHigh, tickLimitLow);
+            if (safeMode) {
+                if (tickLimitLow > tickLimitHigh) {
+                    (tickLimitLow, tickLimitHigh) = (tickLimitHigh, tickLimitLow);
+                }
             }
         }
 
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1499,10 +1499,25 @@ contract PanopticPool is ERC1155Holder, Multicall {
         return _isSafeMode();
     }
 
-    /// @notice Returns s_miniMedia, used to compute the slow oracle tick
-    /// @return s_miniMedian Uint256 that stores a sorted set of 8 price observations used to compute the internal median oracle price.
-    function miniMedian() external view returns (uint256) {
-        return s_miniMedian;
+    /// @notice Computes and returns all oracle ticks.
+    /// @return currentTick The current tick in the Uniswap pool (as returned in slot0)
+    /// @return fastOracleTick The fast oracle tick computed as the median of the past N observations in the Uniswap Pool
+    /// @return slowOracleTick The slow oracle tick as tracked by `s_miniMedian`
+    /// @return latestObservation The latest observation from the Uniswap pool (price at the end of the last block)
+    /// @return medianData the updated value for `s_miniMedian`
+    function getOracleTicks()
+        external
+        view
+        returns (
+            int24 currentTick,
+            int24 fastOracleTick,
+            int24 slowOracleTick,
+            int24 latestObservation,
+            uint256 medianData
+        )
+    {
+        (currentTick, fastOracleTick, slowOracleTick, latestObservation, ) = _getOracleTicks();
+        medianData = s_miniMedian;
     }
 
     /// @notice Get the current number of open positions for an account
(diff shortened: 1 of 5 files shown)


dyedm1:
I'm not sure what the advantage of forcing covered positions/no ITM swaps is. The attack I describe in #80 is done on a covered position in the first place.

guil-lambert:
This PR makes it impossible to access leverage for ITM positions when the price has moved a lot recently (ie. in `safeMode`). This helps reduce the risks to the protocol due to the minting of ITM positions in very volatile environments. This is similar to the DTCC requiring 100% collateralization for $GME because they feel like it. 

Specifically, minting an ITM position at strike K when the price is S requires the user to have a balance equal to (assuming puts):

- collateralRequirement = `0.2*K + (1-S/K)*K = 1.2*K - S`. If K=2000 and S=1000, then the user needs 1400 USDC of collateral.

BEFORE:
The user can mint this position with 1400USDC of collateral.  Since this position is minted ITM, the user receives 1000USDC due to the ITM swap and must only have a starting balance of (400USDC, 0ETH) or (0, 0.4ETH) or anything in between. 

AFTER:
The user can only mint this covered position with a starting balance of (0, 1ETH). The user ends up with (2000, 0ETH), which is more than the 1400 USDC required in terms of collateral. They can withdraw 600 USDC and still be solvent

If the price has recently moved or is part of a broader price manipulation attack, then the user must either use 1ETH they purchased at a much higher price or buy it in the manipulated pool (which means they'd buy their own ETH that was used to dump the price)

Again, this is not aimed at solving oracle-manipulation issues, but simply hardening the collateral requirements and increasing the safety of the protocol during "organic" periods of heightened volatility.

dyedm1:
>which means they'd buy their own ETH that was used to dump the price

Isn't this exactly what happens when ITM swaps are enabled though? The same swap is still occurring (N USDC for 1 ETH in the underlying pool), but the only difference in the second case is that the user needs the funds for the swap upfront (but since they can withdraw 600 USDC anyway, the final collateral required is only 400 USDC net). You could easily just temporarily borrow/flashloan the extra 600 USDC if you didn't have enough tokens to do the swap first.

> Again, this is not aimed at solving oracle-manipulation issues, but simply hardening the collateral requirements and increasing the safety of the protocol during "organic" periods of heightened volatility.

I get the idea, but I don't think this PR does anything concrete to accomplish that goal. If someone wanted to, it's fairly trivial to create the same position (as with ITM swaps enabled) with the same collateral distribution, swap execution and risk to the protocol after withdrawing tokens. 

If more protection is desired, increasing the collateral requirement or disabling cross-collateralization on #82 would be far more effective than this restriction.

guil-lambert:
> > which means they'd buy their own ETH that was used to dump the price
> 
> Isn't this exactly what happens when ITM swaps are enabled though? The same swap is still occurring (N USDC for 1 ETH in the underlying pool), but the only difference in the second case is that the user needs the funds for the swap upfront (but since they can withdraw 600 USDC anyway, the final collateral required is only 400 USDC net). You could easily just temporarily borrow/flashloan the extra 600 USDC if you didn't have enough tokens to do the swap first.

Maybe that's more relevant for "normal" scenarios, where a lot of trades may be happening during volatile markets (and not necessarily in a pool where users can do whatever they want with the price). This is more like a circuit-breaker that limits price squeezes rather than a specific protection for mint/withdraw/liquidate attacks.

The advantage of mandating covered positions means that users don't automatically market swap in low liquidity pools (either adversarially or due to a fat finger trade that didn't set up the tickLimits properly). 

Like if a user wants to close their ITM put option, they could  use 1 ETH already in their account (or even buy that 1 ETH wherever they want in the open market) instead of doing it in that pool:

1. Account 1 sells a 2000 put when the price is 2000 in the ETH-USDC-1bps pool (which has very thin liquidity)
2. Price naturally moves to 1000. slowOracleTick is still at 2000 but fast/current/latest are all at 1000. Account 1 is still solvent.
3. BEFORE: need to swap in the pool when closing unless they check the "covered" box on the UI. That ITM swap squeezes the price and pushes the price even lower. Execution for that swap is at 800 and all accounts in that pool are more insolvent 
4. AFTER: covered positions mean that they can buy 1 ETH in the ETH-USDC-5bps pool and deposit it in their account (or user 1 ETH already in their account) to close that position, guaranteeing a slippage-less execution, leaving the price at 1000 after that burn.

dyedm1:
> Maybe that's more relevant for "normal" scenarios, where a lot of trades may be happening during volatile markets (and not necessarily in a pool where users can do whatever they want with the price). This is more like a circuit-breaker that limits price squeezes rather than a specific protection for mint/withdraw/liquidate attacks.

I see. Yeah, I was actually going to write something in my comment about this being no more effective than a frontend restriction.

>The advantage of mandating covered positions means that users don't automatically market swap in low liquidity pools (either adversarially or due to a fat finger trade that didn't set up the tickLimits properly).

My view at least is that this is something we should definitely be handling on the frontend (and "didn't set up the tickLimits properly" is really not an acceptable outcome there in the first place). It's also not quite as effective as a frontend slippage restriction -- users can still get pwned by swapping on low-liq pools as long as the *starting price* is nominal.

Just not sure it makes sense to add this given that we're supposed to be pretty much frozen at this point and this is something that (imo) can be done better at the frontend level.

PR #82, opened 2024-06-21
Hardening PR2: full collateralization

guil-lambert:

This third PR makes all new mints/burns be 100% collateralized if the slow+fast oracle ticks diverge too much from currentTick by setting the utilization at 100%.

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -656,8 +656,10 @@ contract PanopticPool is ERC1155Holder, Multicall {
     ) internal returns (uint128) {
         bool safeMode = isSafeMode();
         // if safeMode, enforce covered deployment
-        if (safeMode && (tickLimitLow > tickLimitHigh)) {
-            (tickLimitLow, tickLimitHigh) = (tickLimitHigh, tickLimitLow);
+        if (safeMode) {
+            if (tickLimitLow > tickLimitHigh) {
+                (tickLimitLow, tickLimitHigh) = (tickLimitHigh, tickLimitLow);
+            }
         }
 
         (LeftRightUnsigned[4] memory collectedByLeg, LeftRightSigned totalSwapped) = SFPM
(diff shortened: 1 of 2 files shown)


dyedm1:
Can you elaborate on what exactly this is supposed to prevent? Not sure I see what we get out of this that isn't already covered by #80.

guil-lambert:
This one makes all new positions minted when `safeMode` is `true` require 100% collateralization *forever*. This affects both OTM and ITM psoitions.

Again, this is similar to the DTCC requiring 100% collateralization for $GME because they feel like it. But if the DTCC reduces that requirement to say 50%, then existing positions benefit for that change. 

In our case, the collateral requirement/pool utilization is set to 100% for as long as the position is open. This is making multi-block attacks much harder to pull off

Back to the example before: minting an ITM position at strike K when the price is S requires the user to have a balance equal to (assuming puts):

- collateralRequirement = `0.2*K + (1-S/K)*K = 1.2*K - S`. If K=2000 and S=1000, then the user needs 1400 USDC of collateral.

BEFORE:
The user can only mint this covered position with a starting balance of (0, 1ETH). The user ends up with (2000, 0ETH), which is more than the 1400 USDC required in terms of collateral. They can withdraw 600 USDC and still be solvent

AFTER:
The collateral requirement is now:
- collateralRequirement = `1*K + (1-S/K)*K = 2.2*K - S`. If K=2000 and S=1000, then the user needs 3400 USDC of collateral.

The user can only mint this covered position with a starting balance of (1400USDC, 1ETH). The user ends up with (3400, 0ETH), which equal to the 1400 USDC required in terms of collateral. They cannot withdraw anything and will remain solvent as long as they maintain 3400 in their account.


While this is not aimed at solving oracle-manipulation issues, only hardening the protocol.

dyedm1:
Ok, sure why not.

I mentioned in #81 that any restrictions we're attempting to implement there are ineffective, but we could increase the collateral requirement/disable cross-collateralization here for a more concrete effect if needed.  

Don't agree with the current definition of `safeMode` though. It should be the difference of the fast oracle tick, the slow oracle tick, and/or the last Uniswap observation (as I discussed in #80 -- maybe we colocate the safeMode setting with the solvency check, esp if we get rid of #81). `currentTick` is too easy to manipulate (and, as I understand it, this is supposed to be a heuristic that can't be disabled by moving the price around temporarily.)

guil-lambert:
>maybe we colocate the safeMode setting with the solvency check, esp if we get rid of https://github.com/panoptic-labs/panoptic-v1-core-private/pull/81

I don't think we'll get rid of 81: enforcing the covered minting/burning means that we get rid of any price impact for trades whenever the market is volatile. There's no other way to enforce that besides making all operation covered

PR #83, opened 2024-06-21
Hardening PR3: liquidation solvencies

guil-lambert:

This fourth PR does a couple of things.

1. It increases the "StaleTWAP" threshold to `MAX_SLOW_FAST_DELTA`, which will revert if the price move is more than 20%
2. It explicitly checks that the account is insolvent at the twapTick, currentTick, and fastOracleTick before moving forward with liquidations

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1023,10 +1023,33 @@ contract PanopticPool is ERC1155Holder, Multicall {
         // Assert the account we are liquidating is actually insolvent
         int24 twapTick = getUniV3TWAP();
 
+        (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
+        {
+            // Enforce maximum delta between TWAP and currentTick to prevent extreme price manipulation
+            if (Math.abs(currentTick - twapTick) > MAX_SLOW_FAST_DELTA) revert Errors.StaleTWAP();
+
+            (, int24 fastOracleTick, , , ) = _getOracleTicks();
+            // Ensure the accound is insolvent at twapTick, currentTick, and fastOracleTick
+            /// @dev do not check slowOracleTick because slow oracle is covered by twapTick
+            int24[] memory atTicks = new int24[](3);
+            atTicks[0] = twapTick;
+            atTicks[1] = currentTick;
+            atTicks[2] = fastOracleTick;
+
+            bool solvent = _checkSolvencyAtTicks(
+                liquidatee,
+                positionIdList,
+                currentTick,
+                atTicks,
+                NO_BUFFER
+            );
+
+            if (solvent) revert Errors.NotMarginCalled();
+        }
         LeftRightUnsigned tokenData0;
         LeftRightUnsigned tokenData1;
         LeftRightSigned premia;
-        (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
+
         {
             uint256[2][] memory positionBalanceArray = new uint256[2][](positionIdList.length);
             (premia, positionBalanceArray) = _calculateAccumulatedPremia(
@@ -1036,6 +1059,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
                 ONLY_AVAILABLE_PREMIUM,
                 currentTick
             );
+
             tokenData0 = s_collateralToken0.getAccountMarginDetails(
                 liquidatee,
                 twapTick,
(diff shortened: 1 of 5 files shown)


contracts/PanopticPool.sol
```diff
@@ -1025,6 +1036,18 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
             );
 
+            // Enforce maximum delta between TWAP and currentTick to prevent extreme price manipulation
+            if (Math.abs(currentTick - twapTick) > MAX_SLOW_FAST_DELTA) revert Errors.StaleTWAP();
+
+            (, int24 fastOracleTick, , ) = _getOracleTicks();
+            // Ensure the accound is insolvent at twapTick, currentTick, and fastOracleTick
+            /// @dev do not check slowOracleTick because slow oracle is covered by twapTick
+            if (
+                _checkSolvencyCross(tokenData0, tokenData1, twapTick, NO_BUFFER) ||
```
dyedm1: This only evaluates their cross-collateral balance (against the same collateral requirements) at a different tick. I assume what you were attempting to do was check their collateral at all three ticks?

dyedm1:
> It increases the "StaleTWAP" threshold to MAX_SLOW_FAST_DELTA, which will revert if the price move is more than 20%

I would be cautious with this. That 5% number is based on the minimum collateral requirement for long positions. If we increase this to 20%:
1. (P = 2000) Bob mints a short put with S = 2000 (size is irrelevant, but large enough to liquidate them)
2. Bob mints a 1 ETH long call with S = 1200
3. Some time passes and the price+oracles slowly drops to 1000 (assume Bob becomes liquidatable far above this point)
4. Alice swaps the price to ~1200 (crossing Bob's position entirely)
5. The liquidation executes identically to P = 1000, *except* that Bob must provide 1200 USDC instead of 1 ETH (currently worth ~1000 USDC, 200 less). 
6. Alice swaps the price back. In doing so, her execution price is better than her original swap because of the additional liquidity that Bob added (effectively, she gets to sell 1 ETH for 1200 USDC when it is only worth 1000)
7. Bob is left with 200 USDC worth of tokens less (assuming that Bob had the requisite USDC to close the long, if not the liquidator would have to provide USDC that would be converted to ETH at the final tick so the overall loss would be lower) than an unmanipulated liquidation. 

Also, because we use the final tick to convert tokens for protocol loss optimization, extractable value pops up there -- liquidators would be incentivized to manipulate the current price to get their tokens converted at a better price (even if the price manipulation doesn't cross any of the user's positions/the burn executes exactly the same). We could just not use the final tick, but as mentioned below, there's a better way if we do that anyway...

There is still some extractable value at 5%, but quite a bit less. We *can* remove this check and prevent most of the extractable value in another way by forcing the liquidator to compensate for it (as force exercisors are: https://github.com/panoptic-labs/panoptic-v1-core-private/blob/5c27d96ba43c1123f114f91ec9b11c403c8b6fd6/contracts/CollateralTracker.sol#L731), but it does require a bit of a redesign as we use the current/final tick in multiple locations (though I don't consider it a priority given that stale TWAP oracles can easily create similar quantities of extractable value and liquidations are unlikely to be prevented for long under this restriction even in the most extreme price moves).

dyedm1:
>It explicitly checks that the account is insolvent at the twapTick, currentTick, and fastOracleTick before moving forward with liquidations

Can you remind me what the purpose of this is again? Is it just to be more "forgiving" to users if a price wicks down long enough for the TWAP to absorb the move but then runs back up in a few blocks (obviously liquidators can easily swap `currentTick` back down but the fast oracle tick is somewhat more difficult to manipulate while absorbing price moves quickly).

contracts/PanopticPool.sol
guil-lambert: Oh yes, you're right. I probably meant to user `_checkCrossBalances` here and not pre-fetch the tokenDatas

guil-lambert:
> Can you remind me what the purpose of this is again? Is it just to be more "forgiving" to users if a price wicks down long enough for the TWAP to absorb the move but then runs back up in a few blocks

Yes, something like that. We could add `lastObservedTick` to the mix, but that goes along the lines of setting up another invariant that "no account can be liquidated if it would have been solvent at the currentTick" as a way to be more forgiving

guil-lambert:
> * Some time passes and the price+oracles slowly drops to 1000 (assume Bob becomes liquidatable far above this point)

Isn't this is a weird requirement If the price drops slowly enough, then other liquidators would have liquidated Bob before that, much closer to their liquidation price (?)

dyedm1:
> Isn't this is a weird requirement If the price drops slowly enough, then other liquidators would have liquidated Bob before that, much closer to their liquidation price (?)

Yeah, maybe. You could see it dropping quickly and waiting for the TWAP price to catch up too. There's tons of ways to execute this same archetype of thing on different positions, this is just an example of one of them.

PR #84, opened 2024-06-24
Hardening PR4: withdraw protections

guil-lambert:
This specifically address this comments https://github.com/panoptic-labs/panoptic-v1-core-private/pull/80#issuecomment-2183527356, where a user withdraws the max allowed after a transaction to "realize" profits and allow the account to be liquidated. 

### What
This PR changes the criteria for the total amount of collateral that can be removed though the `collateralTracker.withdraw()` function call. That check is not longer based on thge solvency of the account, but rather on the amount of funds in the account to exercise all positions.

### Why

At the beginning, we did not allow accounts to withdraw any collateral with open positions. This was not necessary and overall bad UX because accounts could still 1) close all their positions, 2) withdraw collateral and 3) re-deploy all positions as long as they were solvent at the end of step 3).

This was removed here to improve the UX in this PR: https://github.com/panoptic-labs/panoptic-v1-core-private/commit/61be5f8ca994fe32e48fd2da207aaee1355ffb60

The problem, which was highlighted in this comments https://github.com/panoptic-labs/panoptic-v1-core-private/pull/80#issuecomment-2183527356, is that users could still perform price manipulation attacks and withdraw a significant amount of collateral while solvent at the current/oracle ticks, but could become immediately insolvent at a later time.

### How

In this PR, `validateCollateralWithdrawable` no longer checks whether the account is _solvent_ after withdrawing the funds. That solvency check was indeed vulnerable to price moves, whether at the current block or at a later one.

Rather, it checks that the account has enough funds to close all positions through exercise (ie. by not relying on `swapAtMint` and requiring all exercised amounts to be available in the account). This requirement results in effectively requiring the account to be 100% collateralized after removing the collateral. This would decrease the capital efficiency by quite a bit, but at least it would never leave the account in a vulnerable state.

Specifically, `validateCollateralWithdrawable` now computes the covered amounts necessary for closing all positions using the newly created `PanopticMath.computeCoveredAmounts`. This function does this:

- If the position is `N` contracts of a short put at strike `K`, it requires the account to have at least `N*K` amount of `token1` in the account
  - this means the account has enough funds to "_buy `N` assets at the strike price_"
- If the position is `N` contracts of a short call at strike `K`, it requires the account to have at least `N` amount of `token0` in the account
  - this means the account has enough funds to "_sell `N` assets at the strike price_"- 
- If the position is `N` contracts of a long put at strike `K`, it requires the account to have at least `N` amount of `token0` in the account
  - this means the account has enough funds to "_sell `N` assets at the strike price_" 
- If the position is `N` contracts of a long at strike `K`, it requires the account to have at least `K` amount of `token1` in the account
  - this means the account has enough funds to "_buy `N` assets at the strike price_"

There is some netting happening [here](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/5c5287dab217698037de8c77b5bfaeab5ab5fece/contracts/libraries/PanopticMath.sol#L749-L757) between positions so that a put spread (long put and short put at a higher strike) does not require both `N` of `token0` and `N*K` of `token1`

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -576,7 +576,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         // update tracked asset balance
         s_poolAssets -= uint128(assets);
 
-        // reverts if account doesn't have enough colateral/is not solvent
+        // reverts if account doesn't have enough collateral/is not solvent
         s_panopticPool.validateCollateralWithdrawable(owner, positionIdList);
         s_panopticPool.isAccountSolvent(owner, positionIdList);
 
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1043,8 +1043,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // the inflow or outflow of pool assets is defined by the swappedAmount: it includes both the ITM swap amounts and the short/long amounts used to create the position
             // however, any intrinsic value is paid for by the users, so we only add the portion that comes from PLPs: the short/long amounts
             // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
-            s_poolAssets = uint128(uint256(updatedAssets));
-            s_inAMM = uint128(uint256(int256(uint256(s_inAMM)) + (shortAmount - longAmount)));
+            s_poolAssets = uint256(updatedAssets).toUint128();
+            s_inAMM = uint256(int256(uint256(s_inAMM)) + (shortAmount - longAmount)).toUint128();
 
             utilization = _poolUtilization();
         }
@@ -1089,8 +1089,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // update stored asset balances with net moved amounts
             // any intrinsic value is paid for by the users, so we do not add it to s_inAMM
             // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
-            s_poolAssets = uint128(uint256(updatedAssets + realizedPremium));
-            s_inAMM = uint128(uint256(int256(uint256(s_inAMM)) - (shortAmount - longAmount)));
+            s_poolAssets = uint256(updatedAssets + realizedPremium).toUint128();
+            s_inAMM = uint256(int256(uint256(s_inAMM)) - (shortAmount - longAmount)).toUint128();
 
             return (int128(tokenToPay));
         }
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -576,8 +576,9 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         // update tracked asset balance
         s_poolAssets -= uint128(assets);
 
-        // reverts if account is not solvent/eligible to withdraw
+        // reverts if account doesn't have enough colateral/is not solvent
         s_panopticPool.validateCollateralWithdrawable(owner, positionIdList);
+        s_panopticPool.isAccountSolvent(owner, positionIdList);
 
         // transfer assets (underlying token funds) from the PanopticPool to the LP
         SafeTransferLib.safeTransferFrom(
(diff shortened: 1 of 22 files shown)


dyedm1:
> The problem, which was highlighted in this comments https://github.com/panoptic-labs/panoptic-v1-core-private/pull/80#issuecomment-2183527356, is that users could still perform price manipulation attacks and withdraw a significant amount of collateral while solvent at the current/oracle ticks, but could become immediately insolvent at a later time.

There's another variant of the attack this wouldn't prevent where you use JIT liquidity to capture the profits instead of withdrawing them that I described in a comment on #76. I'll also note that one of the key reasons we opted to allow withdrawals with open positions is that those tokens can *also* be removed by a motivated actor through premium payments (i.e. `settleLongPremium`) subject to the same solvency checks (therefore, prohibiting withdrawals to a greater extent than premium payments was bad UX that didn't really improve security or work to prevent attacks).

> Rather, it checks that the account has enough funds to close all positions through exercise (ie. by not relying on swapAtMint and requiring all exercised amounts to be available in the account). This requirement results in effectively requiring the account to be 100% collateralized after removing the collateral. This would decrease the capital efficiency by quite a bit, but at least it would never leave the account in a vulnerable state.

I think this requirement is likely stringent enough that most (retail) users will never be able to withdraw -- it may make more sense just to get rid of the withdrawals again instead of implementing these restrictions. 

Considering that this is a pretty big UX hit and whatever we're trying to prevent/harden can bypass in multiple ways (even if we add this to `settleLongPremium` as well (which presents serious issues for option sellers), the JIT liquidity thing is still viable unless we disable ITM swaps entirely), I don't think the tradeoff makes sense here. 

I'm open to creating different profiles for illiquid pools in the future (ex. 100% collateral requirement, no cross-collateralization, no ITM swaps, etc.), but this doesn't meaningfully increase security/hardness int he version we have now.

contracts/CollateralTracker.sol
```diff
@@ -576,8 +576,9 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         // update tracked asset balance
         s_poolAssets -= uint128(assets);
 
-        // reverts if account is not solvent/eligible to withdraw
+        // reverts if account doesn't have enough colateral/is not solvent
```
flanagansteve: typo: colateral -> collateral

contracts/PanopticPool.sol
```diff
@@ -345,6 +346,28 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+            s_collateralToken0,
+            s_collateralToken1,
+            int256(coveredAmounts.rightSlot()),
+            int256(coveredAmounts.leftSlot())
+        );
+    }
+
+    /// @notice Determines if account is solvent.
+    /// @dev Checks whether account is solvent with `BP_DECREASE_BUFFER` according to `_validateSolvency`.
+    /// @dev Can be called externallt to identify insolvent and near-insolvent accounts before they are liquidated.
```
flanagansteve: typo: externallt -> externally

contracts/libraries/PanopticMath.sol
```diff
@@ -643,6 +729,51 @@ library PanopticMath {
         }
     }
 
+    /// @notice Compute the amount of funds required to exercise that tokenId.
+    /// @param tokenId The option position identifier
+    /// @param positionSize The number of positions minted
+    /// @return coveredAmounts Left-right packed word where where rightSlot = covered amount for token0 and leftSlot = covered amount for token1
+    function computeCoveredAmounts(
```
flanagansteve: If this method is intended to be internal, it should be prepended with a `_`

flanagansteve (approved):
I followed the logic you moved around and indeed it seems consistent with what we had before, with the only change being what you intended by the PR description:

"Specifically, validateCollateralWithdrawable now computes the covered amounts necessary for closing all positions using the newly created PanopticMath.computeCoveredAmounts."

Its possible I'm missing some second-order impacts of the changes here so I'd definitely get more eyes on this but it makes sense from what I can see.

Only comments are about docs/naming:

PR #87, opened 2024-07-01
fix: some typos in comments

flanagansteve:
- typo on overstimate in _getSolvencyBalances - Found this comment while tracing check solvency logic
- miscapitalisation of For in subRect - should be For not FOr
- (from Henry) remove claim in PanopticFactory about the SFPM that is no longer true

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -168,7 +168,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     SemiFungiblePositionManager internal immutable SFPM;
 
     /*//////////////////////////////////////////////////////////////
-                                STORAGE 
+                                STORAGE
     //////////////////////////////////////////////////////////////*/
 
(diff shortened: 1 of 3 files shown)


contracts/PanopticPool.sol
```diff
@@ -168,7 +168,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     SemiFungiblePositionManager internal immutable SFPM;
 
     /*//////////////////////////////////////////////////////////////
-                                STORAGE 
+                                STORAGE
```
flanagansteve, line 171: (this change seems fine too, i think an automatic linter in my pre-commit made it. happy to revert this change though the typo is what i care more about)

PeakyCryptos (approved):

dyedm1:
Tests flaked - rerunning

dyedm1:
See #89, rerunning tests again

dyedm1 (approved):

PR #88, opened 2024-07-02
fix: prevent `s_poolAssets` and `s_inAMM` underflowing during settlement

dyedm1:
In #78 we fixed a potential `s_poolAssets` underflow that could result in the collateral requirement scaling mechanism not working as intended due to inaccurate pool utilizations. It turns out that similar underflows (where assets are donated to the pool, allowing a call to go through that moves more tokens than are counted in the internal trackers) are also possible during settlement:

- `s_poolAssets` can underflow in `takeCommissionAddData` if a short position is minted with more tokens than counted in `s_poolAssets`
- `s_inAMM` can technically underflow in `takeCommissionAddData` if a long position is minted with `longAmount > s_inAMM` (i.e. opposing long/short positions use opposing `asset` values resulting in different `shortAmount`/`longAmount` corresponding to the same liquidity)
- `s_poolAssets` can underflow in `exercise` if a long position is closed with more assets than counted in `s_poolAssets`
- `s_inAMM` can technically underflow in `exercise` if a short position is closed with `shortAmount > s_inAMM` (again, due to potential discrepancies in movedAmounts/liquidity links between short/long positions)

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1043,7 +1043,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // however, any intrinsic value is paid for by the users, so we only add the portion that comes from PLPs: the short/long amounts
             // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
             s_poolAssets = uint256(updatedAssets).toUint128();
-            s_inAMM = uint128(uint256(int256(uint256(s_inAMM)) + (shortAmount - longAmount)));
+            s_inAMM = uint256(int256(uint256(s_inAMM)) + (shortAmount - longAmount)).toUint128();
 
             utilization = _poolUtilization();
         }
@@ -1089,7 +1089,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // any intrinsic value is paid for by the users, so we do not add it to s_inAMM
             // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
             s_poolAssets = uint256(updatedAssets + realizedPremium).toUint128();
-            s_inAMM = uint128(uint256(int256(uint256(s_inAMM)) - (shortAmount - longAmount)));
+            s_inAMM = uint256(int256(uint256(s_inAMM)) - (shortAmount - longAmount)).toUint128();
 
             return (int128(tokenToPay));
         }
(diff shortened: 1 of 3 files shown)


PeakyCryptos (approved):

contracts/CollateralTracker.sol
```diff
@@ -1042,8 +1042,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // the inflow or outflow of pool assets is defined by the swappedAmount: it includes both the ITM swap amounts and the short/long amounts used to create the position
             // however, any intrinsic value is paid for by the users, so we only add the portion that comes from PLPs: the short/long amounts
             // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
-            s_poolAssets = uint128(uint256(updatedAssets));
-            s_inAMM = uint128(uint256(int256(uint256(s_inAMM)) + (shortAmount - longAmount)));
+            s_poolAssets = uint256(updatedAssets).toUint128();
```
flanagansteve, line 1045: It seems safe currently to do this casting here rather than right when the subtraction happens on line 1020:

```
            int256 updatedAssets = int256(uint256(s_poolAssets)) - swappedAmount;
```

However, in the future we might add some functionality that utilises `updatedAssets` in between lines 1020 and here. And if that's the case, the new logic would be vulnerable to using an underflowed value.

Would it make sense to safecast on line 1020 instead? Or is this not a problem / do you have faith we'll remember this?

contracts/CollateralTracker.sol
```diff
@@ -1088,8 +1088,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // update stored asset balances with net moved amounts
             // any intrinsic value is paid for by the users, so we do not add it to s_inAMM
             // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
-            s_poolAssets = uint128(uint256(updatedAssets + realizedPremium));
-            s_inAMM = uint128(uint256(int256(uint256(s_inAMM)) - (shortAmount - longAmount)));
+            s_poolAssets = uint256(updatedAssets + realizedPremium).toUint128();
```
flanagansteve, line 1091: (The above comment also applies to here VS safecasting on line 1060)

flanagansteve (approved):
I also [left a question in the thread](https://panoptic-global.slack.com/archives/C04F33PCH0S/p1719939192012669?thread_ts=1719921576.298049&cid=C04F33PCH0S) about a possible third place of concern from the message that prompted this, but i think we're all set - only one question from me:

contracts/CollateralTracker.sol
dyedm1: > Would it make sense to safecast on line 1020 instead? Or is this not a problem / do you have faith we'll remember this?

I can't imagine where else we would use that during settlement. Also, it's not really an underflowed value because it's typed as negative signed integer - it only becomes "underflowed" when unsafely cast to an unsigned value.

flanagansteve: discussed live in standup - yep, there shouldn't be many usages for `updatedAssets` in either function aside from updating `s_poolAssets`

PR #90, opened 2024-07-03
 feat: test settleLongPremium

flanagansteve:


contracts/fuzz/FuzzDeployments.sol
```diff
@@ -1455,150 +1451,104 @@ contract FuzzDeployments is FuzzHelpers {
...
+        address settlee = actors[fuzzedActorIndex % actors.length];
 
-        address settledAccount = actors[i_settled % 4];
+        if (userPositions[settlee].length == 0) return;
 
-        if (userPositions[settledAccount].length < 1) {
-            emit LogString("No current positions");
-            revert();
-        }
+        for (uint256 i = 0; i < userPositions[settlee].length; ++i) {
```
flanagansteve, line 1460: Sorry about the lack of convenient local vars such as `settleesPositions` here. This method has a number of steps and can't have too many local vars or it goes "stack too deep". I broke out to helpers where appropriate, and i think the current state is a good tradeoff - hopefully you can still follow

PR #92, opened 2024-07-05
fix: prevent 0-liquidity positions from being created

dyedm1:
This prevents a potential DoS found by our Echidna tests that could interfere with liquidations and other functionality. 

For some reason, Uniswap allows pokes to be executed via 0-liquidity burns, but reverts any 0-liquidity mints (despite the lack of any obvious technical barriers). The unfortunate consequence of this is that while Panoptic users are permitted to create 0-liquidity long positions (which is a valid `.burn(, , 0)` call), they cannot close those positions under any circumstances (SFPM users could increase their position size, but PanopticPool positions cannot be modified once created).

It is important that we prevent those positions from being created because they prevent any otherwise valid liquidations/force exercises from going through on a given account. While this PR prevents *all* 0-liquidity interactions with Uniswap, we could permit 0-liquidity mints in the SFPM to some degree (either by moving the check to the PanopticPool or skipping 0-liquidity mints), though I don't necessarily see any utility in that.

### DIFF
--- a/contracts/SemiFungiblePositionManager.sol
+++ b/contracts/SemiFungiblePositionManager.sol
@@ -988,7 +988,7 @@ contract SemiFungiblePositionManager is ERC1155, Multicall {
 
             // s_accountLiquidity is a LeftRight. The right slot represents the liquidity currently sold (added) in the AMM owned by the user
             // the left slot represents the amount of liquidity currently bought (removed) that has been removed from the AMM - the user owes it to a seller
-            // the reason why it is called "removedLiquidity" is because long options are created by removing -ie.short selling LP positions
+            // the reason why it is called "removedLiquidity" is because long options are created by removing - ie. short selling LP positions
             uint128 startingLiquidity = currentLiquidity.rightSlot();
             uint128 removedLiquidity = currentLiquidity.leftSlot();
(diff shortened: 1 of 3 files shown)


contracts/SemiFungiblePositionManager.sol
```diff
@@ -993,6 +993,10 @@ contract SemiFungiblePositionManager is ERC1155, Multicall {
             uint128 removedLiquidity = currentLiquidity.leftSlot();
```
flanagansteve, line 993: Nit: i notice the comment above this has a missing space, we could add it while we're here:

> // the reason why it is called "removedLiquidity" is because long options are created by removing -ie.short selling LP positions

to

> // the reason why it is called "removedLiquidity" is because long options are created by removing - i.e., short selling LP positions

flanagansteve (approved):
Nice!

> While this PR prevents all 0-liquidity interactions with Uniswap, we could permit 0-liquidity mints in the SFPM to some degree (either by moving the check to the PanopticPool or skipping 0-liquidity mints), though I don't necessarily see any utility in that.

I agree - lets skip on adding this unless its specifically requested down the road

PeakyCryptos (approved):

PR #95, opened 2024-07-08
feat: fixed-size virtual delegation for force exercises

dyedm1:
Our previous method of delegating virtual shares for force exercises involved computing the maximum possible quantities of tokens required to exercise the user's options under any circumstance. Unfortunately, the method we were using to compute those quantities was flawed and would require a force exercisor to send tokens directly to an exercisee without being compensated. While this was unlikely in liquid pools when we had ITM swaps enabled, disabling ITM swaps introduces the potential for long positions that require different tokens of a different type than the corresponding `movedAmounts` when ITM.

This PR moves force exercises to a fixed-size delegation of `DEPOSIT_LIMIT * 10_000` to reduce the amount of calculation required and any further estimation risks. While a position could *in theory* require more tokens than that to close, the reasoning is that:
1. Under any circumstances where a virtual delegation is required, the exercisor is required to cover the difference because the exercisee must pay the full delegation back to the protocol (the only purpose of the virtual delegation is to allow the positions to be settled before the exercisor provides the requisite tokens without breaking the accounting, avoiding an extra precalculation step before the positions are closed.)
2. Thus, any exercisor would need to deposit the *required* delegation amount before executing the force exercise.
3. Any operation that requires ~10,000 distinct deposit calls can for all intents and purposes be considered a DoS
4. Therefore, this fixed-size delegation (which is very large, but not large enough to cause any accounting problems) does not impede force exercises in any meaningful way compared to a solution that attempts to calculate the maximum required delegation, given that a force exercise would be practically impossible in any situation where they differ 

I'm kind of tempted to do a similar thing for liquidations (eliminating the requirement for liquidators to compute the delegation amounts themselves), but the current system works well enough there so will likely keep it the same in the absence of any significant exigence.

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -17,7 +17,6 @@ import {SafeTransferLib} from "@libraries/SafeTransferLib.sol";
 import {LeftRightUnsigned, LeftRightSigned} from "@types/LeftRight.sol";
 import {LiquidityChunk} from "@types/LiquidityChunk.sol";
 import {TokenId} from "@types/TokenId.sol";
-import "forge-std/Test.sol";
 
 /// @title Collateral Tracking System / Margin Accounting used in conjunction with a Panoptic Pool.
 /// @author Axicon Labs Limited
@@ -375,7 +374,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @notice Returns the maximum deposit amount.
     /// @return maxAssets The maximum amount of assets that can be deposited
     function maxDeposit(address) external pure returns (uint256 maxAssets) {
-        return Constants.MAX_DEPOSIT_ASSETS;
+        return type(uint104).max;
     }
 
     /// @notice Returns shares received for depositing given amount of assets.
@@ -393,14 +392,14 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     }
 
     /// @notice Deposit underlying tokens (assets) to the Panoptic pool from the LP and mint corresponding amount of shares.
-    /// @dev There is a maximum asset deposit limit of Constants.MAX_DEPOSIT_ASSETS.
+    /// @dev There is a maximum asset deposit limit of 2 ** 104 - 1.
     /// @dev An MEV tax is levied, which is equal to a single payment of the commissionRate BEFORE adding the funds.
     /// @dev Shares are minted and sent to the LP ('receiver').
     /// @param assets Amount of assets deposited
     /// @param receiver User to receive the shares
     /// @return shares The amount of Panoptic pool shares that were minted to the recipient
     function deposit(uint256 assets, address receiver) external returns (uint256 shares) {
-        if (assets > Constants.MAX_DEPOSIT_ASSETS) revert Errors.DepositTooLarge();
+        if (assets > type(uint104).max) revert Errors.DepositTooLarge();
 
         shares = previewDeposit(assets);
 
@@ -428,9 +427,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @return maxShares The maximum amount of shares that can be minted.
     function maxMint(address) external view returns (uint256 maxShares) {
         unchecked {
-            return
-                (convertToShares(Constants.MAX_DEPOSIT_ASSETS) * (DECIMALS - COMMISSION_FEE)) /
-                DECIMALS;
+            return (convertToShares(type(uint104).max) * (DECIMALS - COMMISSION_FEE)) / DECIMALS;
         }
     }
 
@@ -453,7 +450,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     }
 
     /// @notice Deposit required amount of assets to receive specified amount of shares.
-    /// There is a maximum asset deposit limit of Constants.MAX_DEPOSIT_ASSETS.
+    /// There is a maximum asset deposit limit of 2**104 - 1.
     /// An MEV tax is levied, which is equal to a single payment of the commissionRate BEFORE adding the funds.
     /// @dev Shares are minted and sent to the LP ('receiver').
     /// @param shares Amount of shares to be minted.
@@ -462,7 +459,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     function mint(uint256 shares, address receiver) external returns (uint256 assets) {
         assets = previewMint(shares);
 
-        if (assets > Constants.MAX_DEPOSIT_ASSETS) revert Errors.DepositTooLarge();
+        if (assets > type(uint104).max) revert Errors.DepositTooLarge();
 
         // transfer assets (underlying token funds) from the user/the LP to the PanopticPool
         // in return for the shares to be minted
@@ -1077,8 +1074,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
 
             // add premium and token deltas not covered by swap to be paid/collected on position close
             int256 tokenToPay = swappedAmount - (longAmount - shortAmount) - realizedPremium;
-            console2.log("totalAssetsBefore", totalAssets());
-            console2.log("totalSupplyBefore", totalSupply);
+
             if (tokenToPay > 0) {
                 // if user must pay tokens, burn them from user balance (revert if balance too small)
                 uint256 sharesToBurn = Math.mulDivRoundingUp(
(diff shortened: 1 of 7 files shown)


contracts/PanopticPool.sol
```diff
@@ -1203,34 +1191,22 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-                msg.sender,
-                refundAmounts.rightSlot() - delegatedAmounts.rightSlot()
-            );
-            s_collateralToken1.refund(
-                account,
-                msg.sender,
-                refundAmounts.leftSlot() - delegatedAmounts.leftSlot()
-            );
-        }
+        // settle difference between delegated amounts (from the protocol) and exercise fees/substituted tokens
```
flanagansteve, line 1210: Curious why it was fine to have this `unchecked` before, but no longer? Is it that you expect:

```
LeftRightSigned refundAmounts = PanopticMath.getExerciseDeltas(
  account,
  exerciseFees,
  twapTick,
  s_collateralToken0,
  s_collateralToken1
);
```

is larger than when `refundAmounts` was the argument over `exerciseFees`, and `refund` doesn't have its own logic to prevent overflows?

edit: indeed, `refund` does seem to be doing the `-=` on `balanceOf[user]` outside of any `unchecked`s and protecting against underflow: https://github.com/panoptic-labs/panoptic-v1-core-private/blob/main/contracts/CollateralTracker.sol#L920

so i suppose its the idea that `refundAmounts` now can never be greater than the user's CT balance, but it could have before, and it would have caused some issue?

flanagansteve: Discussed live in standup

contracts/libraries/Constants.sol
```diff
@@ -20,4 +22,11 @@ library Constants {
     /// @notice Maximum possible sqrtPriceX96 in a Uniswap V3 pool
     uint160 internal constant MAX_V3POOL_SQRT_RATIO =
         1461446703485210103287273052203988822378723970342;
+
+    /// @notice Maximum amount of assets permitted for a single collateral deposit
+    uint256 internal constant MAX_DEPOSIT_ASSETS = type(uint104).max;
+
+    /// @notice LeftRight-packed token0:token1 right:left quantities of virtual assets to delegate during liquidations/force exercises
+    /// @dev `LeftRightSigned.wrap(0).toRightSlot(MAX_DEPOSIT_ASSETS * 10_000).toLeftSlot(MAX_DEPOSIT_ASSETS * 10_000)`
+    int128 internal constant STANDARD_DELEGATION = int128(int256(MAX_DEPOSIT_ASSETS * 10_000));
```
flanagansteve: Double checking: Will this always actually be enough? I believe you made it a multiple of the max deposit amount - is it true you always need <= 100% of the initial deposit amount of assets to force exercise somebody?

You mention in your PR description that someone could deposit 10k times to achieve a position size equal to this amount - and i think the implication is that the max amount a force exercisor could need would be 100% of the position size. But just want to check that nothing such as fluctuations in pool price could change the veracity of that implication.

contracts/CollateralTracker.sol
```diff
@@ -374,7 +374,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @notice Returns the maximum deposit amount.
```
flanagansteve: In the docstring for `delegate` we claim that it mints ghost shares for the passed `depositor`:

> /// @notice Delegate and transfer shares corresponding to the incoming assets from the protocol to `delegatee`.
>    /// @dev This is controlled by the Panoptic Pool - not individual users.
>    /// @dev mints ghost shares so a position can be settled - the total supply is not affected.

I would recommend we reword this to: "mints shares to a user without affecting the total supply". I assume we wrote this docstring to communicate that we _intend_ to use `delegate` in a way such that we always revoke the shares we created for the user by the end of the transaction. But i think that describing the method in a way that implies it is _guaranteed_ we do that is inappropriate - we could indeed slip up and fail to that in some usages, and the docstring biases readers (auditors, users) to believe it is guaranteed.

contracts/PanopticPool.sol
```diff
@@ -1161,33 +1161,21 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
         // on forced exercise, the price *must* be outside the position's range for at least 1 leg
         touchedId[0].validateIsExercisable(twapTick);
 
         // The protocol delegates some virtual shares to ensure the burn can be settled.
-        s_collateralToken0.delegate(account, uint128(delegatedAmounts.rightSlot()));
-        s_collateralToken1.delegate(account, uint128(delegatedAmounts.leftSlot()));
+        s_collateralToken0.delegate(account, uint128(Constants.STANDARD_DELEGATION));
+        s_collateralToken1.delegate(account, uint128(Constants.STANDARD_DELEGATION));
 
         // Exercise the option
```
flanagansteve, line 1186: I tried to trace all the logic between the virtual share delegation (ln 1177) and refunding (ln 1208) and believe there's no way the exercisor can use their temporary large token balance to abuse the protocol. Here's the only thing close I could think of:

- `_burnAllOptionsFrom` contains nested calls to the collateral trackers, SFPM and the uniswap pool:
  - The method called in the SFPM, `burnTokenizedPosition`, which contains the call to the uniswap pool, is `ReentrancyLock`ed on calls to the `PanopticPool`. It calls the AMM via its internal helper `_validateAndForwardToAMM` to create positions and execute swaps, calling methods such as `univ3pool.mint`
    - Theoretically, if a user managed to create a malicious Uniswap pool, they could force exercise a position, receive a large amount of CollateralTracker shares from the virtual delegation, and then leverage the call from the SFPM to their malicious pool to then call methods on the Collateral Tracker, I believe. I do not think we are concerned about this - malicious Uniswap pools like this would be easy to identify and Panoptic pools created on top of them should fail to gain usage, leaving the free virtual CollateralTracker shares valueless. 
  - The functions called in the CTs do not call other contracts or send ETH.

flanagansteve (approved):
Logic makes sense - only questions below:

contracts/libraries/Constants.sol
dyedm1: > You mention in your PR description that someone could deposit 10k times to achieve a position size equal to this amount - and i think the implication is that the max amount a force exercisor could need would be 100% of the position size. But just want to check that nothing such as fluctuations in pool price could change the veracity of that implication.

It's more about the idea that *if* the required delegation amounts are greater than the STANDARD_DELEGATION, a newly created force exercising account would need to deposit 10k times to exercise the position *either way*.

contracts/CollateralTracker.sol
dyedm1: > I assume we wrote this docstring to communicate that we intend to use delegate in a way such that we always revoke the shares we created for the user by the end of the transaction.

That is the intention, but I'm not sure where you got that implication from this docstring. It just communicates that the shares do not contribute to the share price/share supply and are used only for accounting purposes (not that the function automatically revokes them or something).

contracts/PanopticPool.sol
dyedm1: >Theoretically, if a user managed to create a malicious Uniswap pool, they could force exercise a position, receive a large amount of CollateralTracker shares from the virtual delegation, and then leverage the call from the SFPM to their malicious pool to then call methods on the Collateral Tracker, I believe. I do not think we are concerned about this - malicious Uniswap pools like this would be easy to identify and Panoptic pools created on top of them should fail to gain usage, leaving the free virtual CollateralTracker shares valueless.

All those shares are eventually burned though, so it would underflow and revert if you tried to do that.

contracts/PanopticPool.sol
dyedm1: > Discussed live in standup

I just realized that I was talking about a *different* unchecked block in standup that I added in a test to make it pass but did not yet push in standup, which is probably why I sounded confusing.

As far as this one goes, the long and short of it is unchecked blocks only apply to operations in the same scope, so after that `refundAmount - delegatedAmount` subtraction is removed, the unchecked block no longer has any effect.

dyedm1:
Re-requesting review on this because significant changes have been made post the last review

contracts/PanopticPool.sol
flanagansteve: Aha ok - agreed makes sense

contracts/CollateralTracker.sol
```diff
@@ -903,22 +903,26 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         _transferFrom(delegator, delegatee, shares);
     }
 
-    /// @notice Delegate and transfer shares corresponding to the incoming assets from the protocol to `delegatee`.
+    /// @notice Increase the share balance of a user by shares worth `assets` without updating the total supply.
```
flanagansteve, line 906: Nice! I like this

contracts/libraries/PanopticMath.sol
```diff
@@ -917,61 +917,76 @@ library PanopticMath {
...
+            ) -
+                int256(collateral0.convertToAssets(collateral0.balanceOf(exercisee))) +
+                exerciseFees.rightSlot();
 
             if (balanceShortage > 0) {
                 return
                     LeftRightSigned
                         .wrap(0)
-                        .toRightSlot(int128(refundValues.rightSlot() - balanceShortage))
+                        .toRightSlot(int128(exerciseFees.rightSlot() - balanceShortage))
```
flanagansteve, line 953: No clue if this is helpful but: I note that balanceShortage above is:

```
int256(
  Math.mulDivRoundingUp(
    delegatedShares.rightSlot(),
    collateral0.totalAssets(),
    collateral0.totalSupply()
  )
)

-

int256(collateral0.convertToAssets(collateral0.balanceOf(exercisee)))

+

exerciseFees.rightSlot();
```

which means that this `exerciseFees.rightSlot() - balanceShortage`, as well as the token1 equivalent on line 978, can be fully expanded to:

```
exerciseFees.rightSlot()

-

int256(
  Math.mulDivRoundingUp(
    delegatedShares.rightSlot(),
    collateral0.totalAssets(),
    collateral0.totalSupply()
  )
)

+

int256(collateral0.convertToAssets(collateral0.balanceOf(exercisee)))

-

exerciseFees.rightSlot();
```

Which means you could cancel out the two `exerciseFees.rightSlot();` terms and arrive at:

```
int256(
  Math.mulDivRoundingUp(
    delegatedShares.rightSlot(),
    collateral0.totalAssets(),
    collateral0.totalSupply()
  )
)

+

int256(collateral0.convertToAssets(collateral0.balanceOf(exercisee)))
```

Which is maybe cheaper? But i doubt it given all the extra method calls compared to this `nt128(exerciseFees.rightSlot() - balanceShortage)`. Just thought i'd call it out though in case it prompted you to rethink anything (moving the conditional logic around, etc)

flanagansteve (approved):
Nice stuff!!

contracts/libraries/PanopticMath.sol
dyedm1: Yep am aware it cancels but definitely not cheaper to expand (still need to do extra operations, `mulDiv` super expensive if not cached, etc...)

PR #97, opened 2024-07-11
feat: test burn

flanagansteve:
need to:

- create the wrapper methods on PanopticPool (optionData and premiaSettlementData) we inherited from the try_settle_long branch

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -565,6 +565,52 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         return shares;
     }
 
+    /// @notice Redeem the amount of shares required to withdraw the specified amount of assets.
+    /// @dev Reverts if the account is not solvent with the given `positionIdList`.
+    /// @dev Shares are burned and assets are sent to the LP ('receiver').
+    /// @param assets Amount of assets to be withdrawn
+    /// @param receiver User to receive the assets
+    /// @param owner User to burn the shares from
+    /// @param positionIdList The list of all option positions held by `owner`
+    /// @return shares The amount of shares burned to withdraw the desired amount of assets
+    function withdraw(
+        uint256 assets,
+        address receiver,
+        address owner,
+        TokenId[] calldata positionIdList
+    ) external returns (uint256 shares) {
+        shares = previewWithdraw(assets);
+
+        // check/update allowance for approved withdraw
+        if (msg.sender != owner) {
+            uint256 allowed = allowance[owner][msg.sender]; // Saves gas for limited approvals.
+
+            if (allowed != type(uint256).max) allowance[owner][msg.sender] = allowed - shares;
+        }
+
+        // burn collateral shares of the Panoptic Pool funds (this ERC20 token)
+        _burn(owner, shares);
+
+        // update tracked asset balance
+        s_poolAssets -= uint128(assets);
+
+        // reverts if account is not solvent/eligible to withdraw
+        s_panopticPool.validateCollateralWithdrawable(owner, positionIdList);
+
+        // transfer assets (underlying token funds) from the PanopticPool to the LP
+        SafeTransferLib.safeTransferFrom(
+            s_underlyingToken,
+            address(s_panopticPool),
+            receiver,
+            assets
+        );
+
+        emit Withdraw(msg.sender, receiver, owner, assets, shares);
+
+        return shares;
+    }
+
+
     /// @notice Returns the maximum amount of shares that can be redeemed for a given user.
     /// If the user has any open positions, the max redeemable balance is zero.
     /// @param owner The redeeming address.
@@ -1370,7 +1416,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                            BUYING
                            POWER
                            REQUIREMENT
-                         
+
                                          ^               .         .
                                          |        <- ITM . <-ATM-> . OTM ->
                            100% + SCR% - |--__           .    .    .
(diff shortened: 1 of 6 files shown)


contracts/fuzz/FuzzDeployments.sol
```diff
@@ -1348,75 +1345,634 @@ contract FuzzDeployments is FuzzHelpers {
...
+                position,
+                legIndex,
+                seller,
+                posSize,
+                premiumAccumulator0,
+                premiumAccumulator1
+            );
+        }
+
+        (
```
flanagansteve, line 2000: there's is some consolidation i can do here to make it so `_calc_premia_from_preburn_values` just sets the idealPremium0 / proratedPremium0 on each `premiaAndAccumulators`, instead of getting these 4 returned values to then cycle through them to set each premia field on each `premiaAndAccumulators`.

contracts/fuzz/FuzzDeployments.sol
```diff
@@ -1348,75 +1345,634 @@ contract FuzzDeployments is FuzzHelpers {
...
+        for (uint legIndex = 0; legIndex < numLegs; legIndex++) {
+            _set_preburn_accumulators(premiaAndAccumulators, position, legIndex);
+
+            (uint128 premiumAccumulator0, uint128 premiumAccumulator1) = _get_account_premium(
+                legIndex,
+                preburnCurrentTick,
+                position
+            );
+
+            _set_premia_calc_inputs(
```
flanagansteve: hate how i had to kick these inputs to their own array of structs instead of having 5 arrays here; feels harder to trace now. But had to fight stack-too-deep

contracts/CollateralTracker.sol
```diff
@@ -565,6 +565,51 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         return shares;
     }
 
+    /// @notice Redeem the amount of shares required to withdraw the specified amount of assets.
```
flanagansteve, line 568: no actual changes in this file - just did a copy-paste to get the right withdraw method

contracts/SemiFungiblePositionManager.sol
```diff
@@ -118,7 +118,7 @@ contract SemiFungiblePositionManager is ERC1155, Multicall {
     }
 
     /*//////////////////////////////////////////////////////////////
-                            IMMUTABLES
```
flanagansteve, line 121: same here - this is just whitespace changes

contracts/PanopticPool.sol
```diff
@@ -1437,6 +1450,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
         twapTick = PanopticMath.twapFilter(s_univ3pool, TWAP_WINDOW);
     }
 
+    // TODO: Move to wrapper
```
flanagansteve, line 1452: and for this file i need to move this and the optionData method to a PanopticPoolWrapper, is all

the only relevant changes to check out are in FuzzDeployments.sol

contracts/fuzz/FuzzDeployments.sol
```diff
@@ -1348,75 +1345,759 @@ contract FuzzDeployments is FuzzHelpers {
...
+            _try_burning_zero_size_position(caller, position, positionsNew, tickLimitLow);
             return;
         }
 
         if (position.isLong(0) == 0 && currentLiquidity.rightSlot() < liquidityChunk.liquidity()) {
-            hevm.prank(caller);
-            try panopticPool.burnOptions(position, positions_new, tickLimitLow, tickLimitHigh) {
-                assertWithMsg(false, "A short position with not enough liquidity was burned.");
-            } catch {}
+            _try_burning_short_pos_with_inadequate_liq(
```
dyedm1: These two negative cases aren't super interesting:
- zero-size positions are rejected in the SFPM
- NotEnoughLiquidity positions are rejected in the SFPM *before* being validated in the PanopticPool - the case you would want to check there is instead 90-100% liquidity utilization which is permitted by the SFPM but prohibited in the PanopticPool

Both of these can be replaced with a simple SFPM burn simulation (similar to the one done during minting, returns shouldRevert and some quotes).

contracts/fuzz/FuzzDeployments.sol
```diff
@@ -1348,75 +1345,759 @@ contract FuzzDeployments is FuzzHelpers {
...
+        assertWithMsg(
+            burnersPostburnToken0Balance ==
+                burnersPreburnToken0Balance + totalProjectedProratedPremium0,
+            "Burners token0 balance did not increase by the amount the premia would indicate"
+        );
+        assertWithMsg(
+            burnersPostburnToken1Balance ==
+                burnersPreburnToken1Balance + totalProjectedProratedPremium1,
+            "Burners token1 balance did not increase by the amount the premia would indicate"
+        );
```
dyedm1, line 1489: I'm unsure how this invariant is passing:
1. totalProjectedProratedPremium is not inclusive of tokens required to settle the position
2. The handling of long premium does not seem to be correct or even present. This goes for settledTokens invariants as well.
3. Even if both of those were handled correctly, strict equality would likely result in an off-by-one invariant failure at some point due to rounding. Tolerances or alternate share conversions would have to be added.

Likely you have an overflow somewhere in your calculations that is preventing these invariant checks from occurring.

dyedm1 (commented):
I saw a few invariant checks here that (should) fail because they are incorrectly implemented (or use incomplete calculated values) but presumably do not. This is probably a result of a revert in some of your calculations higher up that is preventing the invariant checks from running. Until those are running, it is difficult to debug/get feedback about your implementation, so I would start by finding where the shadow revert is. You can confirm that the test execution reaches a certain point by putting an assert(false) at that point and waiting for it to fail.

PR #101, opened 2024-07-15
Hardening PR1: enforcing covered positions 

guil-lambert:
I messed up the merging on https://github.com/panoptic-labs/panoptic-v1-core-private/pull/81/commits, this is the recovery branch

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1404,7 +1404,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
             ) + int24(uint24(medianData >> ((uint24(medianData >> (192 + 3 * 4)) % 8) * 24)))) / 2;
 
             // If ticks have recently deviated more than +/- 10%, enforce covered mints
-            if (Math.abs(currentTick - medianTick) > MAX_SLOW_FAST_DELTA) {
+            if (Math.abs(currentTick - medianTick) > MAX_TICKS_DELTA) {
                 safeMode = true;
             }
         }
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -976,8 +976,10 @@ contract PanopticPool is ERC1155Holder, Multicall {
         {
             bool safeMode = _isSafeMode();
             // if safeMode, enforce covered at assignment
-            if (safeMode && (tickLimitLow > tickLimitHigh)) {
-                (tickLimitLow, tickLimitHigh) = (tickLimitHigh, tickLimitLow);
+            if (safeMode) {
+                if (tickLimitLow > tickLimitHigh) {
+                    (tickLimitLow, tickLimitHigh) = (tickLimitHigh, tickLimitLow);
+                }
             }
         }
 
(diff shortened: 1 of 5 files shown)


contracts/PanopticPool.sol
```diff
@@ -1452,10 +1489,37 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+        view
+        returns (
+            int24 currentTick,
+            int24 fastOracleTick,
+            int24 slowOracleTick,
+            int24 latestObservation,
+            uint256 medianData
+        )
+    {
+        (currentTick, fastOracleTick, slowOracleTick, latestObservation, ) = _getOracleTicks();
```
flanagansteve, line 1522: This function seems fine, but i note that `_getOracleTicks` returns a `medianData` yet only sets it in when we are in `SLOW_ORACLE_UNISWAP_MODE`. Is it OK for it to be returning a nil medianData all other times?

Here's the implementation - maybe I'm missing something, but I only see it setting the `return` value `medianData` in one spot:

```
function _getOracleTicks()
        internal
        view
        returns (
            int24 currentTick,
            int24 fastOracleTick,
            int24 slowOracleTick,
            int24 latestObservation,
            uint256 medianData
        )
    {
        IUniswapV3Pool _univ3pool = s_univ3pool;
        uint16 observationIndex;
        uint16 observationCardinality;

        (, currentTick, observationIndex, observationCardinality, , , ) = _univ3pool.slot0();

        (fastOracleTick, latestObservation) = PanopticMath.computeMedianObservedPrice(
            _univ3pool,
            observationIndex,
            observationCardinality,
            FAST_ORACLE_CARDINALITY,
            FAST_ORACLE_PERIOD
        );

        if (SLOW_ORACLE_UNISWAP_MODE) {
            (slowOracleTick, ) = PanopticMath.computeMedianObservedPrice(
                _univ3pool,
                observationIndex,
                observationCardinality,
                SLOW_ORACLE_CARDINALITY,
                SLOW_ORACLE_PERIOD
            );
        } else {
            (slowOracleTick, medianData) = PanopticMath.computeInternalMedian(
                observationIndex,
                observationCardinality,
                MEDIAN_PERIOD,
                s_miniMedian,
                _univ3pool
            );
        }
    }
```

flanagansteve (approved):

test/foundry/core/Misc.t.sol
```diff
@@ -2141,8 +2141,8 @@ contract Misctest is Test, PositionUtils {
         // deposit bare minimum
         ct1.deposit(17_817, Bob);
 
-        // mint fails
-        vm.expectRevert(Errors.AccountInsolvent.selector);
+        // mint fails, not enough collateral
+        vm.expectRevert();
```
flanagansteve, line 2145: (I believe you fill in the exact `selector` to expect here in other PRs, but if not, would be nice to have)

contracts/PanopticPool.sol
guil-lambert: Good point. So it's OK to only use the outputted medianData where we do, but I think we should also use it when burning options. Will create a PR that also updates the miniMedian at burns.

PR #102, opened 2024-07-15
Hardening PR2: full collateralization

guil-lambert:
I messed up the merging on https://github.com/panoptic-labs/panoptic-v1-core-private/pull/82, this is the recovery branch

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -647,7 +647,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @param tickLimitLow The lower bound of an acceptable open interval for the ending price
     /// @param tickLimitHigh The upper bound of an acceptable open interval for the ending price
     /// @return Packing of the pool utilization (how much funds are in the Panoptic pool versus the AMM pool) at the time of minting,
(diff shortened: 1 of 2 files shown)


contracts/PanopticPool.sol
```diff
@@ -669,7 +669,11 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
         uint128 poolUtilizations = _payCommissionAndWriteData(tokenId, positionSize, totalSwapped);
 
-        return poolUtilizations;
+        if (safeMode) {
+            return uint128(10_000) + uint128(10_000 << 64);
```
flanagansteve, line 673: It might be good to update the return docstring:

>     /// @return Packing of the pool utilization (how much funds are in the Panoptic pool versus the AMM pool) at the time of minting,
>    /// right 64bits for token0 and left 64bits for token1

to also note this behaviour. Perhaps:

>     /// @return Packing of the pool utilization (how much funds are in the Panoptic pool versus the AMM pool) at the time of minting,
>     /// right 64bits for token0 and left 64bits for token1.
>     /// When safeMode is active, we return 100% pool utilisation for both tokens.

flanagansteve (approved):

PR #103, opened 2024-07-15
Hardening PR3: liquidation solvencies

guil-lambert:
I messed up the merging on https://github.com/panoptic-labs/panoptic-v1-core-private/pull/83, this is the recovery branch


contracts/PanopticPool.sol
```diff
@@ -1027,15 +1027,38 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+        int24 currentTick;
+        {
+            // Enforce maximum delta between TWAP and currentTick to prevent extreme price manipulation
+            int24 fastOracleTick;
+            int24 lastObservedTick;
+            (currentTick, fastOracleTick, , lastObservedTick, ) = _getOracleTicks();
+
+            if (Math.abs(currentTick - twapTick) > MAX_SLOW_FAST_DELTA) revert Errors.StaleTWAP();
+
+            // Ensure the accound is insolvent at twapTick, currentTick, and fastOracleTick
```
flanagansteve: Typo: accound -> account

flanagansteve (approved):

contracts/PanopticPool.sol
```diff
@@ -979,15 +979,31 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+        {
+            // Enforce maximum delta between TWAP and currentTick to prevent extreme price manipulation
+            int24 fastOracleTick;
+            int24 lastObservedTick;
+            (currentTick, fastOracleTick, , lastObservedTick, ) = _getOracleTicks();
+
+            if (Math.abs(currentTick - twapTick) > MAX_TICKS_DELTA) revert Errors.StaleTWAP();
+
+            // Ensure the accound is insolvent at twapTick, currentTick, and fastOracleTick
+            /// @dev do not check slowOracleTick because slow oracle is covered by twapTick
```
dyedm1: For future reference, natspec tags only work in function definitions

contracts/PanopticPool.sol
```diff
@@ -979,15 +979,31 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
         int24 twapTick = getUniV3TWAP();
 
+        int24 currentTick;
+        {
+            // Enforce maximum delta between TWAP and currentTick to prevent extreme price manipulation
+            int24 fastOracleTick;
+            int24 lastObservedTick;
+            (currentTick, fastOracleTick, , lastObservedTick, ) = _getOracleTicks();
+
+            if (Math.abs(currentTick - twapTick) > MAX_TICKS_DELTA) revert Errors.StaleTWAP();
```
```suggestion
	   unchecked {
	            if (Math.abs(currentTick - twapTick) > MAX_TWAP_DELTA_LIQUIDATION) revert Errors.StaleTWAP();
	    }
```

dyedm1: I don't think it makes sense to use `MAX_TICKS_DELTA` here; as discussed in the previous iteration of the PR this check serves an entirely different purpose than the fallback oracles in option mints/burns and increasing the threshold also increases the potential amount of extractable value leaked to liquidators.

dyedm1 (commented):
Looks good but concerned about increasing the stale TWAP threshold. The additional liquidation conditions required by this PR do not necessarily mitigate the price manipulation attacks that check prevents, so I think we should keep it at 5%.

PR #105, opened 2024-07-18
fix: increase container size for delegated shares

dyedm1:
The delegated amounts returned by the `CollateralTracker`  for token0/1 were originally placed in a LeftRight container to prevent stack-too-deep issues by saving a stack slot. The reasoning was that the amount of *assets* we delegate is `(2**104 - 1) * 10_000`, which fits comfortably within a `uint128` type. However, when the denomination was switched from *assets* to *shares* (to prevent issues caused by asset conversions after rounding errors are introduced into the share price), this ceased to be the case because the starting share price is 1M shares/asset, so the delegation amount is roughly `(2**104 - 1) * 10_000 * 1_000_000` for a fresh pool, a quantity that greatly exceeds the `uint128` container and causes most of the shares to remain unrevoked post-exercise.

This issue is impossible to detect with the legacy stateless fuzz test suite because the share price is edited to ~1 to prevent compatibility issues. However, the invariant test suite (ran at a natural, unmodified share price) instantly detects a discrepancy between the amount of shares the exercisee receives compared to a burn (most of the original delegation) and the shares contributed by the exercisor (0).

The solution is to separate the delegated amounts for each token into their own (uint256) stack variables, making any overflows impossible. I was able to shift around/scope some statements to accommodate the extra variables without having to stack roll or reread from storage.

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1157,50 +1157,54 @@ contract PanopticPool is ERC1155Holder, Multicall {
         // validate the exercisor's position list (the exercisee's list will be evaluated after their position is force exercised)
         _validatePositionList(msg.sender, positionIdListExercisor, 0);
 
+        uint128 positionBalance = s_positionBalance[account][touchedId[0]].rightSlot();
+
         int24 twapTick = getUniV3TWAP();
 
         // on forced exercise, the price *must* be outside the position's range for at least 1 leg
         touchedId[0].validateIsExercisable(twapTick);
 
-        // The protocol delegates some virtual shares to ensure the burn can be settled.
-        LeftRightUnsigned delegatedShares = LeftRightUnsigned
-            .wrap(0)
-            .toRightSlot(
-                uint128((s_collateralToken0.delegate(account, uint256(type(uint104).max) * 10_000)))
-            )
-            .toLeftSlot(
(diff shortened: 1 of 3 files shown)


contracts/PanopticPool.sol
```diff
@@ -1157,50 +1157,54 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+                longAmounts
+            );
+        }
 
-        // compute the notional value of the short legs (the maximum amount of tokens required to exercise - premia)
-        // and the long legs (from which the exercise cost is computed)
-        (LeftRightSigned longAmounts, ) = PanopticMath.computeExercisedAmounts(
-            touchedId[0],
-            positionBalance
+        // The protocol delegates some virtual shares to ensure the burn can be settled.
```
PeakyCryptos, line 1189: Perhaps add some documentation here as to why the the delegation amount is set to 104 bits * 10_000.

- Reduces calculation and estimation risks by requiring force exercises to have a fixed-size delegation of DEPOSIT_LIMIT * 10,000

- Virtual delegation allows positions to settle before exercisor provides requisite tokens without breaking accounting

-  Operations requiring ~10,000 distinct deposit calls are effectively DoS

PeakyCryptos (approved):

PR #108, opened 2024-07-18
Moved getOracleTicks to PanopticMath

guil-lambert:
This is partly to reduce the bytecode size of PanopticPool.sol, and partly to simplify the calls to PanopticMath when getting the oracle ticks.

- Bytecode size:
  - `main`: `24,532kb`
  - `this branch`: `24,247kb` 
  - = `285b` savings
- PanopticMath external calls in _getOracleTicks:
  - `main` (2):
    - `PanopticMath.computeMedianObservedPrice` and 
    - `PanopticMath.computeInternalMedian`)
  - `this branch` (1):
    - `PanopticMath.getOracleTicks()`

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -117,12 +117,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @notice The minimum window (in seconds) used to calculate the TWAP price for solvency checks during liquidations.
     uint32 internal constant TWAP_WINDOW = 600;
 
-    /// @notice Parameter that determines which oracle type to use for the "slow" oracle price on non-liquidation solvency checks.
-    /// @dev If false, an 8-slot internal median array is used to compute the "slow" oracle price.
-    /// @dev This oracle is updated with the last Uniswap observation during `mintOptions` if MEDIAN_PERIOD has elapsed past the last observation.
-    /// @dev If true, the "slow" oracle price is instead computed on-the-fly from 8 Uniswap observations (spaced 5 observations apart) irrespective of the frequency of `mintOptions` calls.
-    bool internal constant SLOW_ORACLE_UNISWAP_MODE = false;
-
     /// @notice The minimum amount of time, in seconds, permitted between internal TWAP updates.
     uint256 internal constant MEDIAN_PERIOD = 60;
 
@@ -134,13 +128,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @dev Uniswap observations snapshot the last block's closing price at the first interaction with the pool in a block.
     /// @dev In this case, if there is an interaction every block, the "fast" oracle can consider 3 consecutive block end prices (min=36 seconds on Ethereum).
     uint256 internal constant FAST_ORACLE_PERIOD = 1;
-
-    /// @notice Amount of Uniswap observations to include in the "slow" oracle price (in Uniswap mode).
-    uint256 internal constant SLOW_ORACLE_CARDINALITY = 8;
-
-    /// @notice Amount of observation indices to skip in between each observation for the "slow" oracle price.
-    /// @dev Structured such that the minimum total observation time is 8 minutes on Ethereum (similar to internal median mode).
-    uint256 internal constant SLOW_ORACLE_PERIOD = 5;
 
     /// @notice The maximum allowed delta between the currentTick and the Uniswap TWAP tick during a liquidation (~5% down, ~5.26% up).
     /// @dev Mitigates manipulation of the currentTick that causes positions to be liquidated at a less favorable price.
@@ -499,14 +486,12 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
     /// @notice Updates the internal median with the last Uniswap observation if the MEDIAN_PERIOD has elapsed.
     function pokeMedian() external {
-        (, , uint16 observationIndex, uint16 observationCardinality, , , ) = s_univ3pool.slot0();
-
-        (, uint256 medianData) = PanopticMath.computeInternalMedian(
-            observationIndex,
-            observationCardinality,
+        (, , , , uint256 medianData) = PanopticMath.getOracleTicks(
+            s_univ3pool,
+            FAST_ORACLE_CARDINALITY,
+            FAST_ORACLE_PERIOD,
             MEDIAN_PERIOD,
-            s_miniMedian,
-            s_univ3pool
+            s_miniMedian
         );
 
         if (medianData != 0) s_miniMedian = medianData;
@@ -875,7 +860,13 @@ contract PanopticPool is ERC1155Holder, Multicall {
             int24 slowOracleTick,
             int24 lastObservedTick,
             uint256 _medianData
-        ) = _getOracleTicks();
+        ) = PanopticMath.getOracleTicks(
+                s_univ3pool,
+                FAST_ORACLE_CARDINALITY,
+                FAST_ORACLE_PERIOD,
+                MEDIAN_PERIOD,
+                s_miniMedian
+            );
 
         medianData = _medianData;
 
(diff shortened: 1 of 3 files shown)


contracts/libraries/PanopticMath.sol
```diff
@@ -144,6 +144,59 @@ library PanopticMath {
...
+
+    /// @notice Gets several ticks from Uniswap regarding the underlying pair.
+    /// @return currentTick The current tick in the Uniswap pool (as returned in slot0)
+    /// @return fastOracleTick The fast oracle tick computed as the median of the past N observations in the Uniswap Pool
+    /// @return slowOracleTick The slow oracle tick as tracked by `s_miniMedian`
+    /// @return latestObservation The latest observation from the Uniswap pool (price at the end of the last block)
+    /// @return medianData the updated value for `s_miniMedian` (returns 0 if not enough time has passed since last observation)
+    function getOracleTicks(
+        IUniswapV3Pool univ3pool,
+        uint256 FAST_ORACLE_CARDINALITY,
```
dyedm1, line 159: We should move these constants into Constants.sol and access them directly or inline them here to reduce overhead.

contracts/PanopticPool.sol
```diff
@@ -499,14 +486,12 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
     /// @notice Updates the internal median with the last Uniswap observation if the MEDIAN_PERIOD has elapsed.
     function pokeMedian() external {
-        (, , uint16 observationIndex, uint16 observationCardinality, , , ) = s_univ3pool.slot0();
-
-        (, uint256 medianData) = PanopticMath.computeInternalMedian(
```
dyedm1, line 504: `computeInternalMedian` is still the most cost-effective option to use here. If you're having contract size issues after making the function internal, you can opt to make it `public` instead (which will perform intra-library calls from `getOracleTicks` inline but otherwise behave similarly to an `external` library function)

contracts/libraries/PanopticMath.sol
```diff
@@ -207,7 +260,7 @@ library PanopticMath {
         uint256 period,
         uint256 medianData,
         IUniswapV3Pool univ3pool
-    ) external view returns (int24 medianTick, uint256 updatedMedianData) {
+    ) internal view returns (int24 medianTick, uint256 updatedMedianData) {
```
dyedm1, line 263: See above comment, you may want to consider making this function `public`

test/foundry/core/CollateralTracker.t.sol
```diff
@@ -125,24 +125,24 @@ contract PanopticPoolHarness is PanopticPool {
...
 
         unchecked {
             s_miniMedian =
                 (uint256(block.number) << 216) +
                 // magic number which adds (7,5,3,1,0,2,4,6) order and minTick in positions 7, 5, 3 and maxTick in 6, 4, 2
                 // see comment on s_miniMedian initialization for format of this magic number
                 (uint256(0xF590A6F276170D89E9F276170D89E9F276170D89E9000000000000)) +
-                (uint256(uint24(slowOracleTick)) << 24) + // add to slot 4
-                (uint256(uint24(slowOracleTick))); // add to slot 3
+                (uint256(uint24(currentTick)) << 24) + // add to slot 4
```
dyedm1, line 144: Nice catch!

contracts/libraries/PanopticMath.sol
```diff
@@ -144,6 +144,59 @@ library PanopticMath {
...
+            uint256 medianData
+        )
+    {
+        uint16 observationIndex;
+        uint16 observationCardinality;
+
+        IUniswapV3Pool _univ3pool = univ3pool;
+        (, currentTick, observationIndex, observationCardinality, , , ) = univ3pool.slot0();
+
+        {
```
dyedm1, line 180: I don't think this scoping does anything?

contracts/libraries/PanopticMath.sol
```diff
@@ -144,6 +144,59 @@ library PanopticMath {
         }
     }
 
+    /*//////////////////////////////////////////////////////////////
+                          ORACLE CALCULATIONS
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Gets several ticks from Uniswap regarding the underlying pair.
```
dyedm1, line 151: Param natspec?

contracts/PanopticPool.sol
```diff
@@ -117,12 +117,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @notice The minimum window (in seconds) used to calculate the TWAP price for solvency checks during liquidations.
     uint32 internal constant TWAP_WINDOW = 600;
 
-    /// @notice Parameter that determines which oracle type to use for the "slow" oracle price on non-liquidation solvency checks.
-    /// @dev If false, an 8-slot internal median array is used to compute the "slow" oracle price.
-    /// @dev This oracle is updated with the last Uniswap observation during `mintOptions` if MEDIAN_PERIOD has elapsed past the last observation.
-    /// @dev If true, the "slow" oracle price is instead computed on-the-fly from 8 Uniswap observations (spaced 5 observations apart) irrespective of the frequency of `mintOptions` calls.
-    bool internal constant SLOW_ORACLE_UNISWAP_MODE = false;
```
dyedm1, line 124: I don't see any harm in keeping this functionality so we can use it if we want to down the line. I think I remember from our discussion in standup that the reason for removing this was too many parameters being passed into `getOracleTicks` (causing a stack-too-deep (?)), but that becomes a nonissue if you move the constants to `Constants.sol` and access them directly from `getOracleTicks` or inline them as I suggested in another comment.

contracts/libraries/PanopticMath.sol
```diff
@@ -112,7 +112,7 @@ library PanopticMath {
     }
 
     /*//////////////////////////////////////////////////////////////
-                          ORACLE CALCULATIONS
+                              UTILITIES
```
dyedm1, line 115: Not sure why this was in the oracle calcs section, good catch lol! I think we can just put this under the generic MATH HELPERS category though.

contracts/PanopticPool.sol
```diff
@@ -1502,7 +1443,14 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-        (currentTick, fastOracleTick, slowOracleTick, latestObservation, ) = _getOracleTicks();
+        (currentTick, fastOracleTick, slowOracleTick, latestObservation, ) = PanopticMath
+            .getOracleTicks(
+                s_univ3pool,
+                FAST_ORACLE_CARDINALITY,
+                FAST_ORACLE_PERIOD,
+                MEDIAN_PERIOD,
+                s_miniMedian
+            );
         medianData = s_miniMedian;
```
dyedm1, line 1454: I think you may save some contract size by removing this `getOracleTicks` function and just making `s_miniMedian` public or adding a getter for it so contracts can just call the PanopticMath function (& the UI can call a PanopticHelper wrapper).

dyedm1 (changes requested):
Suggested a few improvements, LGTM otherwise

contracts/PanopticPool.sol
flanagansteve: I definitely agree that this is worth keeping for now

flanagansteve (commented):
Henry touched on everything i would mention - can re-review if you pursue his suggestion of retaining the SLOW_ORACLE_UNISWAP_MODE functionality as i imagine the diff will look quite different

contracts/libraries/PanopticMath.sol
guil-lambert: Added all of them to Constants

contracts/PanopticPool.sol
guil-lambert: I added `computeInternalMedian` back

contracts/PanopticPool.sol
guil-lambert: Ok, I've kept it

contracts/PanopticPool.sol
guil-lambert: I'll keep this in as is for now, may remove it if bytecode is limiting

PR #109, opened 2024-07-18
Poke median at burn

guil-lambert:
Currently, the s_miniMedian only gets updated when a new position is minted here: https://github.com/panoptic-labs/panoptic-v1-core-private/blob/6f3ca107f42c3f1081e68d6e3d15cdb91977c759/contracts/PanopticPool.sol#L645

We should also update it when a position is burnt too, otherwise the slow oracle will fail to be updated in a timely manner when people are mostly closing positions --eg. in periods of volatilities when people could be closing only.

I added the same `medianData` logic in the burnOptions workflow.

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -552,7 +552,10 @@ contract PanopticPool is ERC1155Holder, Multicall {
     ) external {
         _burnOptions(COMMIT_LONG_SETTLED, tokenId, msg.sender, tickLimitLow, tickLimitHigh);
 
-        _validateSolvency(msg.sender, newPositionIdList, NO_BUFFER);
+        uint256 medianData = _validateSolvency(msg.sender, newPositionIdList, NO_BUFFER);
+
+        // Update `s_miniMedian` with a new observation if the last observation is old enough (returned medianData is nonzero)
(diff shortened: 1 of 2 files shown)


contracts/PanopticPool.sol
```diff
@@ -552,7 +552,10 @@ contract PanopticPool is ERC1155Holder, Multicall {
     ) external {
         _burnOptions(COMMIT_LONG_SETTLED, tokenId, msg.sender, tickLimitLow, tickLimitHigh);
 
-        _validateSolvency(msg.sender, newPositionIdList, NO_BUFFER);
+        uint256 medianData = _validateSolvency(msg.sender, newPositionIdList, NO_BUFFER);
```
dyedm1, line 555: @flanagansteve just a heads-up that you'll have to edit your Echidna wrapper for burn_multiple to poke before your simulation/call (otherwise, behavior will diverge from burn_multiple if more than 60s has passed since the last poke because collateral checks will be at a non-poked tick in burn_multiple but at a poked tick after the first burn_one). If you want to add a check for the poking itself you can test that in a separate simulation of burn_multiple before your poke or in a different wrapper.

dyedm1 (approved):
LGTM!

PR #111, opened 2024-07-26
feat: segregate long and short premium in collateral requirements

dyedm1:
`
If an account’s premium is net positive, it is included in their balance, and if net negative, their collateral requirement. This means that in situations where a buffer is applied to the collateral requirement (minting, withdrawing, etc), long premium netted against short premium will have a lower effective collateral requirement (given that the short premium reduces the collateral requirement which is subject to that multiplier) than if the short premium was collected and added to their balance.
`

Fix: return separate net long and short premium values from `_calculateAccumulatedPremia`, adding long premium to the collateral requirement and short premium to balance. This means that `BP_DECREASE_BUFFER` will now be applied to *all* long premium regardless of how much can be netted against a users's (prorated) short premium.

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1165,7 +1165,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             tokenRequired = _getTotalRequiredCollateral(atTick, positionBalanceArray) + longPremium;
         }
 
-        uint256 netBalance = convertToAssets(balanceOf[user]) + uint128(shortPremium);
+        uint256 netBalance = convertToAssets(balanceOf[user]) + shortPremium;
 
         // store assetBalance and tokens required in tokenData variable
         tokenData = tokenData.toRightSlot(netBalance.toUint128()).toLeftSlot(
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -543,8 +543,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         );
 
         emit Withdraw(msg.sender, receiver, owner, assets, shares);
-
-        return shares;
     }
 
     /// @notice Redeem the amount of shares required to withdraw the specified amount of assets.
@@ -588,8 +586,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         );
 
         emit Withdraw(msg.sender, receiver, owner, assets, shares);
-
-        return shares;
     }
 
     /// @notice Returns the maximum amount of shares that can be redeemed for a given user.
@@ -648,8 +644,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         );
 
         emit Withdraw(msg.sender, receiver, owner, assets, shares);
-
-        return assets;
     }
 
     /*//////////////////////////////////////////////////////////////
@@ -1148,7 +1142,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param positionBalanceArray The list of all historical positions held by the 'optionOwner', stored as [[tokenId, balance/poolUtilizationAtMint], ...]
     /// @param shortPremium The total amount of premium (prorated by available settled tokens) owed to the short legs of `user`
     /// @param longPremium The total amount of premium owed by the long legs of `user`
-    /// @return tokenData Information collected for the tokens about the health of the account
+    /// @return Information collected for the tokens about the health of the account
     /// The collateral balance of the user is in the right slot and the threshold for margin call is in the left slot.
     function getAccountMarginDetails(
         address user,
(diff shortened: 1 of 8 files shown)


PeakyCryptos (approved):
LGTM - I'll give another pass for comment changes

contracts/CollateralTracker.sol
```diff
@@ -1144,57 +1144,28 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
             // get all collateral required for the incoming list of positions
-            tokenRequired = _getTotalRequiredCollateral(atTick, positionBalanceArray);
-
-            // If premium is negative (ie. user has to pay for their purchased options), add this long premium to the token requirement
-            if (premiumAllPositions < 0) {
-                unchecked {
-                    tokenRequired += uint128(-premiumAllPositions);
-                }
-            }
+            tokenRequired = _getTotalRequiredCollateral(atTick, positionBalanceArray) + longPremium;
```
PeakyCryptos: You removed the unchecked blocks here. Did you reach a state in the echidna tests where this would revert due to it overflowing the uint128 container? or revert after applying the buffer

contracts/CollateralTracker.sol
```diff
@@ -1144,57 +1144,28 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
         }
 
-        // if premium is positive (ie. user will receive funds due to selling options), add this premum to the user's balance
-        uint256 netBalance = convertToAssets(balanceOf[user]);
-        if (premiumAllPositions > 0) {
-            unchecked {
-                netBalance += uint256(uint128(premiumAllPositions));
-            }
-        }
+        uint256 netBalance = convertToAssets(balanceOf[user]) + shortPremium;
```
PeakyCryptos: ^ same comment as above, minus the concern about the buffer

contracts/CollateralTracker.sol
dyedm1: Yeah I don't think these necessarily have to be checked (not really possible for it to overflow the uint256 container - the real check occurs when it's downcast either way).

contracts/CollateralTracker.sol
```diff
@@ -1144,63 +1138,31 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
-        // if premium is positive (ie. user will receive funds due to selling options), add this premum to the user's balance
-        uint256 netBalance = convertToAssets(balanceOf[user]);
-        if (premiumAllPositions > 0) {
-            unchecked {
-                netBalance += uint256(uint128(premiumAllPositions));
-            }
+        uint128 shortPremium,
+        uint128 longPremium
+    ) public view returns (LeftRightUnsigned) {
+        unchecked {
```
flanagansteve, line 1154: What is the reason to remove the if (positionBalanceArray.length > 0)  check? Simply a different logic style choice (e.g. put the onus on the caller not to supply a zero-length array as an arg) or is there a business logic change that makes the check unnecessary?)

contracts/PanopticPool.sol
```diff
@@ -434,28 +441,44 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-                    bytes32 chunkKey = keccak256(
-                        abi.encodePacked(
-                            tokenId.strike(leg),
-                            tokenId.width(leg),
-                            tokenId.tokenType(leg)
-                        )
-                    );
+                if (tokenId.isLong(leg) == 0) {
+                    if (!includePendingPremium) {
+                        bytes32 chunkKey = keccak256(
```
flanagansteve, line 446: We do this chunkKey = keccack256(abi.encodePacked()) a lot - is it worth kicking that to a helper method?

contracts/PanopticPool.sol
```diff
@@ -401,15 +399,24 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @param positionIdList The list of all option positions held by user
     /// @param computeAllPremia Whether to compute accumulated premia for all legs held by the user (true), or just owed premia for long legs (false)
     /// @param includePendingPremium If true, include premium that is owed to the user but has not yet settled; if false, only include premium that is available to collect
-    /// @return portfolioPremium The computed premia of the user's positions, where premia contains the accumulated premia for token0 in the right slot and for token1 in the left slot
+    /// @return shortPremium The total amount of premium owed (according to `includePendingPremium`) to the short legs in `positionIdList` (token0: right slot, token1: left slot)
```
flanagansteve: Very small nit - I'd suggest a slight rewording to:

> /// @return shortPremium The total amount of premium owed (which may `includePendingPremium`) to the short legs in `positionIdList` (token0: right slot, token1: left slot)

or something like that - current wording doesn't quite make it sound like includePendingPremium  is a boolean that decides whether to include the pending premium

flanagansteve (approved):
Great stuff!

contracts/CollateralTracker.sol
dyedm1: The early-exit doesn't really make sense -- the majority of calls to this function will not have empty lists and the alternative behavior is just to go through the empty for loop which has a negligible cost.

contracts/PanopticPool.sol
dyedm1: We could, but it's a flat enough oneliner that I don't think it's necessary. Plus, since we don't compile with via-IR, there is extra jump overhead every time we do that since internal function calls are not inlined.

PR #112, opened 2024-07-26
feat: SFPM optimizations

dyedm1:
- `s_accountFeesBase` is defined as `feeGrowthInsideLastX128 * netLiquidity / 2^128`. Thus, if an account is verified to have `s_accountLiquidity = (removedLiquidity << 128) + netLiquidity = 0`, `s_accountFeesBase` will necessarily be zero, making the second expression in `(LeftRightUnsigned.unwrap(s_accountLiquidity[positionKey_to]) != 0) ||(LeftRightSigned.unwrap(s_accountFeesBase[positionKey_to]) != 0)` redundant.
- All positions are validated at mint, so `tokenId.validate()` calls during transfers (and burns that go through the same pattern at mint) are redundant. You can transfer `0` of an invalid tokenId, but that doesn't induce any state changes beyond an event being emitted.
- The reentrancy lock implementation may not be optimal for most SFPM calls except accounts that burn all tokens/liquidity for a chunk in a single call over a zero -> nonzero -> zero pattern due to gas refunds (SFPM transactions are expensive and other nonzero -> zero SSTOREs seem to be rare outside of complete option/chunk burns for a single SFPM account). From the Panoptic perspective most users will be trading in the same chunks provided by market makers, so we should expect the state to be optimally configured most of the time. 
- Redundant `OptionsBalanceZero` check (subset of ZeroLiquidity positions)
- Simplified position creation flow and removed extreme stack rolling 
- Removed unnecessary LIFO behavior (I opted to keep this in a previous PR, but now that we have removed the stack rolling we would have do additional changes/caching for the LIFO burning to continue to work correctly. I don't think we'll ever reenable non-unique-chunk positions, so no point in keeping it)


contracts/SemiFungiblePositionManager.sol
```diff
@@ -370,17 +347,14 @@ contract SemiFungiblePositionManager is ERC1155, Multicall {
         // There are 281,474,976,710,655 possible pool patterns.
         // A modern GPU can generate a collision such a space relatively quickly,
```
flanagansteve: Missing word in this comment - "// A modern GPU can generate a collision such a space relatively quickly," should likely be "// A modern GPU can generate a collision *in* such a space relatively quickly,"

flanagansteve (approved):
I think everything here makes sense. It touches some bits i'm less familiar with but it makes sense from my naive angle at least.

I tried to think through some possible implications of the lack of tokenID validation - you call out that users can transfer invalid tokenIDs (but they'll have 0 balance of them). I also think you can execute a 0-position-size burn of invalid tokenIDs, but i don't think you can benefit from that. You could make a cleverly crafted invalid tokenID to get through to the _createLegInAMM step (i think) but will fail when we see you have 0 liquidity in the chunk - so i think we're all good

PeakyCryptos (approved):
LGTM - do we want to revisit this: https://panoptic-global.slack.com/archives/C04F33PCH0S/p1705607664035079

PR #113, opened 2024-07-29
Feat: accumulator assertions when minting

flanagansteve:


contracts/fuzz/echidna/FuzzHelpers.sol
```diff
@@ -309,6 +309,17 @@ contract FuzzHelpers is PropertiesAsserts {
         uint256 tokenType;
     }
 
+    struct AccumulatorsForLeg {
```
flanagansteve, line 312: (This is a shared type with the burn tests)

contracts/fuzz/echidna/FuzzHelpers.sol
```diff
@@ -424,6 +435,14 @@ contract FuzzHelpers is PropertiesAsserts {
     int256 $commission0;
     int256 $commission1;
 
+    uint128 $postmintSettledToken0;
```
flanagansteve, line 438: (And the storage vars added for this test also shadow names of the burn tests - can either keep them shared or change their names slightly if we want to avoid faultiness due to failures to clear some of them)

contracts/fuzz/echidna/FuzzHelpers.sol
```diff
@@ -1404,6 +1424,120 @@ contract FuzzHelpers is PropertiesAsserts {
...
+
+        delete $settledToken0;
+        delete $settledToken1;
+        delete $grossPremiaLast0;
+        delete $grossPremiaLast1;
+        delete $sfpmGrossPremiaAccumulator0;
+        delete $sfpmGrossPremiaAccumulator1;
+    }
+
+    function _get_total_short_liquidity(
```
flanagansteve, line 1495: (And finally, nothing to change here, but wanted to note that these 3 _get_ helper methods are mostly shared with the burn tests - only difference is the burn tests have this accept a `TokenId position` arg. Easy to consolidate these if we like, especially if we unified the burn and mint tests to use the same TokenID storage var for the position they're operating on - burn tests call this $position while mint tests call it $tokenIdActive)

dyedm1:
Looks good, implemented a slightly simpler version of this here where total gross premium is calculated before minting and recalculated post-mint (with the actual liquidity values at that time): #99

PR #115, opened 2024-08-01
Remove getPortfolioValue from PanopticPool

guil-lambert:
Not necessary to leave that function in PanopticPool, will move that logic into PanopticHelper. 

It was initially put in PanopticPool to use `userBalances` as an input, but that's something we can fetch in PanopticHelper using `pool.calculateAccumulatedFeesBatch`


dyedm1 (approved):
LGTM

PR #120, opened 2024-08-12
fix: remove redundant finalPrice input on liquidation bonuses

dyedm1:


flanagansteve (approved):
I interpret this PR's changes as:

1. Make the `PanopticMath.getLiquidationBonus` method more generic, by changing its two price tick arguments with explicit designations - TWAP and Final - to instead just accept any arbitrary price as one argument used throughout the method.
2. Make the `liquidate` method pass in a TWAP for the price to evaluate solvency at / determine bonus amounts with in `PanopticMath.getLiquidationBonus`
3. 1 and 2 in combination have the effect of changing our economics such that the liquidation bonuses are now calculated with the TWAP _instead_ of the final (aka current) price. Solvency evaluation remains the same.
4. Less pertinently - make PanopticMath.haircutPremia similarly generic, accepting any price instead of an argument designated by its name as the Final price.
  a. Do note that the call from `.liquidate` to `.haircutPremia` [does pass in a TWAP and not a Final price](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/main/contracts/PanopticPool.sol#L1002)! So the old name was inaccurate to the usage of the method, it appears. Unsure if that's something we need to follow up on.

So overall, these changes sound good, if the economic change in (3) is what we intend. I actually believe we discussed that change in a standup a few weeks ago, i suppose we never got around to executing it? Or maybe I'm misremembering?

EDIT: Reading through getLiquidationBonus, i actually think it doesn't evaluate solvency at all - which makes sense given its name - so reinterpret my comment in number (2) to instead state: "2. Make the `liquidate` method pass in a TWAP for the price to determine bonus amounts / current balances with in `PanopticMath.getLiquidationBonus`"

However, this does highlight that the docstring for the new unified price arg is probably inaccurate:

> /// @param atSqrtPriceX96 The oracle price used to swap tokens between the liquidator/liquidatee and determine solvency for the liquidatee

as it doesn't determine solvency in getLiquidationBonus

dyedm1:
> I actually believe we discussed that change in a standup a few weeks ago, i suppose we never got around to executing it? Or maybe I'm misremembering?

There are actually no new economic changes in this PR, it's just that 3) was effectuated by passing in two of the same (TWAP) prices to `getLiquidationBonus` (with 1 being in the current price input) instead of modifying the function itself to take 1 price input.

dyedm1:
> However, this does highlight that the docstring for the new unified price arg is probably inaccurate:

It is still used to evaluate/interpret solvency *for the purposes of the liquidation bonus calculation*.

flanagansteve (approved):
You are right on both counts - looks good to me!

PR #121, opened 2024-08-13
fix: always return accumulated gross/owed premium when `netLiquidity == 0`

dyedm1:
The SFPM skips all calculation and returns 0 for the gross/owed premium if an attempt is made to calculate premium up to the current block while `netLiquidity` is equal to 0. This behavior is needed given that `netLiquidity` is used as a denominator in the `_getPremiaDeltas()` function, but it is inconsistent with the nature of `grossPremium`/`owedPremium` as accumulators and query results in other states. The `PanopticPool` does not trigger this behavior since premium-up-to-current-block is only ever computed for positions that still have liquidity in them (i.e. minted positions in collateral checks, positions-to-burn preburn), but this could potentially cause overflows in other protocols that use the SFPM and rely on the assumption that `grossPremium` and `owedPremium` are monotonically increasing accumulators.

To resolve this we can just return the current values of `grossPremium` and `owedPremium` while skipping the premium calculation. This is accurate because `netLiquidity == 0` necessarily implies that no premium has been earned (given that premium calculations are based on fees earned by the SFPM, which has had 0 liquidity in the Uniswap pool since the last accumulator update).

### DIFF
--- a/contracts/SemiFungiblePositionManager.sol
+++ b/contracts/SemiFungiblePositionManager.sol
@@ -1298,58 +1298,57 @@ contract SemiFungiblePositionManager is ERC1155, Multicall, TransientReentrancyG
 
         LeftRightUnsigned acctPremia;
 
+        LeftRightUnsigned accountLiquidities = s_accountLiquidity[positionKey];
+        uint128 netLiquidity = accountLiquidities.rightSlot();
+
         // Compute the premium up to the current block (ie. after last touch until now). Do not proceed if atTick == type(int24).max = 8388608
-        if (atTick < type(int24).max) {
+        if (atTick < type(int24).max && netLiquidity != 0) {
             // unique key to identify the liquidity chunk in this uniswap pool
-            LeftRightUnsigned accountLiquidities = s_accountLiquidity[positionKey];
-            uint128 netLiquidity = accountLiquidities.rightSlot();
-            if (netLiquidity != 0) {
-                LeftRightUnsigned amountToCollect;
-                {
-                    IUniswapV3Pool _univ3pool = IUniswapV3Pool(univ3pool);
-                    int24 _tickLower = tickLower;
-                    int24 _tickUpper = tickUpper;
+            LeftRightUnsigned amountToCollect;
+            {
+                IUniswapV3Pool _univ3pool = IUniswapV3Pool(univ3pool);
+                int24 _tickLower = tickLower;
+                int24 _tickUpper = tickUpper;
 
-                    // how much fees have been accumulated within the liquidity chunk since last time we updated this chunk?
-                    // Compute (currentFeesGrowth - oldFeesGrowth), the amount to collect
-                    // currentFeesGrowth (calculated from FeesCalc.calculateAMMSwapFeesLiquidityChunk) is (ammFeesCollectedPerLiquidity * liquidityChunk.liquidity())
-                    // oldFeesGrowth is the last stored update of fee growth within the position range in the past (feeGrowthRange*liquidityChunk.liquidity()) (s_accountFeesBase[positionKey])
-                    LeftRightSigned feesBase = FeesCalc.calculateAMMSwapFees(
-                        _univ3pool,
-                        atTick,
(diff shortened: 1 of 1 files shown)


flanagansteve (approved):
> The PanopticPool does not trigger this behavior since premium-up-to-current-block is only ever computed for positions that still have liquidity in them (i.e. minted positions in collateral checks, positions-to-burn preburn)

Double checking this: I see three usages of `getAccountPremium`:

- In `PanopticPool._getPremia`, we use it to populate `premiumAccumulatorsByLeg` which in turn is used as part of a numerator - a numerator that is multiplied by liquidityChunk.liquidity(). We can therefore confirm that the behaviour is unchanged there - liquidityChunk.liquidity() will zero out the numerator so the premiumAccumulatorsByLeg doesn't matter
- In `PanopticPool.settleLongPremium`, we use the return values from this method to a) update `s_options`, and b) calculate an `accumulatedPremium` as this method's return values each subtracted by the pre-settleLongPremium-call s_options. In turn, those values are also used as parts of numerators that are multiplied by `liquidityChunk.liquidity()`, so those values should be fine - the numerators will still be zero'ed out like they were before. So: the only change is that before we would update PanopticPool.s_options to 0 when a settleLongPremium call occurred & the corresponding SFPM.s_accountLiquidity had 0 liquidity; now, it will be updated to the current value of grossPremium / owedPremium. Is that change OK / have you considered the impacts?
  - (It's possible there's a revert after the s_options write in settleLongPremium that makes it so actually we _always_ were keeping s_options at the last value SFPM.s_accountLiquidity had pre-zero-liquidity - in which case there's nothing to check here)
- In `PanopticPool._updateSettlementPostMint`, we also update s_options and use this value as a numerator that is multiplied by positionLiquidity - so, same considerations as settleLongPremium: No change to premia calculations (still a numerator of 0 if position liquidity is 0, which it should be when SFPM net liquidity is 0); but, possible change to our s_options bookkeeping - may have been getting zero'ed when SFPM net liquidity is 0 before, unless there's a revert that rolls back state changes if SFPM net liquidity is 0

----

Assuming the above is all set - the changes look good to me!

dyedm1:
Yes,
1/ is guaranteed to have nonzero netliquidity because the positions being queried themselves represent liquidity in a chunk
2/ is similarly guaranteed to have nonzero netliquidity
3/ is not calculated up to the current block, so this change is irrelevant

flanagansteve:
> Yes, 1/ is guaranteed to have nonzero netliquidity because the positions being queried themselves represent liquidity in a chunk 2/ is similarly guaranteed to have nonzero netliquidity 3/ is not calculated up to the current block, so this change is irrelevant

Ah! It clicked for me now. 

- With 1 and 2 - the `_validatePositionList` checks the supplied positions' presence in `s_positionsHash`, and positions can only get added in there by providing non-zero liquidity in the chunk, like you're saying
  - Proof to myself: `s_positionsHash` only gets modified in `_updateSettlementPostMint/Burn`'s  nested calls to the `_updatePositionsHash` helper; lets assume that we are indeed using that helper correctly, and only adding positions to the hash in the Mint one, and removing in the Burn. If that's true, then tracing outward: you can only get to _`_updateSettlementPostMint`_ by passing the `SFPM.mintTokenizedPosition` call, which will, in turn, revert if the PanopticMath.getLiquidityChunk representation of that position's liquidity is 0: https://github.com/panoptic-labs/panoptic-v1-core-private/blob/main/contracts/SemiFungiblePositionManager.sol#L854
    - (My earlier tripping point was thinking liquidity chunk referred to the group-level liquidity at a given tick range, not an individual position's. Digging into the PanopticMath call helped unlock that. I _think_ - but am not sure that - we may be using the LiquidityChunk term to sometimes refer to a group-level amount of liquidity too; if that's the case i can try to track down those references and suggest splitting the concepts up into PositionLiquidity and GroupLiquidity or something like that)
- With 3, re-read the args being passed in and the comment right before the getAccountPremium call - i'm on the same page now!

All sounds great thanks for walking me through it 👍

PR #122, opened 2024-08-22
Store ticks in `s_positionBalance`

guil-lambert:
This PR stores a ticks snapshot in `s_positionBalance`, this is something we need in `PanopticHelper` when computing the maintenance margin requirement.

It may also make it easier to get the `tickAtMint` than in the subgraph (ie. no need to call it anymore), and that tick information could also be used to plot svg PnL charts from a RPC call from `PanopticHelper`.

There is enough free space in that storage slot, and it shouldn't add much more gas for mints and extra packing/unpacking logic. So it should be straightforward to merge.

More specifically, `s_positionBalance` is currently structured like this:
```
    /// @notice Tracks the amount of liquidity for a user+tokenId (right slot) and the initial pool utilizations when that position was minted (left slot).
    //    poolUtilizations when minted (left)    liquidity=ERC1155 balance (right)
    //        token0          token1
    //  |<-- 64 bits -->|<-- 64 bits -->|<---------- 128 bits ---------->|
    //  |<-------------------------- 256 bits -------------------------->|
    mapping(address account => mapping(TokenId tokenId => LeftRightUnsigned balanceAndUtilizations))
        internal s_positionBalance; 
```

This PR reduces the utilizations to `uint16` (since they're never over 10,000) and I use the remaining 96bits to store the 4 ticks returned by getOracleTicks:

```
    //  |<- tickData->|<-- util1 -->|<-- util0 -->|<-------- 128 bits -------->|             
    //  |<- 96 bits ->|<- 16 bits ->|<- 16 bits ->|<-------- 128 bits -------->|
    //  |<------------------------------ 256 bits ---------------------------->|
```

Where tickData is:
```
|<- lastObservedTick ->|<- slowOracleTick ->|<- fastOracleTick ->|<- currentTick ->|
|<----- 24 bits ------>|<---- 24 bits ----->|<---- 24 bits ----->|<--- 24 bits --->|
```

I created a new PositionBalance type that facilitates the storage+retrieval of all those quantities. And I also exposed all those quantities in `PanopticPool.positionData(user, tokenId)`

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1187,9 +1187,9 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             uint128 positionSize = PositionBalance.wrap(positionBalanceArray[i][1]).positionSize();
 
             // read the pool utilization at mint
-            uint32 poolUtilization = PositionBalance
-                .wrap(positionBalanceArray[i][1])
-                .utilizations();
+            int16 poolUtilization = s_underlyingIsToken0
+                ? int16(PositionBalance.wrap(positionBalanceArray[i][1]).utilization0())
+                : int16(PositionBalance.wrap(positionBalanceArray[i][1]).utilization1());
 
             // Get tokens required for the current tokenId (a single active position)
             uint256 _tokenRequired = _getRequiredCollateralAtTickSinglePosition(
@@ -1220,7 +1220,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         TokenId tokenId,
         uint128 positionSize,
         int24 atTick,
-        uint32 poolUtilization
+        int16 poolUtilization
     ) internal view returns (uint256 tokenRequired) {
         bool underlyingIsToken0 = s_underlyingIsToken0;
         uint256 numLegs = tokenId.countLegs();
@@ -1253,7 +1253,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         uint256 index,
         uint128 positionSize,
         int24 atTick,
-        uint32 poolUtilization
+        int16 poolUtilization
     ) internal view returns (uint256 required) {
         return
             tokenId.riskPartner(index) == index // does this leg have a risk partner? Affects required collateral
@@ -1285,7 +1285,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         uint256 index,
         uint128 positionSize,
         int24 atTick,
-        uint32 poolUtilization
+        int16 poolUtilization
     ) internal view returns (uint256 required) {
         // extract the tokenType (token0 or token1)
         uint256 tokenType = tokenId.tokenType(index);
@@ -1296,15 +1296,10 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         // amount moved is right slot if tokenType=0, left slot otherwise
         uint128 amountMoved = tokenType == 0 ? amountsMoved.rightSlot() : amountsMoved.leftSlot();
 
-        // match tokenType with the correct pool utilization
-        int16 utilization = tokenType == 0
-            ? int16(uint16(poolUtilization))
-            : int16(uint16(poolUtilization >> 16));
-
         uint256 isLong = tokenId.isLong(index);
 
         // start with base requirement, which is based on isLong value
-        required = _getRequiredCollateralAtUtilization(amountMoved, isLong, utilization);
+        required = _getRequiredCollateralAtUtilization(amountMoved, isLong, poolUtilization);
 
         // if the position is long, required tokens does not depend on price
         unchecked {
@@ -1412,7 +1407,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         uint256 index,
         uint128 positionSize,
         int24 atTick,
-        uint32 poolUtilization
+        int16 poolUtilization
     ) internal view returns (uint256 required) {
         // extract partner index (associated with another liquidity chunk)
         uint256 partnerIndex = tokenId.riskPartner(index);
@@ -1483,7 +1478,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         uint128 positionSize,
         uint256 index,
         uint256 partnerIndex,
-        uint32 poolUtilization
+        int16 poolUtilization
     ) internal view returns (uint256 spreadRequirement) {
         // compute the total amount of funds moved for the position's current leg
         LeftRightUnsigned amountsMoved = PanopticMath.getAmountsMoved(tokenId, positionSize, index);
@@ -1552,9 +1547,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             _getRequiredCollateralAtUtilization(
                 tokenType == 0 ? movedRight : movedLeft,
                 1,
-                tokenType == 0
-                    ? int16(uint16(poolUtilization))
-                    : int16(uint16(poolUtilization >> 16))
+                poolUtilization
             )
         );
     }
@@ -1573,7 +1566,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         uint256 index,
         uint128 positionSize,
         int24 atTick,
-        uint32 poolUtilization
+        int16 poolUtilization
     ) internal view returns (uint256 strangleRequired) {
         // If both tokenTypes are the same, then this is a long or short strangle.
         // A strangle is an options strategy in which the investor holds a position
@@ -1599,14 +1592,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // A negative pool utilization is used to denote a position which is a strangle
             // at low pool utilization's strangle legs are evaluated at 2x capital efficiency
 
-            uint16 poolUtilization0 = uint16(poolUtilization);
-            uint16 poolUtilization1 = uint16(poolUtilization >> 16);
-
             // add 1 to handle poolUtilization = 0
-
-            poolUtilization =
-                uint32(uint16(-int16(poolUtilization0 == 0 ? 1 : poolUtilization0))) +
-                (uint32(uint16(-int16(poolUtilization1 == 0 ? 1 : poolUtilization1))) << 16);
+            poolUtilization = -(poolUtilization == 0 ? int16(1) : poolUtilization);
 
             return
                 strangleRequired = _getRequiredCollateralSingleLegNoPartner(
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -16,6 +16,7 @@ import {SafeTransferLib} from "@libraries/SafeTransferLib.sol";
 // Custom types
 import {LeftRightUnsigned, LeftRightSigned} from "@types/LeftRight.sol";
 import {LiquidityChunk} from "@types/LiquidityChunk.sol";
+import {PositionBalance} from "@types/PositionBalance.sol";
 import {TokenId} from "@types/TokenId.sol";
 
 /// @title Collateral Tracking System / Margin Accounting used in conjunction with a Panoptic Pool.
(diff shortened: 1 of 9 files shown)


guil-lambert:
All previous tests pass and I merged this PR into https://github.com/panoptic-labs/panoptic-v1-core-private/tree/feat/echidna-final and ran Medusa locally for 1M and got some reverts, but none related to that change. Should be mergeable.

contracts/CollateralTracker.sol
```diff
@@ -1183,10 +1184,12 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             TokenId tokenId = TokenId.wrap(positionBalanceArray[i][0]);
 
             // read the position size and the pool utilization at mint
-            uint128 positionSize = LeftRightUnsigned.wrap(positionBalanceArray[i][1]).rightSlot();
+            uint128 positionSize = PositionBalance.wrap(positionBalanceArray[i][1]).positionSize();
 
             // read the pool utilization at mint
-            uint128 poolUtilization = LeftRightUnsigned.wrap(positionBalanceArray[i][1]).leftSlot();
+            uint32 poolUtilization = PositionBalance
```
dyedm1: Do you think it might be cleaner to add a method in PositionBalance to extract `utilization0`/`utilization1` and use `s_underlyingIsToken0` to pick which method to use instead of retrieving both utilizations and extracting the correct one on each leg?

contracts/CollateralTracker.sol
```diff
@@ -1294,9 +1297,9 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         uint128 amountMoved = tokenType == 0 ? amountsMoved.rightSlot() : amountsMoved.leftSlot();
 
         // match tokenType with the correct pool utilization
-        int64 utilization = tokenType == 0
-            ? int64(uint64(poolUtilization))
-            : int64(uint64(poolUtilization >> 64));
+        int16 utilization = tokenType == 0
```
dyedm1: see comment above, might be nice to use `s_underlyingisToken0` to decide once in `_getTotalRequiredCollateral` instead of doing it here

contracts/CollateralTracker.sol
```diff
@@ -1550,8 +1553,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                 tokenType == 0 ? movedRight : movedLeft,
                 1,
                 tokenType == 0
-                    ? int64(uint64(poolUtilization))
-                    : int64(uint64(poolUtilization >> 64))
+                    ? int16(uint16(poolUtilization))
```
dyedm1: correct utilization would be passed into this function in above suggestion

contracts/CollateralTracker.sol
```diff
@@ -1596,14 +1599,14 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
-            uint64 poolUtilization1 = uint64(poolUtilization >> 64);
+            uint16 poolUtilization0 = uint16(poolUtilization);
+            uint16 poolUtilization1 = uint16(poolUtilization >> 16);
 
             // add 1 to handle poolUtilization = 0
 
             poolUtilization =
-                uint128(uint64(-int64(poolUtilization0 == 0 ? 1 : poolUtilization0))) +
-                (uint128(uint64(-int64(poolUtilization1 == 0 ? 1 : poolUtilization1))) << 64);
+                uint32(uint16(-int16(poolUtilization0 == 0 ? 1 : poolUtilization0))) +
```
dyedm1: see above

contracts/PanopticPool.sol
```diff
@@ -568,31 +569,53 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+        uint256 medianData;
+        uint96 tickData;
+        {
+            (
+                int24 currentTick,
+                int24 fastOracleTick,
+                int24 slowOracleTick,
+                int24 lastObservedTick,
+                uint256 _medianData
+            ) = PanopticMath.getOracleTicks(s_univ3pool, s_miniMedian);
```
dyedm1, line 593: This is actually a meaningful change in how we check solvency. We're now looking at the ticks  (particularly the `currentTick`) *at* mint rather than those after a potential swap at mint. I don't foresee any problems arising from this (I was against using the `currentTick` in collateral checks anyway), but your "invariant" about the accounts always being solvent at the `currentTick` could technically be broken now (though you could always just swap immediately after minting, so it doesn't make much of a difference IMO).

contracts/PanopticPool.sol
```diff
@@ -1381,6 +1432,47 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+    /// @param user The account to query
+    /// @param tokenId The tokenId for that account
+    /// @return currentTickAtMint currentTick at mint
+    /// @return fastOracleTickAtMint fast oracle tick at mint
+    /// @return slowOracleTickAtMint slow oracle tick at mint
+    /// @return lastObservedTickAtMint last observed tick at mint
+    /// @return utilization0AtMint utilization of token0 at mint
+    /// @return utilization1AtMint utilization of token1 at mint
+    /// @return positionSize size of the position
+    function positionData(
```
dyedm1, line 1445: If we have contract size issues, it may be prudent to just return the PositionBalance type instead of unpacking everything on our end

contracts/PanopticPool.sol
```diff
@@ -765,8 +785,39 @@ contract PanopticPool is ERC1155Holder, Multicall {
             uint256 _medianData
```
dyedm1: Any reason why this can't be set directly to the return var?

contracts/PanopticPool.sol
```diff
@@ -1381,6 +1432,47 @@ contract PanopticPool is ERC1155Holder, Multicall {
         _numberOfPositions = (s_positionsHash[user] >> 248);
     }
 
+    /// @notice Get the `tokenId` position data for `user`
```
dyedm1: period

contracts/types/PositionBalance.sol
```diff
@@ -0,0 +1,196 @@
...
+// Total                256bits : Total bits used by a PositionBalance.
+// ===============================================================================================
+//
+// The bit pattern is therefore:
+//
+//           (7)             (6)             (5)                (4)             (3)            (2)                 (1)
+//    <-- 24 bits -->  <-- 24 bits -->  <-- 24 bits -->   <-- 24 bits --> <-- 16 bits -->  <-- 16 bits -->   <-- 128 bits -->
+//   lastObservedTick   slowOracleTick   fastOracleTick     currentTick   utilization0      utilization1      positionSize
+//
+//        <--- most significant bit        least significant bit --->
```
dyedm1: nit: maybe make these symmetrical (see LiquidityChunk or TokenId)

contracts/types/PositionBalance.sol
```diff
@@ -0,0 +1,196 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.8.24;
+import "forge-std/Test.sol";
```
dyedm1: nit: rm this

contracts/types/PositionBalance.sol
```diff
@@ -0,0 +1,196 @@
...
+type PositionBalance is uint256;
+using PositionBalanceLibrary for PositionBalance global;
+
+/// @title A Panoptic Position Balance. Tracks the Position Size, the Pool Utilizations at mint, and the current/fastOracle/slowOracle/latestObserved ticks at mint.
+/// @author Axicon Labs Limited
+///
```
dyedm1: nit: rm this 3-slash padding if you're not going to put a notice tag underneath

contracts/types/PositionBalance.sol
```diff
@@ -0,0 +1,196 @@
...
+// (4) currentTick      24bits  : The currentTick at mint (int24).
+// (5) fastOracleTick   24bits  : The fastOracleTick at mint (int24).
+// (6) slowOracleTick   24bits  : The slowOracleTick at mint (int24).
+// (7) lastObservedTick 24bits  : The lastObservedTick at mint (int24).
+// Total                256bits : Total bits used by a PositionBalance.
+// ===============================================================================================
+//
+// The bit pattern is therefore:
+//
+//           (7)             (6)             (5)                (4)             (3)            (2)                 (1)
```
dyedm1: nit: can these be a little more aligned?

contracts/types/PositionBalance.sol
```diff
@@ -0,0 +1,196 @@
...
+        currentTick = self.currentTick();
+        fastOracleTick = self.fastOracleTick();
+        slowOracleTick = self.slowOracleTick();
+        lastObservedTick = self.lastObservedTick();
+    }
+
+    /// @notice Get token0 utilization of `self`.
+    /// @param self The PositionBalance to get utilization
+    /// @return The token0 utilization, stored in bips
+    function utilization0(PositionBalance self) internal pure returns (int256) {
```
dyedm1, line 172: I see you already added these, so they can be used in the CollateralTracker suggestion I made above

contracts/types/PositionBalance.sol
```diff
@@ -0,0 +1,196 @@
...
+//        <--- most significant bit        least significant bit --->
+//
+library PositionBalanceLibrary {
+    /*//////////////////////////////////////////////////////////////
+                                ENCODING
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Create a new `PositionBalance` given by positionSize, utilizations, and its tickData.
+    /// @param positionSize The amount of option minted
+    /// @param utilizations packing of two uint16 utilizations into a 32 bit word
```
dyedm1: nit: capitalize param descriptions

contracts/types/PositionBalance.sol
```diff
@@ -0,0 +1,196 @@
...
+            return
+                PositionBalance.wrap(
+                    (uint256(tickData) << 160) +
+                        (uint256(utilizations) << 128) +
+                        uint256(positionSize)
+                );
+        }
+    }
+
+    function packTickData(
```
dyedm1, line 63: nit: natspec

contracts/types/PositionBalance.sol
```diff
@@ -0,0 +1,196 @@
...
+        int24 currentTick,
+        int24 fastOracleTick,
+        int24 slowOracleTick,
+        int24 lastObservedTick
+    ) internal pure returns (uint96) {
+        return
```
dyedm1: unchecked?

contracts/types/PositionBalance.sol
```diff
@@ -0,0 +1,196 @@
...
+    /*//////////////////////////////////////////////////////////////
+                                DECODING
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Get the last observed tick of `self`.
+    /// @param self The PositionBalance to get the requested tick
+    /// @return The last observed tick of self
+    function lastObservedTick(PositionBalance self) internal pure returns (int24) {
+        unchecked {
+            return int24(int256((PositionBalance.unwrap(self) >> 232) % 2 ** 24));
```
dyedm1: Isn't the % 2**24 here (and in other functions) redundant with the int24 cast?

contracts/types/PositionBalance.sol
```diff
@@ -0,0 +1,196 @@
...
+        internal
+        pure
+        returns (
+            int24 currentTick,
+            int24 fastOracleTick,
+            int24 slowOracleTick,
+            int24 lastObservedTick
+        )
+    {
+        PositionBalance self = storeBalanceData(0, 0, tickData);
```
dyedm1: I think the storeBalanceData here is redundant, you can just wrap a shifted tickData and save some gas

dyedm1 (commented):
Looks good overall, made a few suggestions

contracts/PanopticPool.sol
dyedm1: nope it's still post-mint

PR #124, opened 2024-08-27
fix: correct reversed force exercise compensation

dyedm1:
My `ForceExercise: Sanity check - Exercisee token value is lower than before` invariant broke, and it  turns out we’ve been calculating loss compensation for force exercisees wrong the entire time.

The idea behind loss compensation is that *for long positions*, manipulating the current price can lead to less favorable (benchmarked against the organic market price or oracle tick) outcomes than burning the position under normal circumstances. Therefore, the decision was made to compensate force exercisees for any possible such changes in their long legs caused by price manipulation.

To implement this change and revert the exercisee’s token balances back to the value they would have received burning at the oracle tick, we’ve always done something like:
`exerciseFee -= (oracle/medianvalue - currentValue)` for each token. For example, if the liquidity underlying the position is worth 5000 token0 at the current price (out-of-range) and 1000 token1 at the oracle price (also out-of-range), the exercisee would be compensated with 1000 token1 in exchange for refunding 5000 token0 to the exercisor (a negative exercise fee represents a payment from exercisee to exercisor), in theory returning their account to the state it would have been had they burned at the oracle price.

However, the key error here is that this calculation is for *long positions*. When that exercisee closes their position, they add liquidity to Uniswap and actually *pay* 1000 USDC, so in order to return their account to the oracle-tick state you would actually have to do the exact opposite, compensating them with 1000 token1 in exchange for a refund of 5000 token0 (i.e. we need to swap currentValue and oracleValue in that expression)

My guess as to why auditors didn’t catch this is that it’s more of an economic issue and they may have believed that it was intended behavior. Still, we need to fix this because currently:
- The logic here *does not prevent* the intended attack vector of atomic price manipulation across far-out-of-range exercisees
- It actually makes the original attack vector *more profitable* because it about doubles the size of the forced trade at the (unfavorable) far-out-of-range strike price

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -725,7 +725,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                 // reverse any token deltas between the current and oracle prices for the chunk the exercisee had to mint in Uniswap
                 // the outcome of current price crossing a long chunk will always be less favorable than the status quo, i.e.,
                 // if the current price is moved downward such that some part of the chunk is between the current and market prices,
-                // the chunk will composition will swap token1 for token0 at a price (token0/token1) more favorable than market (token1/token0),
+                // the chunk composition will swap token1 for token0 at a price (token0/token1) more favorable than market (token1/token0),
                 // forcing the exercisee to provide more value in token0 than they would have provided in token1 at market, and vice versa.
                 // (the excess value provided by the exercisee could then be captured in a return swap across their newly added liquidity)
                 exerciseFees = exerciseFees.sub(
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -722,13 +722,17 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                     );
                 }
 
-                // compensate user for loss in value if chunk has lost money between current and median tick
-                // NOTE: the delta for one token will be positive and the other will be negative. This cancels out any moves in their positions
+                // reverse any token deltas between the current and oracle prices for the chunk the exercisee had to mint in Uniswap
+                // the outcome of current price crossing a long chunk will always be less favorable than the status quo, i.e.,
+                // if the current price is moved downward such that some part of the chunk is between the current and market prices,
+                // the chunk will composition will swap token1 for token0 at a price (token0/token1) more favorable than market (token1/token0),
+                // forcing the exercisee to provide more value in token0 than they would have provided in token1 at market, and vice versa.
+                // (the excess value provided by the exercisee could then be captured in a return swap across their newly added liquidity)
                 exerciseFees = exerciseFees.sub(
                     LeftRightSigned
                         .wrap(0)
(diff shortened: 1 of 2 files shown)


guil-lambert (approved):

guil-lambert:
LGTM

contracts/CollateralTracker.sol
```diff
@@ -722,13 +722,17 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                     );
                 }
 
-                // compensate user for loss in value if chunk has lost money between current and median tick
-                // NOTE: the delta for one token will be positive and the other will be negative. This cancels out any moves in their positions
+                // reverse any token deltas between the current and oracle prices for the chunk the exercisee had to mint in Uniswap
+                // the outcome of current price crossing a long chunk will always be less favorable than the status quo, i.e.,
+                // if the current price is moved downward such that some part of the chunk is between the current and market prices,
+                // the chunk will composition will swap token1 for token0 at a price (token0/token1) more favorable than market (token1/token0),
```
flanagansteve: Typo - should likely be "the chunk composition will swap"

flanagansteve (approved):
I think this makes sense, based on your PR description. The code changes certainly implement the changes you describe in the PR. Only noticed one thing:

contracts/CollateralTracker.sol
```diff
@@ -722,13 +722,17 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                     );
                 }
```
flanagansteve, line 723: Oh also - is the docstring for this method fully accurate now?

```
/// @notice Get the cost of exercising an option. Used during a forced exercise.
    /// @dev This one computes the cost of calling the forceExercise function on a position:
    /// - The forceExercisor will have to *pay* the exercisee because their position will be closed "against their will"
    /// - The cost must be larger when the position is close to being in-range, and should be minimal when it is far from being in range. eg. Exercising a (1000, 1050)
    ///   position will cost more if the price is 999 than if it is 100
    /// - The cost is an exponentially decaying function of the distance between the position's strike and the current price
    /// - The cost decreases by a factor of 2 for every "position's width"
    /// - Note that the cost is the largest among all active legs, not the sum
    /// @dev Example exercise costs:
    /// - 10% if the position is liquidated when the price is between 950 and 1000, or if it is between 1050 and 1100
    /// - 5% if the price is between 900 and 950 or (1100, 1150)
    /// - 2.5% if between (850, 900) or (1150, 1200)
    /// @param currentTick The current price tick
    /// @param oracleTick The price oracle tick
    /// @param positionId The position to be exercised
    /// @param positionBalance The balance in `account` of the position to be exercised
    /// @param longAmounts The amount of longs in the position
/// @return exerciseFees The fees for exercising the option position
```

I _think_ it still is, since the docstring doesn't seem to make specific claims about what happen when currentTick and oracleTick diverge - it talks more about whether the "price" is in-range, leaving it ambiguous which tick it refers to by "price". But wanted to double check.

PR #125, opened 2024-08-27
feat: improve cross-collateral precision

dyedm1:
Our previous method of combining token0 and token1 balances for cross-collateralization calculations involved multiplying/dividing token0/token1 values by the square root of the current price, adding them together, and comparing the results without converting either token fully into token1 or token0. While this approach works out mathematically (and practically, for most position sizes/prices), it can lead to precision loss for small positions (and larger positions at more extreme prices), potentially causing accounts to be marked at insolvent when solvent or vice versa, depending on which rounding direction is used. 

The solution, implemented here, is to *fully convert* collateral balances into the *highest precision token* available. Practically speaking, this means converting collateral balances into token0 when the tick is negative (token1/token0 price < 1) and into token1 when the tick is positive (token1/token0 price > 1). While precision can still be lost with this approach (e.g. price is 1.5 and a user has 1 wei of token0), it ensures that a user who is solvent by more than one worth of the lowest-precision token cannot be marked insolvent, and vice versa.


contracts/libraries/Math.sol
```diff
@@ -736,6 +736,20 @@ library Math {
...
     }
 
+    /// @notice Calculates ceil(a×b÷2^192) with full precision.
+    /// @param a The multiplicand
+    /// @param b The multiplier
+    /// @return result The 256-bit result
+    function mulDiv192RoundingUp(uint256 a, uint256 b) internal pure returns (uint256 result) {
+        unchecked {
+            result = mulDiv192(a, b);
+            if (mulmod(a, b, 2 ** 192) > 0) {
```
flanagansteve, line 746: Small nit: We do this `if(mulmod(a, b, denominator) > 0)` in seemingly all the `RoundingUp` functions (this addition makes 4 instances). It might make sense to kick these 4 lines to a helper by now

contracts/libraries/PanopticMath.sol
```diff
@@ -564,6 +521,26 @@ library PanopticMath {
         }
     }
 
+    /// @notice Convert an amount of token0 into an amount of token1 given the sqrtPriceX96 in a Uniswap pool defined as sqrt(1/0)*2^96.
+    /// @dev Uses reduced precision after tick 443636 in order to accommodate the full range of ticks
+    /// @param amount The amount of token0 to convert into token1
+    /// @param sqrtPriceX96 The square root of the price at which to convert `amount` of token0 into token1
+    /// @return The converted `amount` of token0 represented in terms of token1
+    function convert0to1RoundingUp(
```
flanagansteve, line 529: Was there a reason to make dedicated `convert0to1RoundinUp`/`convert1to0RoundingUp` methods as opposed to adding a `bool` argument for whether we should round up? Maybe its more gas efficient this way as you're passing fewer args around?

flanagansteve (approved):
I _think_ your PR description has the token1/token0 < 1 or > 1 flipped around here:

"Practically speaking, this means converting collateral balances into token0 when the tick is negative (token1/token0 price < 1) and into token1 when the tick is positive (token1/token0 price > 1)"

Negative tick would mean token1 / token0 is > 1, and v.v.; your comment here also indicates you wish to use the lowest priced token:

```
// `bonusCross` and `thresholdCross` are returned in terms of the lowest-priced token
```

Apart from that, I think I'm understanding things, and it overall looks good - let some small comments:

contracts/libraries/PanopticMath.sol
dyedm1: yes adding a branch there is more expensive than just calling a different function. Flags really only make sense for larger functions that are costly (contract size or aesthetics-wise) to replicate.

contracts/libraries/Math.sol
dyedm1: solc doesn't inline functions on the legacy pipeline, so moving very small bits of logic into helpers wastes gas on overhead

dyedm1:
> Negative tick would mean token1 / token0 is > 1

Nope, the formula is `1.0001 ^ T = token1/token0`.

guil-lambert:
Added an unchecked to reduce bytecode size, LGTM otherwise

PR #127, opened 2024-09-04
Fix: revoke during liquidations with protocol loss

guil-lambert:
I finally got to the bottom of the shares megainflation issue.

---
When calling `s_collateralToken.revoke` in `liquidate`, the goal is to return to the liquidator the assets they delegated plus any bonus amount they're entitled to receive. 

However, if the protocol loss wipes most of the assets in the pool, the amount of asset requested could be larger than the actual assets in the PanopticPool.

We currently describe the logic for the new share issuance in revoke : 

```
// this is paying out protocol loss, so correct for that in the amount of shares to be minted
// X: total assets in vault
// Y: total supply of shares
// Z: desired value (assets) of shares to be minted
// N: total shares corresponding to Z
// T: transferred shares from liquidatee which are a component of N but do not contribute toward protocol loss
// Z = N * X / (Y + N - T)
// Z * (Y + N - T) = N * X
// ZY + ZN - ZT = NX
// ZY - ZT = N(X - Z)
// N = (ZY - ZT) / (X - Z)
// N = Z(Y - T) / (X - Z)   
// subtract delegatee balance from N since it was already transferred to the delegator
```

In the case of protocol loss, `totalAsset()` = the amount of existing assets in the pool may be smaller than `assets` = the value requested. Which means `X-Z` is negative

We kind of take care of this when we create those shares here:

```
_mint(
    delegator,
    Math.mulDiv(
        assets,
        totalSupply - delegateeBalance,
        uint256(Math.max(1, int256(totalAssets()) - int256(assets)))
    ) - delegateeBalance
);
```

where the `uint256(Math.max(1, int256(totalAssets()) - int256(assets)))` expression returns `1` when it gets negative. But this means the amount of shares minted gets very large ~ `assets * totalSupply` and the final share price will become inaccurate. 

---
The compute the number of shares to be issued, we could replicate what we're doing for the commission: base the calculation on the pre-mint values and under-estimate the number of shares. In this case:

```
N = Z*Y/X
```

so we would simply call [here](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/c99fb2e0032ff1b4a43a8e1976e8e76b1c927826/contracts/CollateralTracker.sol#L970): 

```
_mint(delegator, Math.mulDiv(assets, totalSupply, totalAssets())-T)
```


dyedm1:
> under-estimate the number of shares

This statement is doing a lot of legwork. `_mint(delegator, Math.mulDiv(assets, totalSupply, totalAssets())-T)` was actually the original formula (though it was a little more efficient and looked something like this, no need to do the conversion twice `_mint(delegator, shares - delegateeBalance)`). However, I had to develop the current formula because our tests failed on the original one; the value of the shares transferred to the liquidator are subject to the share price decrease caused by the `totalSupply` increase, which can result in a reduced bonus or even highly unprofitable liquidations.

Effectively, this change would:
- slightly reduce the bonus liquidators receive at low levels of protocol loss
- make liquidations unprofitable/economically non-viable at high levels of protocol loss

Let's look at an example with 50% protocol loss:
- Bob collateralizes an OTM 2000 USDC put with 400 USDC. The total amount of assets in the USDC pool is 3600 USDC
- The price moves down by >99% and causes the put to become ITM such that it is worth ~0 USDC at the current tick
- Bob is liquidated by Alice. There is a shortfall of 1600 USDC, so Alice delegates 1600 USDC to Bob
- Post-liquidation, Bob has 0 USDC (1600 + 400 = 2000) and owes the liquidator 1800 USDC (1600 delegated + 200 bonus)
- The protocol loss amount is 1800-0=1800 USDC, 50% of the original TVL of the pool
- PREVIOUS LOGIC: During `revoke`, 1800 USDC worth of shares (valued at post-liquidation) are minted to the liquidator. Practically speaking, this means `1 * totalSupply` shares are minted to the liquidator, such that they own 50% (=1800) of the 3600 assets in the pool. (1800 -1600 = 200 USDC profit)
- NEW LOGIC: During `revoke`, 1800 USDC worth of shares (valued at PRE-liquidation) are minted to the liquidator. This means that `0.5 * totalSupply` shares are minted to the liquidator, resulting in the final share supply being `1.5 * totalSupply` and the liquidator owning 1/3 (=1200) of the 3600 assets in the pool. (1200 - 1600 = 400 USDC loss for the liquidator)

Additionally, despite this new logic, ~100% protocol loss can still increase the share price by ~2x, meaning that the share supply can still become high with a (relatively) reasonable number of liquidations.

PR #128, opened 2024-09-04
feat: fixed-size virtual delegation on liquidation/force exercise

dyedm1:


guil-lambert:
I know this is still WIP, but can you split the PR into one which adds `netPaid` to `forceExercise()` and one which adds the virtual delegation to liquidation? The former will be easier to review+merge than the latter.

dyedm1:
This should be easier to review. Main changes are just:
- standard `2^248 - 1` shares virtual delegation for liquidations and force exercising
- Combining revoke/refund logic for liquidations into a single function: `settleLiquidation`
- Modifying balance shortage calculation in `getExerciseDeltas` a little to use shares instead of assets in order to avoid converting the balance directly and make the rounding a little more robust (in favor of the exercisee).

contracts/PanopticPool.sol
```diff
@@ -1059,21 +1050,9 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-            uint256(int256(uint256(_delegations.rightSlot())) + liquidationBonus0)
-        );
-        s_collateralToken1.revoke(
-            msg.sender,
-            liquidatee,
-            uint256(int256(uint256(_delegations.leftSlot())) + liquidationBonus1)
-        );
-
-        // ensure the liquidator is still solvent after the liquidation
-        _validateSolvency(msg.sender, positionIdListLiquidator, BP_DECREASE_BUFFER);
```
guil-lambert, line 1076: Why remove this check for the liquidator's solvency?

contracts/PanopticPool.sol
```diff
@@ -1111,6 +1088,8 @@ contract PanopticPool is ERC1155Holder, Multicall {
         {
             (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
 
+            uint128 positionBalance = s_positionBalance[account][touchedId[0]].positionSize();
```
guil-lambert: `uint128 positionBalance` -> `uint128 positionSize`

contracts/CollateralTracker.sol
```diff
@@ -872,117 +872,107 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+
+            SafeTransferLib.safeTransferFrom(s_underlyingToken, liquidator, msg.sender, bonusAbs);
+
+            _mint(liquidatee, convertToShares(bonusAbs));
+
+            s_poolAssets += uint128(bonusAbs);
+
+            uint256 liquidateeBalance = balanceOf[liquidatee];
+
+            if (type(uint248).max > liquidateeBalance) {
```
guil-lambert, line 915: Small nit: why not `liquidateeBalance < type(uint248).max`

It's the same either way, but the `>` is less common in our codebase

contracts/CollateralTracker.sol
```diff
@@ -872,117 +872,107 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+                unchecked {
+                    totalSupply += type(uint248).max - liquidateeBalance;
+                }
+                balanceOf[liquidatee] = liquidateeBalance = 0;
+            } else {
+                unchecked {
+                    balanceOf[liquidatee] = liquidateeBalance -= type(uint248).max;
```
guil-lambert: Took me a while to realize `-=` was intended. 

Please split it over two lines for clarity:

```
liquidateeBalance -= type(uint248).max;
balanceOf[liquidatee] = liquidateeBalance;
```

Same with 

```
liquidateeBalance = 0;
balanceOf[liquidatee] = liquidateeBalance'
```

contracts/libraries/PanopticMath.sol
```diff
@@ -1020,72 +1020,84 @@ library PanopticMath {
...
-    /// @param collateral0 CollateralTracker for token0
-    /// @param collateral1 CollateralTracker for token1
+    /// @param ct0 The collateral tracker for token0
+    /// @param ct1 The collateral tracker for token1
     /// @return The LeftRight-packed deltas for token0/token1 to move from the exercisor to the exercisee
     function getExerciseDeltas(
-        address exercisee,
-        uint256 delegated0,
-        uint256 delegated1,
+        address account,
```
guil-lambert: Small nit: `account` -> `exercisee`? 

`forceExercise(account, ...)` could also be renames to `forceExercise(exercisee, ...)` in PanopticPool too

contracts/CollateralTracker.sol
```diff
@@ -872,117 +872,107 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
-        */
-
-        uint256 shares = convertToShares(assets);
-
-        // transfer shares from the delegator to the delegatee
-        _transferFrom(delegator, delegatee, shares);
-    }
-
-    /// @notice Increase the share balance of a user by shares worth `assets` without updating the total supply.
+    /// @notice Increase the share balance of a user by 2^248 - 1 without updating the total supply.
```
guil-lambert, line 875: Curious:  any reason for the `2^248-1` vs any other exponent?

guil-lambert (changes requested):

contracts/PanopticPool.sol
dyedm1: That check was there because a liquidator's collateral balance in one token could decrease. However, we now settle any payments *from* the liquidator directly in the  underlying token, so their collateral balance can only increase (technically, the value of their collateral can decrease due to protocol loss, but this would happen regardless of whether it was they who executed the liquidation)

contracts/CollateralTracker.sol
dyedm1: It's not a "magic number" in any sense, but it is ~256x smaller than the maximum, which strikes a good balance between being large enough to cover any settlement of token amounts on a non-hypershareinflated pool while still being small enough to ensure it won't exceed a user's balance unless the share supply is 1-2 liquidations away from approaching the maximum.

contracts/PanopticPool.sol
guil-lambert: > technically, the value of their collateral can decrease due to protocol loss, but this would happen regardless of whether it was they who executed the liquidation

They could withdraw funds and incur more protocol loss because they're the ones that control the timing of that liquidation. It goes against the FREI-PI pattern by removing it too. So since it doesn't hurt to keep that solvency check there, I'd say you should add it back.

dyedm1: > It goes against the FREI-PI pattern by removing it too

If this goes against the FREI-PI pattern, the fact that we don't check solvency after collateral deposits does also.

> They could withdraw funds and incur more protocol loss because they're the ones that control the timing of that liquidation

Are you saying they could withdraw funds *before* the liquidation and assign more of the protocol loss to the remaining depositors? There may certainly be an incentive to do that (even with accounts who are not necessarily the liquidator), but I don't see the relation between that and this solvency check.

> So since it doesn't hurt to keep that solvency check there

You can make this argument for a lot of things, and sometimes it's subjective (e.g. the "protocol hardening" changes you added), but in this case it is *objectively impossible* for a liquidation to decrease the liquidator's collateral balance (unless a user self-liquidates, in which case the solvency check would always pass given all their positions would be closed).

> technically, the value of their collateral can decrease due to protocol loss, but this would happen regardless of whether it was they who executed the liquidation

An account will always be *more* solvent if *they* perform a liquidation than they would be if any other user performed the same liquidation. The invariant we were (indirectly) enforcing here is that insolvent accounts are not permitted to decrease their collateral balances under any circumstances besides a liquidation of their own account (which was something that could happen when we were settling payments from the liquidator in collateral shares), but now that their collateral balance *cannot decrease* under any circumstances during a liquidation, the solvency check is pointless. It's actually fine if an insolvent account performs a liquidation -- as long as their collateral balance does not decrease.

guil-lambert: > If this goes against the FREI-PI pattern, the fact that we don't check solvency after collateral deposits does also.

Maybe we could add it there too. 

> I don't see the relation between that and this solvency check.

If they do this, then the liquidation will revert because the shares dilution made them insolvent. 

That may be more relevant for accounts which already have positions open + funds before the crash, and in that case they can be solvent after a withdrawal but insolvent after the liquidation.

>but in this case it is objectively impossible for a liquidation to decrease the liquidator's collateral balance (unless a user self-liquidates, in which case the solvency check would always pass given all their positions would be closed)

I'm not so sure it's impossible though (who knows?). We could add a condition in the echidna test to check to really make sure.

> An account will always be more solvent if they perform a liquidation than they would be if any other user performed the same liquidation.

Yes, but three could be so much protocol loss that the bonus makes liquidations less profitable (even at a loss in extreme cases).

contracts/CollateralTracker.sol
```diff
@@ -872,117 +872,107 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
     /// @dev This is controlled by the Panoptic Pool - not individual users.
     /// @param delegatee The account to increase the balance of
-    /// @param assets The amount of assets worth of shares to mint to `delegatee`
-    /// @return shares The amount of shares minted to `delegatee`
```
flanagansteve, line 909: What is the motivation to hardcode type(uint248).max as the amount to increase the delegatee's balance by, rather than making it an argument which PanopticPool passes in `type(uint248).max` to?

I believe its gas saving - and if so, would it make sense to rename `delegate` to something like `delegateUint248Max` or `delegate2ToThe248th`, to show that its not just a generic `delegate` method and rather delegates a specific amount?

flanagansteve: Alternatively, maybe noting the specific amount in the docstring is enough, but personally I'd expect it in the function name

contracts/CollateralTracker.sol
```diff
@@ -872,117 +872,107 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
     }
 
-    /// @notice Decrease the share balance of a user without updating the total supply.
-    /// @dev Assumes that `delegatee` has enough money to pay for the refund, will revert otherwise.
+    /// @notice Decrease the share balance of a user by 2^248 - 1 without updating the total supply.
+    /// @dev Assumes that `delegatee` has >=(2^248 - 1) tokens, will revert otherwise.
     /// @dev This is controlled by the Panoptic Pool - not individual users.
     /// @param delegatee The account to decrease the balance of
-    /// @param shares The amount of shares to burn from `delegatee`
-    function revoke(address delegatee, uint256 shares) external onlyPanopticPool {
```
flanagansteve, line 923: Same comment RE revoking a hardcoded amount and possibly noting that in the function name (though you do so in the docstring here which is nice) - e.g., could be revokeUint248Max

contracts/PanopticPool.sol
```diff
@@ -920,24 +920,22 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
     /// @param liquidatee Address of the distressed account
-    /// @param delegations LeftRight amounts of token0 and token1 (token0:token1 right:left) delegated to the liquidatee by the liquidator so the option can be smoothly exercised
     /// @param positionIdList List of positions owned by the user. Written as [tokenId1, tokenId2, ...]
     function liquidate(
-        TokenId[] calldata positionIdListLiquidator,
         address liquidatee,
-        LeftRightUnsigned delegations,
-        TokenId[] calldata positionIdList
+        TokenId[] calldata positionIdList,
+        TokenId[] calldata positionIdListLiquidator
```
flanagansteve: Need to bring back the docstrings for these args now that we have restored them

contracts/PanopticPool.sol
```diff
@@ -1024,7 +1015,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
                     tokenData0,
                     tokenData1,
                     Math.getSqrtRatioAtTick(twapTick),
-                    netExchanged,
+                    netPaid,
                     shortPremium
                 );
```
flanagansteve, line 1024: In the comment below, you still reference the `netExchanged` variable which is now netPaid - need to update:

"// reusing variables to save stack space; netExchanged = deltaBonus0, premia = deltaBonus1"

to

"// reusing variables to save stack space; netPaid = deltaBonus0, premia = deltaBonus1"

contracts/CollateralTracker.sol
```diff
@@ -872,117 +872,107 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+            uint256 liquidateeBalance = balanceOf[liquidatee];
+
+            if (type(uint248).max > liquidateeBalance) {
+                balanceOf[liquidatee] = 0;
+                unchecked {
+                    totalSupply += type(uint248).max - liquidateeBalance;
+                }
+            } else {
+                unchecked {
+                    balanceOf[liquidatee] = liquidateeBalance - type(uint248).max;
```
flanagansteve, line 922: Why do the revocation here instead of after the liquidation  bonus settling like we did before? If we move this to a `revoke` call after `settleLiquidationBonus` it would match the force exercise pattern more closely (call .delegate -> do accounting -> call .revoke)

contracts/CollateralTracker.sol
```diff
@@ -872,117 +872,107 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+                // Y: total supply of shares
+                // Z: desired value (assets) of shares to be minted
+                // N: total shares corresponding to Z
+                // T: transferred shares from liquidatee which are a component of N but do not contribute toward protocol loss
+                // Z = N * X / (Y + N - T)
+                // Z * (Y + N - T) = N * X
+                // ZY + ZN - ZT = NX
+                // ZY - ZT = N(X - Z)
+                // N = (ZY - ZT) / (X - Z)
+                // N = Z(Y - T) / (X - Z)
```
flanagansteve, line 958: Double checking the math:

- We will never divide by 0 using this formula, because there isn't a situation where X (total assets in vault) and Z (assets that are being protocol loss'ed) are equal, because a liquidatee will always have non-zero assets so `X-Z` will always be at least 1
- Similarly, `X-Z` can never be negative which would cause this to try and mint a negative amount
- `X-Z` is in whole numbers, so `X-Z` would never be between 0 and 1 and blow the output `N` up & threaten overflow or something like that
- `X-Z` can never exceed `Z(Y-T)` and result in an attempt to mint 0 shares - this one I'm least sure on - what if:
  - X is large
  - Z is small (but liquidatee has even fewer assets in the CT)
  - Y is small (which might not make sense if X is large, but share price could be small)
  - and T is small (which might make sense if Z is small-but-larger-than-liquidatees-assets-in-CT)
  - You'd then possibly get a small `Z(Y-T)` if Y is sufficiently small, and a large `X-Z`
  - Is that realistic? Or would a scenario where Z is small (e.g. the protocol loss is small) and liquidatee's assets in the CT is even smaller be unrealistic/impossible?
  - Or, perhaps Y-T always exceeds X-Z and Z is always >= 1 since its a whole number, therefore we are OK? But i don't think i have a good explanation for Y-T would always exceed X-Z yet

contracts/CollateralTracker.sol
```diff
@@ -872,117 +872,107 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+                }
+                balanceOf[liquidatee] = liquidateeBalance = 0;
+            } else {
+                unchecked {
+                    balanceOf[liquidatee] = liquidateeBalance -= type(uint248).max;
+                }
+            }
+
+            // transfer the bonus to the liquidator
+            uint256 shares = convertToShares(uint256(bonus));
```
flanagansteve: Would you consider renaming this to `bonusShares`? or are you aiming to be concise in subsequent references?

contracts/CollateralTracker.sol
```diff
@@ -872,117 +872,107 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+
+            // if requested amount is larger than user balance, transfer shares back,
```
flanagansteve: You call `liquidatee` `delegatee` in comments in this block - can we instead refer to them as `liquidatee` for consistency?

contracts/libraries/PanopticMath.sol
```diff
@@ -808,8 +808,8 @@ library PanopticMath {
             int256 balance1 = int256(uint256(tokenData1.rightSlot())) -
```
flanagansteve, line 809: You also still reference netExchanged in a comment on line 805, which should likely be netPaid now

contracts/CollateralTracker.sol
```diff
@@ -872,117 +872,107 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+                            uint256(bonus),
+                            totalSupply - liquidateeBalance,
+                            uint256(Math.max(1, int256(totalAssets()) - bonus))
+                        ) - liquidateeBalance
+                    );
+                }
+            }
+            // if requested amount < delegatee balance, then just transfer shares back
+            else {
+                _transferFrom(liquidatee, liquidator, shares);
```
flanagansteve: `shares` might be `0` - if you trace through from the start of this method you end up here. Do we allow 0 transfers? I believe we do, but just double checking

contracts/CollateralTracker.sol
flanagansteve: (I now realise your Math.max(1, (X-Z)) in the mulDiv arguments takes care of bullets 1 through 3 - though that may not actually be necessary if i'm correct in my double checking above)

contracts/CollateralTracker.sol
flanagansteve: Confirmed we do: https://github.com/panoptic-labs/panoptic-v1-core-private/blob/main/contracts/tokens/ERC20Minimal.sol#L103

contracts/CollateralTracker.sol
dyedm1: That would cost more gas on the = case (the balance would be reset to 0 and totalSupply would be increased by 0, instead of the balance just being set to `type(uint248).max - type(uint248).max = 0`)

contracts/libraries/PanopticMath.sol
dyedm1: The latter is somewhat redundant but implemented the former

contracts/CollateralTracker.sol
dyedm1: I don't think it should reference the specific number -- the point of that function is to temporarily give an account an "unlimited" balance (of course it's finite/not actually unlimited but the assumption should be that it is) to draw from that can be settled later. If you think of any non-awkward name that better represents that purpose let me know

contracts/PanopticPool.sol
dyedm1: Oh Guillaume must have added that back -- the comment itself is actually just out of date/wrong entirely

contracts/CollateralTracker.sol
dyedm1: Because the ordering of the revocation/bonus is significant and needs to be arranged differently depending on the bonus.

For negative bonuses, the underlying token transfer + subsequent liquidatee balance increase needs to happen *before* the revocation to ensure that `totalSupply` does not increase more than necessary.

Likewise, for positive bonuses, the revocation needs to happen *before* the bonus payment to ensure the liquidator's bonus is calculated with the correct share price (given that some protocol loss can be realized through a `totalSupply` increase during the revocation)

contracts/CollateralTracker.sol
dyedm1: > We will never divide by 0 using this formula, because there isn't a situation where X (total assets in vault) and Z (assets that are being protocol loss'ed) are equal, because a liquidatee will always have non-zero assets so X-Z will always be at least 1

> Similarly, X-Z can never be negative which would cause this to try and mint a negative amount

No, the bonus can contribute to the protocol loss as well. 

> X-Z is in whole numbers, so X-Z would never be between 0 and 1 and blow the output N up & threaten overflow or something like that

No, though It may overflow (and actually may revert a little earlier than the threshold for number of liquidations from #129 so maybe I'll do another PR to fallback cleanly there without reverting)

dyedm1: > Or, perhaps Y-T always exceeds X-Z and Z is always >= 1 since its a whole number, therefore we are OK? But i don't think i have a good explanation for Y-T would always exceed X-Z yet

It may not always exceed it, but it is guaranteed to be greater than or equal to `liquidateeBalance` (so zero shares can be minted, but if that's the case it's because less than 1 share worth of assets was missing, which could happen on a pool that saw extreme increases in share price).

Proof:
- `liquidateeBalance` is guaranteed to be less than `shares` in this branch
- `shares` is equivalent to `bonus * totalSupply / totalAssets()`
- `N = bonus * (totalSupply - liquidateeBalance) / (totalAssets() - bonus)`
- `bonus / totalAssets()` is guaranteed to be more than `liquidateeBalance / totalSupply` given the first point
- Thus, the ratio `(totalSupply - liquidateeBalance) / (totalAssets() - bonus)` must be greater than `totalSupply / totalAssets()`
- Therefore, `bonus * (totalSupply - liquidateeBalance) / (totalAssets() - bonus) >= liquidateeBalance`

flanagansteve: > so zero shares can be minted,

Ah - and minting 0 protocol-loss-derived-shares to the liquidator actually isn't a problem, just wastes a little gas.

PR #129, opened 2024-09-04
fix: cap protocol loss per liquidation at `100% - 1bp`

dyedm1:
Fixes the problem #127 is attempting to solve while issuing the correct value of shares to liquidators at reasonable levels of protocol loss (<99.99%). 

We can adjust it, but a threshold of `totalSupply * 10_000` requires 10+ liquidations at 99.99%+ protocol loss to mint a problematically large quantity of shares (on an 18-decimal vault with 100 million tokens).

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -977,7 +977,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                         _totalSupply - delegateeBalance,
                         uint256(Math.max(1, int256(totalAssets()) - int256(assets)))
                     ) - delegateeBalance,
-                    totalSupply * DECIMALS
+                    _totalSupply * DECIMALS
                 )
             );
         }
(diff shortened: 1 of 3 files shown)


guil-lambert:
Could we do a direct transfer of the assets instead of inflating shares during revoke? Basically diluting the value of the remaining participants by decreasing `totalAssets()` instead of inflating `totalSupply()`? 

There will of course be situations where `assets > totalAssets()`, so in that case we could cap the transfer of assets to 100% - 1bps without inflating the shares at all.

```
if (shares > delegateeBalance) {
    // transfer delegatee balance to delegator
    _transferFrom(delegatee, delegator, delegateeBalance);

    // transfer remaining assets, cap if necessary
    uint256 remainingAssets = Math.min(assets - convertToAssets(delegateeBalance), (totalAssets() * DECIMALS) / DECIMALS);

    SafeTransferLib.safeTransferFrom(s_underlyingToken, msg.sender, delegator, remainingAssets); 
    // Q: Does this safeTransferFrom need Reentrancy lock?
 
} else {
    _transferFrom(delegatee, delegator, shares);
}
```

guil-lambert (approved):
LGTM!

PR #141, opened 2024-10-03
Code freeze: v1.0.0

dyedm1:
Past this commit, no logic changes can be deployed until they are re-audited and released as a new version. 

A few very minor changes in response to our final C4 audited are included in the diff here. The size of the diff is six (significant) lines of code. The only change that will have effects post-deployment(`maxWithdraw`/`maxRedeem`) is not within the scope of changes for our Uniswap V4 port (v1.1), so that will be reviewed in our upcoming Cantina audit.

Changelog:
- Return 0 instead of reverting when `s_poolAssets == 0` to comply with EIP-4626. I wanted to see if I could get away with not handling this in #134 because as far as I was aware it was impossible for `s_poolAssets` to reach zero. It is true that withdrawing the 1 virtual asset is impossible, however, one of our auditors made the valid point that it *can* be moved to `s_inAMM` temporarily if someone donates one asset to the pool and creates a position collateralized in the other token borrowing the entire stock of `s_poolAssets`. The only impacts this has are:
	- `maxRedeem` and `maxWithdraw` revert instead of returning zero in a few edge cases when no assets are available to withdraw, which could potentially confuse integrating contracts or frontends (though this could be easily resolved by depositing 1 token or creating a long position)
	- The *standard* ERC-4626 functions `redeem` and `withdraw` will revert with a different error than usual when this is the case. EIP-4626 does not define standard errors, so this is not a problem.
- Deploy the `CollateralTracker` instances in the factory using CREATE2 to ensure all system contracts have address spaces segregated by Uniswap pool. 
	- This was actually a reiteration of [this QA finding](https://github.com/code-423n4/2024-04-panoptic-findings/issues/573) from our April audit describing an extraordinarily unlikely scenario involving a large reorg redirecting a user's collateral deposit into a CollateralTracker with the same address, but owned by a different (malicious) Panoptic pool. 	
	- The attack remains extremely unlikely, but while writing my rebuttal I realized that a far more likely scenario involves a user submitting a collateral deposit and a pool deployment (perhaps a token launcher) in the same transaction and having their transaction fail because someone else deployed an unrelated pool first.
	- Or, someone could submit an underpriced pair deployment + collateral deposit transactions to the mempool and have their deployment frontrun by a malicious pool.
	- These aren't really realistic scenarios (and impacts are limited to interception of one or two (likely small) collateral deposits) with our current frontend setup, but might affect UX at some unspecified point in the future and I'm tired of auditors bugging me about it.
- Fixed an incorrect/outdated docstring.

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -324,13 +324,13 @@ contract PanopticPool is ERC1155Holder, Multicall {
         _validateSolvency(user, positionIdList, BP_DECREASE_BUFFER);
     }
 
-    /// @notice Returns the total amount of premium accumulated for a list of positions and a list containing positions data.
+    /// @notice Returns the total amount of premium accumulated for a list of positions and a list containing the corresponding `PositionBalance` information for each position.
     /// @param user Address of the user that owns the positions
     /// @param positionIdList List of positions. Written as `[tokenId1, tokenId2, ...]`
     /// @param includePendingPremium If true, include premium that is owed to the user but has not yet settled; if false, only include premium that is available to collect
     /// @return The total amount of premium owed (which may `includePendingPremium`) to the short legs in `positionIdList` (token0: right slot, token1: left slot)
     /// @return The total amount of premium owed by the long legs in `positionIdList` (token0: right slot, token1: left slot)
-    /// @return A list of balances and pool utilization for each position, of the form `[[tokenId0, balances0], [tokenId1, balances1], ...]`
+    /// @return A list of `PositionBalance` data (balance and pool utilization/oracle ticks at last mint) for each position, of the form `[[tokenId0, PositionBalance_0], [tokenId1, PositionBalance_1], ...]`
     function getAccumulatedFeesAndPositionsData(
         address user,
         bool includePendingPremium,
@@ -1156,7 +1156,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     //////////////////////////////////////////////////////////////*/
 
     /// @notice Check whether an account is solvent at a given `atTick` with a collateral requirement of `buffer/10_000` multiplied by the requirement of `positionIdList`.
-    /// @dev This will revert if `account` is not solvent at all provided ticks and `expectedSolvent == true`, or if `account` is solvent at all ticks and `expectedSolvent == false`.
+    /// @dev Reverts if `account` is not solvent at all provided ticks and `expectedSolvent == true`, or if `account` is solvent at all ticks and `!expectedSolvent`.
     /// @param account The account to check solvency for
     /// @param positionIdList The list of positions to check solvency for
     /// @param currentTick The current tick of the Uniswap pool (needed for fee calculations)
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -324,14 +324,14 @@ contract PanopticPool is ERC1155Holder, Multicall {
         _validateSolvency(user, positionIdList, BP_DECREASE_BUFFER);
     }
 
-    /// @notice Compute the total amount of premium accumulated for a list of positions.
+    /// @notice Returns the total amount of premium accumulated for a list of positions and a list containing positions data.
     /// @param user Address of the user that owns the positions
     /// @param positionIdList List of positions. Written as `[tokenId1, tokenId2, ...]`
     /// @param includePendingPremium If true, include premium that is owed to the user but has not yet settled; if false, only include premium that is available to collect
     /// @return The total amount of premium owed (which may `includePendingPremium`) to the short legs in `positionIdList` (token0: right slot, token1: left slot)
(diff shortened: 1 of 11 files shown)


guil-lambert (approved):
LGTM!

contracts/PanopticPool.sol
```diff
@@ -324,14 +324,14 @@ contract PanopticPool is ERC1155Holder, Multicall {
         _validateSolvency(user, positionIdList, BP_DECREASE_BUFFER);
     }
 
-    /// @notice Compute the total amount of premium accumulated for a list of positions.
+    /// @notice Returns the total amount of premium accumulated for a list of positions and a list containing positions data.
```
flanagansteve: Nit: We could consider a more specific purpose statement here, especially given that the functions name is just "...positionData":

```
"and a list of position balances & pool utilisations."
```

But the @return variable description covers this, so not a big deal.

contracts/PanopticPool.sol
```diff
@@ -324,14 +324,14 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
     /// @param user Address of the user that owns the positions
     /// @param positionIdList List of positions. Written as `[tokenId1, tokenId2, ...]`
     /// @param includePendingPremium If true, include premium that is owed to the user but has not yet settled; if false, only include premium that is available to collect
     /// @return The total amount of premium owed (which may `includePendingPremium`) to the short legs in `positionIdList` (token0: right slot, token1: left slot)
     /// @return The total amount of premium owed by the long legs in `positionIdList` (token0: right slot, token1: left slot)
     /// @return A list of balances and pool utilization for each position, of the form `[[tokenId0, balances0], [tokenId1, balances1], ...]`
-    function calculateAccumulatedFeesBatch(
+    function getAccumulatedFeesAndPositionsData(
```
flanagansteve, line 335: Some teams avoid using `get` as a prefix for function names, arguing that it's redundant - if we followed that rule here, we could name this function `accumulatedFeesAndPositionsData()`, which will arguably feel less redundant when integrators call it in their code.

however, we already use `getX` in many places in this codebase, so probably better to just stick with that pattern for now..

contracts/PanopticPool.sol
```diff
@@ -1147,7 +1147,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     //////////////////////////////////////////////////////////////*/
 
     /// @notice Check whether an account is solvent at a given `atTick` with a collateral requirement of `buffer/10_000` multiplied by the requirement of `positionIdList`.
-    /// @dev This will return true if solvent at any of the provided tick, and return false iff the account is insolvent at all ticks.
+    /// @dev This will revert if `account` is not solvent at all provided ticks and `expectedSolvent == true`, or if `account` is solvent at all ticks and `expectedSolvent == false`.
```
flanagansteve: another nit: you could make this docstring more concise if you wanted:

"/// @dev Reverts if `account` is not solvent at all provided ticks and `expectedSolvent == true`, or if `account` is solvent at all ticks and `!expectedSolvent`."

But feel free to ignore if you like the original version better

flanagansteve (approved):
> and I'm tired of auditors bugging me about it.

LOL - exasperation-driven development

All changes make sense to me - great job - just some nits about docstrings and naming below:

contracts/PanopticPool.sol
dyedm1: It's actually not just the position balance and pool utilization, there's ticks at mint too. I think I'm going to change the natspec both here and in the return statement to refer to the `PositionBalance` type (which is documented elsewhere) anyway.

contracts/PanopticPool.sol
dyedm1: we love our verb prefixes

dyedm1: if we were actually being super anal about naming schemes I think we would have to spend a bunch of time changing things. Perhaps V2 will be a clean-slate of a high enough degree that we can implement a stricter styleguide.

contracts/PanopticPool.sol
dyedm1: Original version would make more sense if it was more of a parenthetical statement, but this is describing the normal operation of the function so I think the more concise version is better.

contracts/PanopticPool.sol
flanagansteve: >  I think I'm going to change the natspec both here and in the return statement to refer to the PositionBalance type (which is documented elsewhere) anyway.

Nice sounds great!

guil-lambert:
Do we need to add an unchecked box around the oracle tick conditions check (like we do in v1.1): https://github.com/panoptic-labs/panoptic-v1-core-private/blob/adfa057f1f7a7534ffc0a8b2fb95e6a3a985547e/contracts/PanopticPool.sol#L825-L839

PeakyCryptos (approved):
LGTM

flanagansteve (approved):
Noice, LGTM

PR #143, opened 2024-10-11
feat: add native currency support to Panoptic V1.1

dyedm1:
Was reading through the Uniswap contest findings, and found an "issue" in our own contracts hiding in plain sight: we forgot to add support for native currencies!

We'll add a couple targeted test cases in `Misc.t.sol` for this; shoehorning support for pools with native currencies into our regular test suite is too much effort given that we only have to handle with native currencies directly during pool creation, collateral deposits (native currencies are treated as first-class citizens in `take` so we don't have to branch for withdrawals), and negative-bonus liquidations.

A design discussion never occured around this feature, but I believe this is the best way to do it:
- remove `payable` modifier from `multicall` to prevent bugs related to reuse of `msg.value`
- make `deposit`, `mint`, and `deployNewPool` payable, call `settle` with the correct delta for native currencies, then refund any unused tokens.
- make `liquidate` payable, pass attached value onto (now payable) `settleLiquidation` for token0 (native cannot be token1) and use the same flow in `_settleTokenDelta` to handle negative bonuses in token0.
	- otherwise (no negative bonus), refund any attached native currency to the liquidator in `settleLiquidation`
	
The refunds are important because it’s not always possible to know the exact amount of ETH to send (except in deposit) at the time of transaction creation. This will avoid breaking any potential integrations because refunds only occur if value was attached to the calls (which was unsupported previously). We also remain ERC4626-compliant for non-native pools without having to add any new interfaces.

TODO:
- [x] Change base to #145


contracts/CollateralTracker.sol
```diff
@@ -346,29 +346,39 @@ contract CollateralTracker is Clone, ERC20Minimal, Multicall {
     /// @param account The address of the account to transfer the underlying token to/from
     /// @param delta The amount of the underlying token to wrap/unwrap and transfer
```
flanagansteve: You should possibly modify the @notice purpose statement:

```
/// @notice Initiates the unlock callback to wrap/unwrap `delta` amount of the underlying token (or handle `msg.value` of the chain's native token) and transfer to/from the Panoptic Pool.
```

contracts/CollateralTracker.sol
```diff
@@ -346,29 +346,39 @@ contract CollateralTracker is Clone, ERC20Minimal, Multicall {
...
         if (msg.sender != address(POOL_MANAGER_V4)) revert Errors.UnauthorizedUniswapCallback();
 
-        (address account, int256 delta) = abi.decode(data, (address, int256));
+        (address account, int256 delta, uint256 valueOrigin) = abi.decode(
+            data,
+            (address, int256, uint256)
+        );
 
         address underlyingToken = _underlyingToken();
         if (delta > 0) {
```
flanagansteve, line 366: Are we guaranteed by Uniswap that we don't have to handle the delta < 0 case with a native-currency `underlyingToken`? Perhaps because Uniswap is always returning WETH in that case or something like that?

contracts/CollateralTracker.sol
```diff
@@ -346,29 +346,39 @@ contract CollateralTracker is Clone, ERC20Minimal, Multicall {
...
-            POOL_MANAGER_V4.sync(Currency.wrap(underlyingToken));
-            SafeTransferLib.safeTransferFrom(
-                underlyingToken,
-                account,
-                address(POOL_MANAGER_V4),
-                uint256(delta)
-            );
-            POOL_MANAGER_V4.settle();
+            if (Currency.wrap(underlyingToken).isAddressZero()) {
+                POOL_MANAGER_V4.settle{value: uint256(delta)}();
```
flanagansteve, line 368:
```suggestion
                uint256 deltaUint = uint256(delta);
                POOL_MANAGER_V4.settle{value: deltaUint}();
                 uint256 surplus = valueOrigin - deltaUint;
                if (surplus > 0) SafeTransferLib.safeTransferETH(account, surplus);
```

Might be more efficient to store `uint256(delta)` in a memory var since you cast it twice?

contracts/libraries/SafeTransferLib.sol
```diff
@@ -9,6 +9,24 @@ import {Errors} from "@libraries/Errors.sol";
...
 /// @dev Caution! This library won't check that a token has code, responsibility is delegated to the caller.
 library SafeTransferLib {
+    /*//////////////////////////////////////////////////////////////
+                             ETH OPERATIONS
+    //////////////////////////////////////////////////////////////*/
+
+    /// @notice Safely transfers ETH to a specified address.
+    /// @param to The address to transfer ETH to
+    /// @param amount The amount of ETH to transfer
+    function safeTransferETH(address to, uint256 amount) internal {
```
flanagansteve, line 19:
```suggestion
    function safeTransferETH(address to, uint256 amount) internal {
        if (amount) {
            bool success;

            assembly {
                // Transfer the ETH and store if it succeeded or not.
                success := call(gas(), to, amount, 0, 0, 0, 0)
            }

            if (!success) revert Errors.TransferFailed();
        }
    }
```

You could also consider moving your `surplus > 0` check to this method and that way you wouldn't need the `surplus` var in `CollateralTracker.unlockCallback`.  This version fails silently if `amount` is 0, which plays nicely with `unlockCallback`'s needs, but could be unintuitive to other users of this helper though - they might expect a proper revert

if you adopted this i'd probably also modify the docstring:

```
/// @notice Safely transfers a positive amount of ETH to a specified address.
/// @param to The address to transfer ETH to
/// @param amount The amount of ETH to transfer
```

contracts/CollateralTracker.sol
```diff
@@ -346,29 +346,39 @@ contract CollateralTracker is Clone, ERC20Minimal, Multicall {
...
+
+                uint256 surplus = valueOrigin - uint256(delta);
+                if (surplus > 0) SafeTransferLib.safeTransferETH(account, surplus);
```
flanagansteve, line 371: We now can trigger reentering loops by `call`ing with native currency values into unknown smart contracts. I believe that's OK - i don't think there's anything inherently wrong with someone interacting with Panoptic in a loop. Is this something you've considered / do we have any documented reasoning for what we expect could happen in that case / why we're DoS-via-reentrancy resistant?

I know we've gotten junk audit reports about not having `ReentrancyGuard` - maybe we've stated something in those threads (although it would have been purely academic before considering we were just doing erc20 transfers)?

contracts/CollateralTracker.sol
dyedm1: Nice catch -- I think a less verbose way of doing this is just `token` -> `asset`, which is what I've been doing in the docs.

contracts/CollateralTracker.sol
dyedm1: int256 and uint256 are the same size so this is a no-op. They only make you cast it in the first place to make it clear that you *might* accidentally be casting a possibly-negative number into an unsigned integer.

contracts/CollateralTracker.sol
dyedm1: Yes. Uniswap will send native ETH to the recipient on `take(address(0))`.

contracts/libraries/SafeTransferLib.sol
dyedm1: > if (amount) {

There is no way to cast anything to bool type in solidity without using comparison operators unfortunately :(

> way you wouldn't need the surplus var in CollateralTracker.unlockCallback.

When you see variables being assigned to an expression in code, that doesn't necessarily mean that the gas cost increases significantly over the expression itself, it just means that the stack item is duplicated and persisted when it gets use within its scope.

i.e. the current sequence of opcodes would be something like SUB DUP1 PUSH0 LT PUSHN jmpdest JUMPI
and the new one would be SUB PUSHN jmpdest JUMP DUP1 PUSH0 LT PUSHN jmpdest JUMPI (jumpback logic based on that if statement...), which is not a clear improvement and may even cost more gas (I have of course omitted stack management for the other variables).

ceteris paribus I think the current version is a little more explicit about the behavior on different `surplus` values

contracts/CollateralTracker.sol
dyedm1: I determined it was safe because we always require a successful Uniswap currency mint before creating any shares. It's possible to capture some of your own commission or deposit fees (because the share price may change between the share conversion and share mint during a reentrant call). However, this does not provide a unique financial advantage. If you capture your own commission fees, the end state would be the same as if you had simply fee deposited beforehand, and if you capture your own mint fees, the effect would be the same as if you had deposited all at once.

The latter is because your effective mint fee decreases as the share of the pool you own post-deposit increases. The price that you mint shares at does not include the effect of your own mint fee on the existing share price; only the absolute amount of the mint fee itself is considered. 

Commission fee (deposit before mint, assume no mint fee):
- 100 shares : 300 assets
- deposit: +300 shares, +900 assets
- 400 shares : 1200 assets
- commission fee payment: -50 shares
- 350 shares : 1200 assets

Commission fee (reentrant mint):
- 100 shares : 300 assets
- deposit start (queued: +300 shares, +900 assets)
- 100 shares : 300 assets
- Commission fee payment: -50 shares
- 50 shares: 300 assets
- (deposit end, +300 shares, +900 assets)
- 350 shares : 1200 assets

Deposit (all-in-one, 9.09090909...% mint fee):
- 100 shares, 1000 assets
- (deposit 110 assets)
- 200 shares, 1110 assets

Deposit (reenter half of the deposit):
- 100 shares, 1000 assets
- deposit start (queued: +50 shares, +55 assets)
- reentrant deposit: (+50 shares, +55 assets)
- 150 shares: 1055 assets
- (deposit end, +50 shares, +55 assets)
- 200 shares, 1100 assets

contracts/CollateralTracker.sol
dyedm1: I actually reduced our reliance on the word `token` significantly. Let me know if you see any mistakes or spots I missed

contracts/CollateralTracker.sol
```diff
@@ -144,42 +144,42 @@ contract CollateralTracker is Clone, ERC20Minimal, Multicall {
...
-            underlyingIsToken0 := shr(0xf8, calldataload(add(offset, 20)))
+            underlyingIsCurrency0 := shr(0xf8, calldataload(add(offset, 20)))
         }
     }
 
-    /// @notice Retrieve the address of the underlying token.
-    /// @return The address of the underlying token
+    /// @notice Retrieve the address of the underlying asset.
+    /// @return The address of the underlying asset
     function _underlyingToken() internal pure returns (address) {
```
flanagansteve: If we're committing to changing all `token` references to `asset` or `currency`, what would you think of renaming this method to `underlyingAsset` or `underlyingCurrency` too then? Slight preference for the latter from me since `_currency0` and so on use currency

contracts/libraries/SafeTransferLib.sol
flanagansteve: > When you see variables being assigned to an expression in code, that doesn't necessarily mean that the gas cost increases significantly over the expression itself, it just means that the stack item is duplicated and persisted when it gets use within its scope.
> and the new one would be SUB PUSHN jmpdest JUMP DUP1 PUSH0 LT PUSHN jmpdest JUMPI (jumpback logic based on that if statement...), which is not a clear improvement and may even cost more gas (I have of course omitted stack management for the other variables).

I decided to go ahead and make a simple gas test to just compare the two approaches on the average non-zero transfer amount. The version I recommended is indeed a little more expensive. Thank you for the guidance on how to think about these - will make a gas test in the future prior to recommending something like that!

> ceteris paribus I think the current version is a little more explicit about the behavior on different surplus values

Its now a moot point regardless but i do hear you here

contracts/CollateralTracker.sol
flanagansteve: OK cool. And by similar logic, we have no issues with sequences such as:

- `mint -> liquidate -> mint` loops
- `mint -> burn` via reentrance rather than a multicall

I can't think of a way to abuse the former, and the latter seems indistinguishable from a multicall to me, but just throwing it out there

flanagansteve (approved):

contracts/CollateralTracker.sol
dyedm1: hmmmmm I don't really care about this being before the liquidator's solvency check. I can't think of anything specific, but I am becoming kind of uncomfortable with allowing someone to get a lock in an intermediate state with virtual shares still issued. Easy fix is just to move the delta settlement call after the virtual shares are dealt with, those currencies aren't going to be used until another call so the ordering doesn't really matter.
 
ps just realized that function is called settleTokenDelta lol, we need to change the name.

contracts/CollateralTracker.sol
dyedm1: yes I missed this, changing to underlyingAsset. Trying to use asset more than currency except when referring to Uniswap currency0 and currency1 because it's a uniswap-specific construct/data type

PeakyCryptos (commented):
Missed Token to Currency references:
TokenId.sol:
L134, L250

PositionBalance.sol:
L163

V4StateReader: 
L122

PanopticPool:
L39
L51

contracts/CollateralTracker.sol
```diff
@@ -445,23 +455,23 @@ contract CollateralTracker is Clone, ERC20Minimal, Multicall {
...
+    function deposit(uint256 assets, address receiver) external payable returns (uint256 shares) {
         if (assets > type(uint104).max) revert Errors.DepositTooLarge();
 
         shares = previewDeposit(assets);
 
-        // transfer assets (underlying token funds) from the user/the LP to the PanopticPool
+        // transfer assets from the user/the LP to the PanopticPool
         // in return for the shares to be minted
-        _settleTokenDelta(msg.sender, int256(assets));
+        _settleCurrencyDelta(msg.sender, int256(assets));
```
PeakyCryptos: Since we're following CEI(check-effects-interactions) you may want to move the settlement of currencies to after we adjust the tracked asset balances.

contracts/CollateralTracker.sol
```diff
@@ -497,22 +507,22 @@ contract CollateralTracker is Clone, ERC20Minimal, Multicall {
...
+    function mint(uint256 shares, address receiver) external payable returns (uint256 assets) {
         assets = previewMint(shares);
 
         if (assets > type(uint104).max) revert Errors.DepositTooLarge();
 
-        // transfer assets (underlying token funds) from the user/the LP to the PanopticPool
+        // transfer assets from the user/the LP to the PanopticPool
         // in return for the shares to be minted
-        _settleTokenDelta(msg.sender, int256(assets));
+        _settleCurrencyDelta(msg.sender, int256(assets));
```
PeakyCryptos: CEI: same as above comment

contracts/CollateralTracker.sol
```diff
@@ -606,7 +615,6 @@ contract CollateralTracker is Clone, ERC20Minimal, Multicall {
             if (allowed != type(uint256).max) allowance[owner][msg.sender] = allowed - shares;
         }
 
-        // burn collateral shares of the Panoptic Pool funds (this ERC20 token)
         _burn(owner, shares);
 
         // update tracked asset balance
```
PeakyCryptos, line 620: Missed an unchecked block on the line below L621

PeakyCryptos (changes requested):

PeakyCryptos:
I don't believe any of the increments to s_poolAssets in the collateralTracker can overflow due to the limits/constraints of our system. `s_poolAssets += uint128(assets)`. So we should be able to wrap these in an unchecked block, correct me if I'm wrong though

contracts/CollateralTracker.sol
dyedm1: that one is on purpose, it prevents users from withdrawing more than the pool assets (this overload of the withdraw function doesn't have that covered in a maxWithdraw call)

contracts/CollateralTracker.sol
dyedm1: We discussed reentrancy above, but thought about it again and I don't think it can hurt to move it lower. Might make it a little more auditable.

dyedm1:
> So we should be able to wrap these in an unchecked block

Yes, technically, but I think capping the pool assets at `type(uint128).max` is still an extra layer of protection, so it's worth a few extra gas.

PR #144, opened 2024-10-21
fix: increase capital requirement for liquidity addition DoS

dyedm1:
Implementing a solution to this liquidity addition DoS we found. This PR removes the problematic full-range liquidity add in the PanopticFactory and implements further restrictions on the enforced tick range in the SFPM that yield a capital requirement approximately equal to the greater of a default token value we set or a percentage of the token supply. 

To compute the tick bounds, we use eqn 9 of [this desmos](https://www.desmos.com/calculator/jtu6qa3n4m) (the capital requirement is the same in both signs of tick space, so we just use the inverse function for the token1 requirement because it's simpler). For `log_sqrt(1.0001)(x)` we use an iterative log2 method similar to the one used in `getTickAtSqrtRatio` in Uniswap except with one less iteration allowing for an error of up to two ticks (for more information about the error bounds see some sketchy ABDK math [here](https://hackmd.io/@abdk/SkVJeHK9v#Base-sqrt10001-Logarithm-Approximation) or my code comments) and then change-of-base to `log_sqrt(1.0001)` (with a different constant than Uniswap's because they used some weirdly low level of precision to compute it and I can't reproduce).

Tick bounds are computed at initialization in the SFPM and emitted in the event. Also, there is a separate function that can recompute and potentially increase (not increase) the available tick range if the token supply for either token decreased that emits its own event. The purpose of this function is to prevent flash mints for tokens like DAI creating artificially narrow tick bounds that can't be widened post-initialization.

I still need to figure out the best configuration for the tick bounds. Depending on the bounds chosen, additional validation/limits may be required to prevent the tick calculation from breaking across all tickSpacings.

Also need to add some more tests besides the differential one for `log_Sqrt1p0001`

See original message below:

I found this [liquidity addition DoS](https://panoptic-global.slack.com/archives/C04F33PCH0S/p1714606249016749) related to the maxLiquidityPerTick limit (that is also in the [Certora V4 audit](https://github.com/Uniswap/v4-core/blob/main/docs/security/audits/DRAFT_Certora_audit_core.pdf) as L-02) a few months back, and at the time we concluded that it was reversible because you can decrease the net liquidity at any given tick by minting an adjacent chunk.I was rereading some of the V4 issues and I realized that this is not actually true (the DoS is non-reversible by an external actor --- someone has to remove liquidity). The max liquidity limit is applied to the **gross** liquidity of a tick (of which the sole purpose is to enforce that limit), which is the sum of both the positive and negative components of liquidityNet. Given the implications, it's kind of shocking to me that this has continued to be acceptable to the Uniswap team through V4 and they haven't found an alternate solution --- it only takes 2^106 tokens to DoS an entire 1bps pool and **significantly** less to DoS a realistic price range for many pairs (If you consider a SHIB-ETH-1bps pool, it would only take 5 figures to block +/- 20% around the main price). This will probably break someone's hook somewhere.Anyway, what are the implications for us? I created this [Desmos](https://www.desmos.com/calculator/r26zkeccvr) to help find out:

-   New pool creations can be prevented indefinitely (through their full-range liquidity add) at a very low cost on most pools (as the Desmos illustrates, the cost of filling the maxLiquidity as the ticks go to their limits approaches zero. The actual minimum cost on 1bps is something like 500 million wei which is peanuts for most pools with at least one 18-decimal token. This cost increases along with the tick spacing, but remains economical for the ticks touched by the factory.)
-   Users can prevent themselves from being liquidated or force exercised indefinitely by selling themselves a very-far-out-of-range position, adding the long leg to the position they want to keep open, then filling the liquidity in one of the ticks of that long leg to the maxLiquidityPerTick.
-   On some pools (USDe-USDT-1bps, PEPE-ETH-30bps, etc.), it is economically feasible to prevent users with open long legs around the current price from closing their positions by filling the maxLiquidityPerTick (USDe pool about $10k per tick pair, PEPE is about $30k)

What are our options here? Solving the issue at its root would require us to create our own AMM (i.e. fork Uniswap and redesign the accounting model to raise the liquidity cap), which probably wouldn't be all that hard technically but isn't feasible from a business perspective. We pretty much have to work around it for now. Here's a few things we could do:

-   The liquidity added by the factory isn't really load-bearing. We could remove it entirely, constrain the tick bounds such that filling the max liquidity at that tick would exceed some token threshold we set (based on the `tickSpacing` and perhaps the token supply), or retry until we find a set of tick bounds that works. The second option likely makes the most sense if we want to keep it, but removing the **full-range** aspect might defeat the point of adding the liquidity in the first place (because it no longer prevents liquidity gaps).
-   To prevent DoS on closing long positions, there's a few approaches we could take:
    -   Create a Uniswap hook that effectively lowers the `maxLiquidityLimit` to `maxLiquidityLimit - removedLiquidity` by controlling liquidity adds. This would solve any serious DoS issues, but would preclude us from using V3 pools and limit V4 usage to pools using our hookbase with that functionality
    -   Implement a mechanism to force-close or asynchronously settle all (or all long?) positions touching a given chunk if a maxLiquidityLimit error occurs in the Uniswap pool during a position close. The force-close one can cause unbounded gas consumption (so we would need a very stringent per-chunk position limit), and the asynchronous settlement, while doable, would require a bunch of extra logic to add a "ban" flag to a chunk, track the settlement state, and propagate that state through collateral checks and burns as the chunk unwinds.
    -   Limit the min/max tick during TokenId valuation to prevent positions being created with ticks that are economically feasible to fill. This won't work well for every pool (as I mentioned above there are some pools where it is economically feasible to fill ticks adjacent to the current price, so for those pools it will either be impossible to mint in-range options or they will remain at risk of a DoS that prevents positions from being liquidated), but I went through a bunch of the top pools on the Uniswap analytics and restricting the available tickrange to a point where filling a tick costs $100k+ wouldn't affect most of them (though, I suspect that distribution might shift as we go to memecoin/weird-fee-tier/tickspacing pools with low TVL)

On the factory I'm okay with either removing the liquidity add or constraining the bounds.I think our only realistic option (for now, at least) to mitigate long position DoS is constraining the min/max ticks. The new enforced min/max tick will be the result of a formula with inputs of `tickSpacing` and `minAcceptableTokenDOSCost` (such that creating a position `(outputTick, outputTick + tickSign * tickSpacing)` with `maxLiquidityLimit` liquidity will require `minAcceptableTokenDOSCost` if the current tick falls within the enforced limits).Unfortunately, we're kind of limited with how high we can set `minAcceptableTokenDOSCost` because setting it too high will effectively brick pools by pushing the min tick above or the max tick below the current price. Given that, using the token supply for anything load-bearing is typically a bad idea, but we could consider using something like `max(tokenSupply/x, (lower)minAcceptableDOSCost)` , which will allow us to work with a wider range of potential pools that we can whittle down if necessary in the UI by calculating the USD amount of tokens required to fill the tick limit set in the smart contracts.


contracts/PanopticFactory.sol
```diff
@@ -165,20 +110,13 @@ contract PanopticFactory is FactoryNFT, Multicall {
     /// @param token1 Address of token1 for the underlying Uniswap V3 pool
     /// @param fee The fee tier of the underlying Uniswap V3 pool, denominated in hundredths of bips
     /// @param salt User-defined component of salt used in CREATE2 for the PanopticPool (must be a uint96 number)
-    /// @param amount0Max The maximum amount of token0 to spend on the full-range deployment
-    /// @param amount1Max The maximum amount of token1 to spend on the full-range deployment
     /// @return newPoolContract The address of the newly deployed Panoptic pool
     function deployNewPool(
         address token0,
```
flanagansteve, line 115: If we're no longer going to sort these (~~which does resolve that cantina finding~~ nevermind, forgot its no longer relevant anyways without a mint here), then perhaps we should call these tokenA and tokenB like the non-sorting methods in uniswap do:

```
/// @notice Returns the pool address for a given pair of tokens and a fee, or address 0 if it does not exist
    /// @dev tokenA and tokenB may be passed in either token0/token1 or token1/token0 order
    /// @param tokenA The contract address of either token0 or token1
    /// @param tokenB The contract address of the other token
    /// @param fee The fee collected upon every swap in the pool, denominated in hundredths of a bip
    /// @return pool The pool address
    function getPool(
        address tokenA,
        address tokenB,
        uint24 fee
    ) external view returns (address pool);
```

contracts/PanopticFactory.sol
```diff
@@ -222,24 +160,11 @@ contract PanopticFactory is FactoryNFT, Multicall {
         // When that happens, there will be a period of time where the PanopticPool is deployed, but not (safely) usable
```
flanagansteve, line 163: collateralTracker0/1 are now unsorted - is that fine?

TODO: steve to investigate this and expand here if applicable

flanagansteve, line 163: TODO: Check that `startPool(v3Pool, token0, token1, collateralTracker0, collateralTracker1);` is fine with unsorted tokens

contracts/SemiFungiblePositionManager.sol
```diff
@@ -314,27 +353,119 @@ contract SemiFungiblePositionManager is ERC1155, Multicall, TransientReentrancyG
...
+        uint256 maxLiquidityPerTick = Math.getMaxLiquidityPerTick(tickSpacing);
+
+        int24 minEnforcedTick;
+        int24 maxEnforcedTick;
+        unchecked {
+            minEnforcedTick = int24(
+                Math.getApproxTickWithMaxAmount(
+                    Math.max(
+                        MIN_ENFORCED_TICKFILL_COST,
+                        (Math.min(IERC20Partial(token0).totalSupply(), uint128(type(int128).max)) *
```
flanagansteve, line 382: Is it correct to not care whether this `token0` is truly the same as `pool.token0` (given that deployNewPool no longer sorts token0 and token1)?

flanagansteve: In `expandEnforcedTickRange` you then do:

```
dataOld.pool.token0()
```

which makes me think the two could be using different token0s in this initializeAmmPool call compared to a subsequent expand call

contracts/SemiFungiblePositionManager.sol
```diff
@@ -736,6 +867,23 @@ contract SemiFungiblePositionManager is ERC1155, Multicall, TransientReentrancyG
                 positionSize
             );
 
+            // validate tick range for newly minted positions
+            if (!isBurn) {
+                int24 tickSpacing = tokenId.tickSpacing();
+                int24 tickLower = liquidityChunk.tickLower();
+                int24 tickUpper = liquidityChunk.tickUpper();
+
+                // Revert if the upper/lower ticks are not multiples of tickSpacing
```
flanagansteve, line 879: If this one is a silent revert in univ3Pool.mint, why bother checking here at all instead of just letting it revert there eventually? just want to supply a better error message?

flanagansteve (commented):
> Also, there is a separate function that can recompute and potentially increase (not increase) the available tick range if the token supply for either token decreased that emits its own event.

I think you meant (not decreased) in the parenthetical here right?

Anyway - main thought here is about whether SFPM.initializeAmm and SFPM.expandRange could possibly use different token0s than one another, and if that's an issue

I also didn't check your math super thoroughly yet but can do so later if helpful. Wanted to get this back to you first though

contracts/PanopticFactory.sol
flanagansteve: Yes it is, apart from the fact that your CollateralTokens could be flipped now too, which i need to check. but the method's own logic is fine with any ordering

contracts/PanopticFactory.sol
flanagansteve, line 115: Also flagging that the NFTs use pool.token0/token1 to set symbol0 and symbol1, so these would now be inconsistent with the under-the-hood truth of which tokens correspond to collateralTracker0/1 in our pool

contracts/SemiFungiblePositionManager.sol
flanagansteve: And moreover - i have a vague idea in my head that you need to use the correct token0/1 ordering here so that you're applying your `* -1` on the correct one of these

dyedm1: You can stop looking for token sort issues now 🤣

I forgot that Uniswap also populates the getPool mapping in reverse, so I was assuming a revert on mis-ordered tokens (we used to calculate the pool address ourselves, which is order-sensitive). 

Will  add back token sort in the factory and also add it to the SFPM.

flanagansteve: Sounds good!

contracts/SemiFungiblePositionManager.sol
dyedm1: yeah this was/is a super easy failure case to hit (if you don't have software validating the strikes you pick) and they have... no error message whatsoever

contracts/PanopticFactory.sol
dyedm1: we are now sorting them again

PR #145, opened 2024-10-22
fix: increase capital requirement for liquidity addition DoS (v1.1)

dyedm1:
Port of #144 to v1.1


contracts/PanopticFactory.sol
```diff
@@ -199,15 +113,11 @@ contract PanopticFactory is FactoryNFT, Multicall {
     /// @param oracleContract The external oracle contract to be used by the newly deployed Panoptic Pool
     /// @param key The Uniswap V4 pool key
     /// @param salt User-defined component of salt used in deployment process for the PanopticPool
-    /// @param amount0Max The maximum amount of token0 to spend on the full-range deployment
-    /// @param amount1Max The maximum amount of token1 to spend on the full-range deployment
     /// @return newPoolContract The address of the newly deployed Panoptic pool
     function deployNewPool(
```
flanagansteve, line 116: More of a V1.1 comment: I just realised the `@dev` in this docstring  needs to be updated:

```
/// @dev A Uniswap pool is uniquely identified by its tokens and the fee.
```

Should be just removed i believe for v4?

dyedm1: WIll do this on the premium settlement PR so I don't have to pass the changes up through the chain

PR #146, opened 2024-10-22
feat: emit commissions in `optionMinted`

dyedm1:
cc @0xTARC

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1002,20 +1002,20 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param shortAmount The amount of shorts
     /// @param swappedAmount The amount of tokens moved during creation of the option position
     /// @param isCovered Whether the option was minted as covered (no swap occured if ITM)
-    /// @return utilization The final utilization of the collateral vault
+    /// @return The final utilization of the collateral vault
+    /// @return The total amount of commission (base rate + ITM spread) paid
     function takeCommissionAddData(
         address optionOwner,
         int128 longAmount,
         int128 shortAmount,
         int128 swappedAmount,
         bool isCovered
-    ) external onlyPanopticPool returns (uint32 utilization) {
+    ) external onlyPanopticPool returns (uint32, uint128) {
         unchecked {
             // current available assets belonging to PLPs (updated after settlement) excluding any premium paid
             int256 updatedAssets = int256(uint256(s_poolAssets)) - swappedAmount;
 
-            // constrict premium to only assets not belonging to PLPs (i.e premium paid by sellers or collected from the pool earlier)
-            int256 tokenToPay = _getExchangedAmount(
+            (int256 tokenToPay, uint128 commission) = _getExchangedAmount(
                 longAmount,
                 shortAmount,
                 swappedAmount,
@@ -1045,7 +1045,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             s_poolAssets = uint256(updatedAssets).toUint128();
             s_inAMM = uint256(int256(uint256(s_inAMM)) + (shortAmount - longAmount)).toUint128();
 
-            utilization = uint32(_poolUtilization());
+            return (uint32(_poolUtilization()), commission);
         }
     }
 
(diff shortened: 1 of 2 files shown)


contracts/CollateralTracker.sol
```diff
@@ -1103,38 +1103,31 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param swappedAmount The amount of tokens moved during creation of the option position
     /// @param isCovered Whether the option was minted as covered (no swap occured if ITM)
     /// @return exchangedAmount The amount of funds to be exchanged for minting an option (includes commission, swapFee, and intrinsic value)
```
flanagansteve: Technically, the return values no longer have the var names `exchangedAmount` / `commission`, so you should arguably remove these names from the docstrings.

But i kind of like the clarity that an unofficial name for the unnamed return val provides, so also fine with keeping these here

contracts/PanopticPool.sol
```diff
@@ -81,11 +81,13 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @param poolUtilizations Packing of the pool utilization (how much funds are in the Panoptic pool versus the AMM pool at the time of minting),
     /// right 64bits for token0 and left 64bits for token1, defined as `(inAMM * 10_000) / totalAssets()`
     /// where totalAssets is the total tracked assets in the AMM and PanopticPool minus fees and donations to the Panoptic pool
+    /// @param commissions The total amount of commissions (base rate + ITM spread) paid for token0 (right) and token1 (left)
     event OptionMinted(
         address indexed recipient,
         uint128 positionSize,
         TokenId indexed tokenId,
-        uint128 poolUtilizations
+        uint128 poolUtilizations,
```
flanagansteve: I was chatting with @guil-lambert about the gas tradeoff of adding this onto the event - he posed some other ideas that i'll post for him here:

> The gas costs could actually remain the same. We’re emitting poolUtilizations as a uint128, but it is now being tracked as a uint32 like [here](https://github.com/panoptic-labs/panoptic-v1-core-private/blob/30710db652ffc52a1b684aaa2c3122eeb29b0a58/contracts/PanopticPool.sol#L579C9-L579C15).

> Maybe we need to check, but the costs are per 32byte word and we may be “wasting” one word. This current structure is 4 words:

```
event OptionMinted(
address indexed recipient, <- 1st word
uint128 positionSize, <- second word
TokenId indexed tokenId, <- third word
uint128 poolUtilizations <- fourth word
);
```

> By splitting the two uint128, maybe we don’t need 4 words but only 3? Or we can simply keep the same structure and have the last item by:

```
|--- commission1 ---|--- commission0 ---|--- poolUtilization1 ---|--- poolUtilization0 ---|
|----- 112 bits ----|--- 112 bits ------|-------- 16 bits -------|-------- 16 bits -------|
```

What do you think? Seems like this kind of packing might be a _bit_ annoying to deal with in the subgraph logic, but not insurmountable, and would mean getting this data into the event without having to increase mint cost

flanagansteve (approved):
Changes look GTM - evaluated the new `_getExchangedAmounts` and it indeed seems to be the same calculations, just simplified down + cleaned up - nice stuff!

contracts/CollateralTracker.sol
dyedm1: Good catch, forgot to remove those. I like it too, but if we leave those names in, they get added to the variable description in the table on our generated markdown docs, which looks really weird...

contracts/PanopticPool.sol
dyedm1: > Maybe we need to check, but the costs are per 32byte word

No, the cost is per byte.

I have a better idea though, let's emit the `PositionBalance` from this event instead of the positionSize/poolUtilizations.

dyedm1: > TokenId indexed tokenId, <- third word
> address indexed recipient, <- 1st word

The indexed parameters are handled separately from the others and passed as stack arguments to the log opcode, and are priced per word. The rest is copied from a memory area and priced per byte.

flanagansteve (approved):
Awesome!

PR #148, opened 2024-10-23
feat: improve premium settlement, force exercise, and haircut APIs

dyedm1:
@0xTARC and I were working through some subgraph sync issues, and I realized that we weren't emitting enough information to track the premium settlement state.

This PR adds the following event changes to allow tracking of per-chunk settlement states. Closing the loop here will also enable correct handling of premium haircuts on the subgraph:
- Emit premium by leg in `OptionBurned`
- Emit leg in `PremiumSettled` event for `settleLongPremium(user, tokenId, leg)`
- Emit a `PremiumSettled` event with *negative* for every leg haircut in `haircutPremia` during a liquidation

Also, `forceExercise` was still working with the exercised position in the format `TokenId[touchedId]`. The original reasoning behind that design decision is no longer valid (i.e., we no longer compute the fees for the singular force exercised position, which would require an array of positions in `calldata` to be passed into `calculateAccumulatedFeesBatch`), so we can save some gas and clean up the interface there a little by operating on `touchedId` directly.

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -721,7 +721,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @param commitLongSettled Whether to commit the long premium that will be settled to storage (disabled during liquidations)
     /// @param positionIdList The list of option positions to close
     /// @return netPaid The net amount of tokens paid after closing the positions
-    /// @return premiasByLeg The amount of premia paid by the user for each leg of the position
+    /// @return premiasByLeg The amount of premia settled by the user for each leg of the position
     function _burnAllOptionsFrom(
         address owner,
         int24 tickLimitLow,
@@ -753,7 +753,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @param tickLimitHigh The upper bound of an acceptable open interval for the ending price on each option close
     /// @param commitLongSettled Whether to commit the long premium that will be settled to storage (disabled during liquidations)
     /// @return paidAmounts The net amount of tokens paid after closing the position
-    /// @return premiaByLeg The amount of premia paid by the user for each leg of the position
+    /// @return premiaByLeg The amount of premia settled by the user for each leg of the position
     function _burnOptions(
         bool commitLongSettled,
         TokenId tokenId,
@@ -763,9 +763,8 @@ contract PanopticPool is ERC1155Holder, Multicall {
     ) internal returns (LeftRightSigned paidAmounts, LeftRightSigned[4] memory premiaByLeg) {
         uint128 positionSize = s_positionBalance[owner][tokenId].positionSize();
 
-        LeftRightSigned premiaOwed;
         // burn position and do exercise checks
-        (premiaOwed, premiaByLeg, paidAmounts) = _burnAndHandleExercise(
+        (premiaByLeg, paidAmounts) = _burnAndHandleExercise(
             commitLongSettled,
             tickLimitLow,
             tickLimitHigh,
@@ -863,8 +862,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @param tokenId The option position to burn
     /// @param positionSize The size of the option position, expressed in terms of the asset
     /// @param owner The owner of the option position
-    /// @return realizedPremia The net premia paid/received from the option position
-    /// @return premiaByLeg The premia paid by the user for each leg of the option position
+    /// @return premiaByLeg The premia settled by the user for each leg of the option position
     /// @return paidAmounts The net amount of tokens paid after closing the position
     function _burnAndHandleExercise(
         bool commitLongSettled,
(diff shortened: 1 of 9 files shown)


contracts/PanopticPool.sol
```diff
@@ -1124,7 +1121,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
         // Exercise the option
         // Turn off ITM swapping to prevent swap at potentially unfavorable price
-        _burnAllOptionsFrom(account, MIN_SWAP_TICK, MAX_SWAP_TICK, COMMIT_LONG_SETTLED, touchedId);
+        _burnOptions(COMMIT_LONG_SETTLED, tokenId, account, MIN_SWAP_TICK, MAX_SWAP_TICK);
```
flanagansteve, line 1116: The internal `_burnOptions` method should arguably be renamed to `burnOption` since it just accepts 1 option (<=>position<=>tokenId) to burn at a time

contracts/PanopticPool.sol
```diff
@@ -1880,6 +1877,11 @@ contract PanopticPool is ERC1155Holder, Multicall {
                     LeftRightSigned.wrap(int256(LeftRightUnsigned.unwrap(availablePremium)))
                 );
 
+                // update the base `premiaByLeg` value to reflect the amount of premium that will actually be settled
+                premiaByLeg[leg] = LeftRightSigned.wrap(
```
flanagansteve, line 1873: The docstring for this method's `premiaByLeg` still uses the "owed to" verbiage, but you use this same value as the value you eventually emit in `OptionsBurnt` which now has a docstring saying its the premia _paid by_ the user; should it also be updated like the OptionBurnt event's `.premiaByLeg` description was? perhaps to:

```
/// @return premiaByLeg The amount of premia paid by the user for each leg of the position
```

`realizedPremia` also does the same, but that one is more likely to be right

contracts/PanopticPool.sol
```diff
@@ -772,7 +774,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
             owner
```
flanagansteve, line 773: We now don't seem to use the `premiaOwed` var at all that is being saved from `_burnAndHandleExercise` (which corresponds to what _burnAndHandleExercise calls `realizedPremia`. Should we consider:

- Removing the local `premiaOwed` var from this method, and calling with an empty return slot there (e.g. `        (, premiaByLeg, paidAmounts) = _burnAndHandleExercise(...`
- Going farther and just removing the `realizedPremia` return variable entirely from `_burnAndHandleExercise`?

contracts/SemiFungiblePositionManager.sol
```diff
@@ -1235,18 +1235,26 @@ contract SemiFungiblePositionManager is ERC1155, Multicall, TransientReentrancyG
...
             liquidityChunk,
             false
         ).subRect(s_accountFeesBase[positionKey]);
 
         if (isLong == 1) {
-            amountToCollect = amountToCollect.sub(movedInLeg);
+            amountToCollect = LeftRightUnsigned.wrap(
+                uint256(
+                    LeftRightSigned.unwrap(
+                        LeftRightSigned.wrap(int256(LeftRightUnsigned.unwrap(amountToCollect))).sub(
```
flanagansteve, line 1249: Agh. Would it be worth it to just add a `function sub(LeftRightUnsigned x, LeftRightSigned y)` into LeftRight.sol to avoid this casting dance?

contracts/libraries/PanopticMath.sol
```diff
@@ -935,39 +936,64 @@ library PanopticMath {
...
-                            uint128(-_premiasByLeg[i][leg].rightSlot()) * uint256(haircut0),
-                            uint128(longPremium.rightSlot())
+                        haircutTotal = haircutTotal.add(
+                            LeftRightUnsigned.wrap(uint256(LeftRightSigned.unwrap(haircutAmounts)))
                         );
-                        uint256 settled1 = Math.unsafeDivRoundingUp(
-                            uint128(-_premiasByLeg[i][leg].leftSlot()) * uint256(haircut1),
-                            uint128(longPremium.leftSlot())
+
+                        emit PanopticPool.PremiumSettled(
```
flanagansteve, line 992: We could consider moving some event definitions to a shared file so you don't have to import PanopticPool just for this. Could maybe combine it with Errors.sol (although we don't yet use that here so would be a fresh import either way)

flanagansteve (commented):
Tried to follow the changes in `haircutPremia` and ensure we are still doing the same operations that we were doing before, just more concisely. I am fairly sure you transformed it correctly, but calling this out for other reviewers too to triple check

Main thought is about `_burnAndHandleExercise`'s `premiaOwed` var - lmk if i'm missing some way it is indeed used

contracts/PanopticPool.sol
dyedm1: The external interfaces are all plural, so I think it makes sense to keep this. Each position can have up to 4 "options" (legs) anyway.

contracts/PanopticPool.sol
dyedm1: Changed all to "settled by"

contracts/PanopticPool.sol
dyedm1: I assumed it was returned/used somewhere at the top level, but guess not... we'll just stop passing it up after that function

contracts/SemiFungiblePositionManager.sol
dyedm1: I thought about it, but writing such a function to the same standard would require more gas to safely cast the result back into unsigned integers. These are all no-ops anyway so harmless despite being slightly unwieldy.

contracts/libraries/PanopticMath.sol
dyedm1: This event is only ever emitted from the `PanopticPool` contract (*including* during this function call) so I think it makes more sense to keep it there. This function is effectively a part of `PanopticPool` (it even modifies internal storage state), the only reason why it's here is to offload bytecode size.

PR #150, opened 2024-10-23
feat: improve premium settlement and force exercise events and API (v1.1)

dyedm1:
Port of #148 to v1.1


flanagansteve (commented):
These changes seem to be a subset of #148 , so same comments mostly apply

PR #158, opened 2025-08-26
test: whitehat contracts

flanagansteve:
good to have these if we need them. Run each test like:

```bash
FOUNDRY_PROFILE=attack_test forge test --match-test testTakeFlashLoanAndAttack_USDC_WETH30bpsMainnetAttacker -vvv
```

TODO:

1. Convert the test file & test attacker, `MakingFakeTokenIdListPassTest` and `USDC_WETH30bpsMainnetAttacker`, into a deploy script (remove console logs, move constants into storage vars, use a real `withdrawer` (going with vault.panoptic.eth for now))
  a. I'll set this up such that we deploy 1 or more Attacker contracts, and call takeFlashLoanAndAttack, all inside one multicall 

Then, if we have to use these its a matter of running that script and withdrawing later. Since we're trying to protect funds on multiple chains, we should queue the transaction from the deploy script to execute on each chain simultaneously.

next TODO: deal with other markets - next one is ETH/USDC 5bps on Base

1. get target fingerprint for the two deep-ITM real positions to open on ETH/USDC 5bps on Base
2. mine for fake_token_uints for that target fingerprint
3. insert into `ETH_USDC5bpsBaseAttacker`
4. adapt ETH_USDC5bpsBaseAttacker to work with native ETH (currently treats token0 and token1 both as erc20s)
5. get tests to pass & prove profitability
6. Convert ETH_USDC5bpsBaseAttacker into a deployable contract and insert into the deploy script

Then, next TODO could be either to repeat the above TODO for more pools, or optimise a bit:

1. for both ETH_USDC5bpsBaseAttacker and USDC_WETH30bpsMainnetAttacker, find the right balance of maximising the size of each position & withdrawal within `drainPool` + the number of times to call `drainPool` to get the most out of the protocol without hitting `AccountInsolvent()`
2. make whole thing more gas efficient and see if that lets you increase amount extracted in (1) (e.g. more gas-efficient drainPool can be called more times if the block gas limit is your barrier)


test/foundry/core/USDC_WETH30bpsMainnetAttacker.sol
```diff
@@ -353,7 +353,7 @@ contract Attacker is IFlashLoanReceiver {
     function phase1Attack() public {
         drainPool(); // takes 1/2             = 1/2 of s_poolAssets
         drainPool(); // takes 1/2 + 1/4       = 3/4 of s_poolAssets
-        drainPool(); // takes 1/2 + 1/4 + 1/8 = 7/8 of s_poolAssets
+        /* drainPool(); // takes 1/2 + 1/4 + 1/8 = 7/8 of s_poolAssets */
```
flanagansteve, line 356: (Market conditions changed since yesterday, so i get accountInsolvent if i try to run this with all 3 drains. you could probably just tweak the 20% margin applied within drainPool for a similar affect / get a little more out)

flanagansteve:
Status EoD tuesday: I have a local modification of the test contracts that could be deployed and executes the logic described. I am now writing a test that explicitly checks that deployable contract's effects - token0/1.balanceOf(deployableContract) before and after takeFlashLoanAndAttack

it then also checks the withdrawability

Am not pushing it up to this branch for security reasons

i am trying to run that test, but in an offline environment, which is a technical challenge - see this thread: https://panoptic-global.slack.com/archives/C09CY4V0QN4/p1756264335644799

PR #159, opened 2025-09-10
Feat1: add per-block interest accrual system to CollateralTracker

guil-lambert:
This PR introduces a protocol-wide interest accrual mechanism into CollateralTracker, along with math utilities and tests. 

This is in direct response to the commissions-based yield being too low and not reactive enough to allow for new capital to respond to higher yields (yields are ex post).

Not doing it here, but this means the size of the commissions will need to be adjusted down to reflect on the interest-rate model being the main source of yield of passive LPs instead of pure commissions (will still keep a small commission rate to prevent attacks)

### Global accumulator (s_interestRateAccumulator) 

- Left slot: `unrealizedGlobalInterest` (uint128)
- Right slot: packed (`lastBlock:uint32`, `borrowIndex:uint96`)
- Updated on every accrual to track protocol-wide interest growth.

### Per-account state (s_interestState)

- Left slot: `netBorrows` (int128)
- Right slot: `baseIndex` snapshot (int128)
- Used to compute user-level interest owed relative to global borrow index.

### Accrual flow

- New `_accrueInterest(owner)` called on deposits, withdrawals, transfers, exercises, settles, liquidations, refunds.
- Computes `deltaBlock` using `uint32` wrap-safe subtraction.
- Updates borrow index and `unrealizedGlobalInterest`.
- Charges user interest by burning shares, redistributing value to LPs.
- Clears user’s owed interest from global unrealized bucket.

### Accounting updates

- `totalAssets()` and `_poolUtilization()` now include unrealizedGlobalInterest.
- `interestState(address)` getter exposes per-user baseIndex + netBorrows.
- Added getters for `borrowIndex`, `lastInteractionBlock`, `unrealizedGlobalInterest`.

### Math library

- Added `mulDivWad` and `mulDivWadRoundingUp` optimized for `1e18` denominator
- Extensive tests against FullMath.mulDiv.

### Tests

- Added end-to-end scenarios for one-block and multi-block accrual, deposits without borrowing, borrowing via option mints, long/short flows, and negative borrow cases.
- Verified accumulator packing, borrow index evolution, global interest bucket behavior, and per-user owed interest.

### Notes / TODO

- Current interest rate is constant (≈20% APR per block basis) once utilization > 0. Utilization-dependent curve to be added later. Likely follow Morpho's PD controller on the log of the rate (using TickMath)
- While I added a test that involved selling+buying from multiple users, more tests to explore more complex scenarios

## Summary by CodeRabbit

* **New Features**
  * Protocol-wide block-based interest accrual with per-account tracking and a global borrow index; interest now affects balances, pool utilization, and margin health. New read-only views to inspect index, last interaction, unrealized global interest, per-account interest state, rate, and owed/previewed interest.

* **Libraries**
  * Added WAD fixed-point helpers and a Taylor-series utility for compounded math.

* **Style**
  * Minor whitespace cleanup.

* **Tests**
  * Extensive interest-accrual and WAD/math tests with new harness accessors for deterministic validation.

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.24;
-
+import "forge-std/Test.sol";
 // Interfaces
 import {PanopticPool} from "./PanopticPool.sol";
 // Inherited implementations
@@ -110,7 +110,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @notice Cached amount of assets accounted to be held by the Panoptic Pool — ignores donations, pending fee payouts, and other untracked balance changes.
     uint128 internal s_poolAssets;
 
-    /// @notice Amount of assets moved from the Panoptic Pool to the AMM.
+    /// @notice Amount of assets moved from the Panoptic Pool to the AMM, scaled with the current borrow index.
     uint128 internal s_inAMM;
 
     /// @notice Additional risk premium charged on intrinsic value of ITM positions,
@@ -123,6 +123,11 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @notice Interest rate accumulator, contains the last block timestamp in the upper 32 bits, the () in the next 96 bits, and the accumulator itself in the lower 128 bits in Wad.
     ///
     uint256 internal s_interestRateAccumulator;
+
+    /// @notice Mapping that tracks the collateral tracker's baseIndex at mint
+    /// @dev baseIndex is taking a snapshot of the interest rate accumulator in the collateral trackers, which is measuring the basis upon which
+    /// the interest rate will be computed
+    mapping(address account => LeftRightUnsigned interestState) internal s_interestState;
 
     /*//////////////////////////////////////////////////////////////
                             RISK PARAMETERS
@@ -264,7 +269,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         returns (uint256 poolAssets, uint256 insideAMM, uint256 currentPoolUtilization)
     {
         poolAssets = s_poolAssets;
-        insideAMM = Math.mulDivWad(s_inAMM, uint128(s_interestRateAccumulator));
+        insideAMM = s_inAMM;
         currentPoolUtilization = _poolUtilization();
         //TODO add: currentBorrowIndexWad = uint128(s_interestRateAccumulator);
     }
@@ -354,8 +359,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @return The total amount of assets managed by the CollateralTracker vault
     function totalAssets() public view returns (uint256) {
         unchecked {
-            return
-                uint256(s_poolAssets) + Math.mulDivWad(s_inAMM, uint128(s_interestRateAccumulator));
+            return uint256(s_poolAssets) + s_inAMM;
         }
     }
 
@@ -401,7 +405,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param receiver User to receive the shares
     /// @return shares The amount of Panoptic pool shares that were minted to the recipient
     function deposit(uint256 assets, address receiver) external returns (uint256 shares) {
-        accrueInterest();
+        accrueInterest(receiver);
         if (assets > type(uint104).max) revert Errors.DepositTooLarge();
         shares = previewDeposit(assets);
 
@@ -457,7 +461,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param receiver User to receive the shares
     /// @return assets The amount of assets deposited to mint the desired amount of shares
     function mint(uint256 shares, address receiver) external returns (uint256 assets) {
-        accrueInterest();
+        accrueInterest(receiver);
         assets = previewMint(shares);
 
         if (assets > type(uint104).max) revert Errors.DepositTooLarge();
@@ -515,7 +519,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         address receiver,
         address owner
     ) external returns (uint256 shares) {
-        accrueInterest();
+        accrueInterest(owner);
         if (assets > maxWithdraw(owner)) revert Errors.ExceedsMaximumRedemption();
 
         shares = previewWithdraw(assets);
@@ -560,7 +564,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         address owner,
         TokenId[] calldata positionIdList
     ) external returns (uint256 shares) {
-        accrueInterest();
+        accrueInterest(owner);
         shares = previewWithdraw(assets);
 
         // check/update allowance for approved withdraw
@@ -621,7 +625,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         address receiver,
         address owner
     ) external returns (uint256 assets) {
-        accrueInterest();
+        accrueInterest(owner);
         if (shares > maxRedeem(owner)) revert Errors.ExceedsMaximumRedemption();
 
         // check/update allowance for approved redeem
@@ -652,15 +656,14 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         emit Withdraw(msg.sender, receiver, owner, assets, shares);
     }
 
-    function accrueInterest() public {
+    function accrueInterest(address user) public {
         // Get block delta
         uint256 currentBlock = block.number;
         uint256 previousBlock = s_interestRateAccumulator >> 224;
+        uint128 deltaBlock = uint128(currentBlock - previousBlock);
 
         // return if same block
         if (currentBlock == previousBlock) return;
-
-        uint128 deltaBlock = uint128(currentBlock - previousBlock);
 
         //  extract interest rate owed during that period
         //uint128 inAmm = s_inAmm;
@@ -681,9 +684,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     }
 
     function interestRate() public view returns (uint128) {
-        //uint256 utilization = _poolUtilization();
-
-        return uint128(76103500761); // 0.2 * 10**18/(365*24*60*5) = 20% per year;
+        uint256 utilization = _poolUtilization();
+        return utilization == 0 ? uint128(0) : uint128(76103500761); // 0.2 * 10**18/(365*24*60*5) = 20% per year;
     }
 
     /*//////////////////////////////////////////////////////////////
@@ -793,9 +795,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @return poolUtilization The pool utilization in basis points
     function _poolUtilization() internal view returns (uint256 poolUtilization) {
         unchecked {
-            return
-                (Math.mulDivWad(s_inAMM * DECIMALS, uint128(s_interestRateAccumulator))) /
-                totalAssets();
+            return (s_inAMM * DECIMALS) / totalAssets();
         }
     }
 
@@ -1045,8 +1045,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         int128 shortAmount,
         int128 swappedAmount,
         bool isCovered
-    ) external onlyPanopticPool returns (uint32, LeftRightUnsigned) {
-        accrueInterest();
+    ) external onlyPanopticPool returns (uint32, uint128) {
+        accrueInterest(optionOwner);
         unchecked {
             // current available assets belonging to PLPs (updated after settlement) excluding any premium paid
             int256 updatedAssets = int256(uint256(s_poolAssets)) - swappedAmount;
@@ -1079,12 +1079,13 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // however, any intrinsic value is paid for by the users, so we only add the portion that comes from PLPs: the short/long amounts
             // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
             s_poolAssets = uint256(updatedAssets).toUint128();
-            s_inAMM = uint256(int256(uint256(s_inAMM)) + (shortAmount - longAmount)).toUint128();
 
-            return (
-                uint32(_poolUtilization()),
-                LeftRightUnsigned.wrap(commission).toLeftSlot(uint128(s_interestRateAccumulator))
-            );
+            {
+                int128 currentIndex = int128(uint128(s_interestRateAccumulator));
+                s_inAMM = uint256(int256(uint256(s_inAMM)) + ((shortAmount - longAmount)))
+                    .toUint128();
+            }
+            return (uint32(_poolUtilization()), commission);
         }
     }
 
@@ -1095,23 +1096,21 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param shortAmount The notional value of the short legs of the position (if any)
     /// @param swappedAmount The amount of tokens moved during the option close
     /// @param realizedPremium Premium to settle on the current positions
-    /// @param interestSnapshot snapshot of the interest rate a mint
     /// @return The amount of tokens paid when closing that position
     function exercise(
         address optionOwner,
         int128 longAmount,
         int128 shortAmount,
         int128 swappedAmount,
-        int128 realizedPremium,
-        uint128 interestSnapshot
+        int128 realizedPremium
     ) external onlyPanopticPool returns (int128) {
-        accrueInterest();
+        accrueInterest(optionOwner);
         unchecked {
             // current available assets belonging to PLPs (updated after settlement) excluding any premium paid
             int256 updatedAssets = int256(uint256(s_poolAssets)) - swappedAmount;
 
             uint128 netBorrows = uint128(uint256(Math.max(shortAmount - longAmount, 0)));
-
+            uint128 interestSnapshot = s_interestState[optionOwner].rightSlot();
             int128 netInterest = (netBorrows == 0 || interestSnapshot == 0)
                 ? int128(0)
                 : int128(
@@ -1123,6 +1122,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                         )
                     )
                 );
+            console2.log("exerciseData", netInterest);
 
             // add premium and token deltas not covered by swap to be paid/collected on position close
             int256 tokenToPay = int256(swappedAmount) -
@@ -1150,8 +1150,11 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             s_poolAssets = uint256(updatedAssets + realizedPremium).toUint128();
 
             // update the inAMM value, removing the part that was paid as interest (CHECK Math)
-            s_inAMM = uint256(int256(uint256(s_inAMM)) - (shortAmount - longAmount)).toUint128();
-
+            {
+                int128 currentIndex = int128(uint128(s_interestRateAccumulator));
+                s_inAMM = uint256(int256(uint256(s_inAMM)) - ((shortAmount - longAmount)))
+                    .toUint128();
+            }
             return (int128(tokenToPay));
         }
     }
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -119,9 +119,9 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @notice The fee of the Uniswap pool in hundredths of basis points.
     uint24 internal s_poolFee;
 
-    /// @notice Interest rate accumulator, contains the last block timestamp in the upper 32 bits, the () in the next 96 bits, and the accumulator itself in the lower 128 bits in Wad.
+    /// @notice Interest rate accumulator, contains the net interest owed in the left slot, the last block timestamp in the upper 32 bits of the right slow and the accumulator itself in the lower 96 bits in Wad of the right slot.
     ///
-    uint256 internal s_interestRateAccumulator;
+    LeftRightUnsigned internal s_interestRateAccumulator;
 
     /// @notice Mapping that tracks the collateral tracker's baseIndex at mint
     /// @dev rightSlot: baseIndex is taking a snapshot of the interest rate accumulator in the collateral trackers, which is measuring the basis upon which
@@ -239,7 +239,9 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         s_poolFee = fee;
 
         // store the initial block and initialize the borrowIndex
-        s_interestRateAccumulator = (block.number << 224) + 1e18;
+        s_interestRateAccumulator = LeftRightUnsigned.wrap(0).toRightSlot(
+            uint128((block.number << 96) + 1e18)
+        );
 
         // Stores the addresses of the underlying tracked tokens.
         s_univ3token0 = token0;
@@ -277,13 +279,19 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @notice Returns current borrowIndex.
     /// @return The borrowIndex
     function borrowIndex() external view returns (uint128) {
-        return uint128(s_interestRateAccumulator);
+        return uint96(s_interestRateAccumulator.rightSlot());
     }
 
     /// @notice Returns the last block at which interest rates were compounded.
     /// @return The last block at which the interest rates were compounded
     function lastInteractionBlock() external view returns (uint256) {
-        return (s_interestRateAccumulator >> 224);
+        return (s_interestRateAccumulator.rightSlot() >> 96);
+    }
+
+    /// @notice Returns the last block at which interest rates were compounded.
+    /// @return The last block at which the interest rates were compounded
+    function globalInterestOwed() external view returns (uint256) {
+        return s_interestRateAccumulator.leftSlot();
     }
 
     /// @notice Returns the baseIndex and the net borrows for the user
@@ -382,7 +390,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @return The total amount of assets managed by the CollateralTracker vault
     function totalAssets() public view returns (uint256) {
         unchecked {
-            return uint256(s_poolAssets) + s_inAMM;
+            uint128 globalInterestOwed = s_interestRateAccumulator.leftSlot();
+            return uint256(s_poolAssets) + s_inAMM + globalInterestOwed;
         }
     }
 
@@ -686,26 +695,28 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     function _accrueInterest(address owner) internal {
         // Get block delta
         uint256 currentBlock = block.number;
-        uint256 previousBlock = s_interestRateAccumulator >> 224;
+        uint256 previousBlock = s_interestRateAccumulator.rightSlot() >> 96;
         uint128 deltaBlock = uint128(currentBlock - previousBlock);
 
         // only update global quantities if not in the same block
-        uint128 currentBorrowIndex = uint128(s_interestRateAccumulator);
+        uint128 currentBorrowIndex = uint96(s_interestRateAccumulator.rightSlot());
+        uint128 globalInterestOwed = s_interestRateAccumulator.leftSlot();
         if (deltaBlock > 0) {
             // PROTOCOL
             //  extract interest rate owed during that period
             uint128 rawInterest = interestRate() * deltaBlock;
-            uint128 interestOwed = Math.mulDivWad(s_inAMM, rawInterest).toUint128();
+            uint128 interestOwed = Math
+                .mulDivWad(s_inAMM + globalInterestOwed, rawInterest)
+                .toUint128();
             // add that value to the "inAMM" tracker
-            s_inAMM += interestOwed;
+
+            globalInterestOwed += interestOwed;
 
             // Update the borrow index (check that math, it's probably wrong)
             uint128 borrowIndex = 10 ** 18 + rawInterest;
 
             // update current borrow index
             currentBorrowIndex = Math.mulDivWad(currentBorrowIndex, borrowIndex).toUint128();
-
-            s_interestRateAccumulator = (currentBlock << 224) + currentBorrowIndex;
         }
 
         // USER
@@ -715,13 +726,19 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             uint128 userInterestOwed = _getUserInterest(owner, userState, currentBorrowIndex);
             uint256 shares = previewWithdraw(userInterestOwed);
             _burn(owner, shares);
-            s_inAMM -= userInterestOwed;
 
-            s_interestState[owner] = LeftRightSigned
-                .wrap(0)
-                .toRightSlot(int128(currentBorrowIndex))
-                .toLeftSlot(netBorrows);
+            globalInterestOwed = userInterestOwed < globalInterestOwed
+                ? globalInterestOwed - userInterestOwed
+                : 0;
         }
+        s_interestRateAccumulator = LeftRightUnsigned
+            .wrap(0)
+            .toLeftSlot(globalInterestOwed)
+            .toRightSlot(uint128((currentBlock << 96) + uint96(currentBorrowIndex)));
+        s_interestState[owner] = LeftRightSigned
+            .wrap(0)
+            .toRightSlot(int128(currentBorrowIndex))
+            .toLeftSlot(netBorrows);
     }
 
     function interestRate() public view returns (uint128) {
@@ -747,7 +764,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
 
     function owedInterest(address owner) public view returns (uint128 interestOwed) {
         LeftRightSigned userState = s_interestState[owner];
-        uint256 borrowIndex = uint128(s_interestRateAccumulator);
+        uint256 borrowIndex = uint96(s_interestRateAccumulator.rightSlot());
         interestOwed = _getUserInterest(owner, userState, borrowIndex);
     }
 
@@ -858,7 +875,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @return poolUtilization The pool utilization in basis points
     function _poolUtilization() internal view returns (uint256 poolUtilization) {
         unchecked {
-            return (s_inAMM * DECIMALS) / totalAssets();
+            uint128 globalInterestOwed = s_interestRateAccumulator.leftSlot();
+            return ((s_inAMM + globalInterestOwed) * DECIMALS) / totalAssets();
         }
     }
 
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -241,7 +241,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
 
         // store the initial block and initialize the borrowIndex
         s_interestRateAccumulator = LeftRightUnsigned.wrap(0).toRightSlot(
-            uint128((block.number << 96) + uint96(1e18))
+            uint128((block.timestamp << 96) + uint96(1e18))
         );
 
         // Stores the addresses of the underlying tracked tokens.
@@ -702,19 +702,19 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // Get block delta
             LeftRightUnsigned accumulator = s_interestRateAccumulator;
 
-            // keep uint32 to allow for wrapping inside unchecked scope
-            uint32 currentBlock = uint32(block.number);
-            uint32 previousBlock = uint32(accumulator.rightSlot() >> 96);
-            uint128 deltaBlock = uint32(currentBlock - previousBlock);
+            // keep uint32 to allow for wrapping inside unchecked scope, works as long as the deltaTime is less than 2**32 seconds  = 136 years
+            uint256 currentTime = block.timestamp;
+            uint256 previousTime = accumulator.rightSlot() >> 96;
+            uint128 deltaTime = uint32(currentTime - previousTime);
 
             uint128 currentBorrowIndex = uint128(uint96(accumulator.rightSlot()));
             uint128 unrealizedGlobalInterest = accumulator.leftSlot();
             uint128 inAMM = s_inAMM;
             // only update global quantities if not in the same block
-            if (deltaBlock > 0) {
+            if (deltaTime > 0) {
                 // PROTOCOL
-                //  extract interest rate owed during that period
-                uint128 rawInterest = interestRate() * uint128(deltaBlock);
+                //  extract interest rate owed during that period using: rawInterest = (exp(-interest * ∆t) - 1)
+                uint128 rawInterest = interestRate() * uint128(deltaTime);
                 uint128 interestOwed = Math
                     .mulDivWadRoundingUp(inAMM + unrealizedGlobalInterest, rawInterest)
                     .toUint128();
@@ -757,13 +757,14 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             s_interestRateAccumulator = LeftRightUnsigned
                 .wrap(0)
                 .toLeftSlot(unrealizedGlobalInterest)
-                .toRightSlot(uint128((currentBlock << 96) + uint96(currentBorrowIndex)));
+                .toRightSlot(uint128((currentTime << 96) + uint96(currentBorrowIndex)));
         }
     }
 
+    /// @notice returns the interest rate per second
     function interestRate() public view returns (uint128) {
         uint256 utilization = _poolUtilization();
-        return utilization == 0 ? uint128(0) : uint128(76103500761); // 0.2 * 10**18/(365*24*60*5) = 20% per year;
+        return utilization == 0 ? uint128(0) : uint128(6341958396); // 0.2 * 10**18/(365*24*60*60) = 20% per year;
     }
 
     function _getUserInterest(
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.24;
+import "forge-std/Test.sol";
 // Interfaces
 import {PanopticPool} from "./PanopticPool.sol";
 // Inherited implementations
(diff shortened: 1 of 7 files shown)


guil-lambert:
@coderabbitai please review/provide feedback

guil-lambert:
> The author of this PR is on the CodeRabbit Free Plan. In order to use the Chat feature, please upgrade the PR author to CodeRabbit Pro by visiting https://app.coderabbit.ai/login.

@coderabbitai We should be on the PRO plan now.

contracts/CollateralTracker.sol
```diff
@@ -639,6 +689,113 @@
...
+                    uint256 shares = Math.mulDivRoundingUp(
+                        userInterestOwed,
+                        totalSupply,
+                        s_poolAssets + inAMM + unrealizedGlobalInterest
+                    );
+                    if (shares > 0) _burn(owner, shares);
+
+                    unrealizedGlobalInterest = userInterestOwed > unrealizedGlobalInterest
+                        ? 0
+                        : unrealizedGlobalInterest - userInterestOwed;
```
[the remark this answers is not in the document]
guil-lambert: This netting to 0 happens because the amount borrowed is netted at a protocol level as `net = (amountBorrowedForSelling - amountReturnedForBuying)`

`net` may be smaller than the total `amountBorrowedForSelling`. But the interest rate paid by the borrower will be based on their own contribution to `amountBorrowedForSelling`, whereas options buyers do not receive a "negative interest rate"

Example:

1. Alice borrows 1ETH to sell a call
2. Bob buys 0.5 ETH of that call
3. There is 2ETH supplied to the protocol, so the protocol utilization is (1 - 0.5)/2 = 25% and let's say the interest rate is 5%
4. The interest rate is based on 0.5ETH of net capital borrowed, and so the protocol expects the `unrealizedGlobalInterest` to collect 5% on that 0.5ETH per year = 0.025ETH
5. But Alice pays 5% on her 1ETH borrowed, so she pays 0.05ETH per year.
6. The `unrealizedGlobalInterest` would be 0.025ETH before Alice closes her position at the end of the year, but she'll pay 0.05ETH instead. That would underflow `unrealizedGlobalInterest`, which is why we net it to `0`

test/foundry/core/CollateralTracker.t.sol
```diff
@@ -578,6 +590,610 @@
...
+        );
+
+        console2.log(
+            "cc",
+            collateralToken0.totalAssets(),
+            collateralToken0.previewRedeem(collateralToken0.balanceOf(Alice)),
+            collateralToken0.previewRedeem(collateralToken0.balanceOf(Bob))
+        );
+        vm.stopPrank();
+    }
```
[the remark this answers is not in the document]
guil-lambert: Please do not review the test files until we tell you to @coderabbitai

test/foundry/core/CollateralTracker.t.sol
```diff
@@ -578,6 +590,610 @@
...
+            Constants.MIN_V3POOL_TICK
+        );
+        console2.log(
+            "cc",
+            collateralToken0.totalAssets(),
+            collateralToken0.previewRedeem(collateralToken0.balanceOf(Alice)),
+            collateralToken0.previewRedeem(collateralToken0.balanceOf(Bob))
+        );
+        vm.stopPrank();
+    }
```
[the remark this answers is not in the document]
guil-lambert: Please do not review the test files until we tell you to https://github.com/coderabbitai

contracts/CollateralTracker.sol
```diff
@@ -1077,8 +1251,13 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
             // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
             s_poolAssets = uint256(updatedAssets + realizedPremium).toUint128();
-            s_inAMM = uint256(int256(uint256(s_inAMM)) - (shortAmount - longAmount)).toUint128();
 
+            // update the inAMM value, removing the part that was paid as interest (CHECK Math)
+            {
+                int128 netBorrows = shortAmount - longAmount;
+                s_inAMM = uint256(int256(uint256(s_inAMM)) - netBorrows).toUint128();
+                s_interestState[optionOwner] = s_interestState[optionOwner].toLeftSlot(netBorrows);
+            }
```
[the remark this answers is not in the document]
guil-lambert: I believe `agg` can never be larger than `type(int128).max` because it comes from quantities that themselves cannot be larger than `type(int128).max`

test/foundry/core/CollateralTracker.t.sol
```diff
@@ -578,6 +590,675 @@ contract CollateralTrackerTest is Test, PositionUtils {
...
+        );
+        uint256 unrealizedGlobalInterest = Math.mulDivWadRoundingUp(
+            collateralToken0._inAMM(),
+            interestForPeriod
+        );
+
+        uint256 expectedAccumulator = (unrealizedGlobalInterest << 128) +
+            ((startingTime + 12) << 96) +
+            expectedNewIndex;
+
```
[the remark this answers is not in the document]
guil-lambert: Please do not review the test files until we tell you to @coderabbitai

test/foundry/core/CollateralTracker.t.sol
```diff
@@ -578,6 +590,675 @@ contract CollateralTrackerTest is Test, PositionUtils {
...
+        vm.warp(2 ** 32 - 10);
+        _initWorld(0);
+        uint256 startingTime = block.timestamp;
+        uint128 initialBorrowIndex = 1e18;
+        collateralToken0.setPoolAssets(500);
+        collateralToken0.setInAMM(500);
+
+        vm.warp(2 ** 32 + 10);
+
+        collateralToken0.accrueInterest();
```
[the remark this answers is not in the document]
guil-lambert: Please do not review the test files until we tell you to https://github.com/coderabbitai

test/foundry/core/CollateralTracker.t.sol
```diff
@@ -578,6 +590,675 @@ contract CollateralTrackerTest is Test, PositionUtils {
...
+            expectedNewIndex;
+        uint256 actualAccumulator = collateralToken0._interestRateAccumulator();
+
+        // 4. Assert the final state is correct.
+        assertEq(
+            actualAccumulator,
+            expectedAccumulator,
+            "FAIL: Interest did not accrue correctly after borrow was made"
+        );
+    }
```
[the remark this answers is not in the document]
guil-lambert: Please do not review the test files until we tell you to https://github.com/coderabbitai

contracts/CollateralTracker.sol
```diff
@@ -1029,8 +1212,15 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // however, any intrinsic value is paid for by the users, so we only add the portion that comes from PLPs: the short/long amounts
             // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
             s_poolAssets = uint256(updatedAssets).toUint128();
-            s_inAMM = uint256(int256(uint256(s_inAMM)) + (shortAmount - longAmount)).toUint128();
+            int128 netBorrows = shortAmount - longAmount;
+            s_inAMM = uint256(int256(uint256(s_inAMM)) + netBorrows).toUint128();
```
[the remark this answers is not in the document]
guil-lambert: The way `toLeftSlot` works is that is adds it to the existing value, so `toLeftSlot(x)` results into `previous + x`

contracts/CollateralTracker.sol
```diff
@@ -639,6 +688,164 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+            if (deltaTime > 0) {
+                // Calculate interest growth
+                uint128 rawInterest = uint128(
+                    Math.wTaylorCompounded(interestRate(), uint128(deltaTime))
+                );
+                // Calculate interest owed on borrowed amount
+                uint128 inAMM = s_inAMM;
+                uint128 interestOwed = Math.mulDivWadRoundingUp(inAMM, rawInterest).toUint128();
+                unrealizedGlobalInterest += interestOwed;
+
```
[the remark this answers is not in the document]
guil-lambert: Yes, fixed!

contracts/CollateralTracker.sol
```diff
@@ -120,6 +120,16 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @notice The fee of the Uniswap pool in hundredths of basis points.
     uint24 internal s_poolFee;
 
+    /// @notice Interest rate accumulator, contains the net interest owed in the left slot, the last block timestamp in the upper 32 bits of the right slot and the accumulator itself in the lower 96 bits in Wad of the right slot.
```
flanagansteve: I would be wary of using a timestamp here - timestamps can be manipulated easily by colluding with block producers / MEV searchers: https://medium.com/coinmonks/smart-contract-security-block-timestamp-manipulation-baec1b95c921

Alternatives include:
- Using a timestamp oracle
- Storing the block number, and either redefining your interest variables to be per-block rather than per-second (or per-year), or doing time calculations using a static SECONDS_PER_BLOCK (12s for mainnet, 2s for base, 1s for unichain - should be updateable to handle chain-level upgrades (Base is going from 2 to 1 soon, i think))
  - This will cause you to show less time passing than in reality if a reorg or chain downtime happens, but I think that's fine - interest failing to accumulate during that period would actually arguably be _preferred_ behaviour
  - Missed slots would also cause some drift here, which does happen on mainnet - real average blocktime is 12.05 seconds
    - A cleaner approach for mainnet would be using the slot number: https://www.blocknative.com/blog/anatomy-of-a-slot, since you can get an exact time from that and there's a fixed number of them. However, getting the slot value into a smart contract requires proving the CL root and proof too ([EIP-4788](https://eips.ethereum.org/EIPS/eip-4788)), so probably overkill for our usecase. [EIP-7843](https://eips.ethereum.org/EIPS/eip-7843) will let us get the slot num directly, so we should keep an eye on that, but just hardcoding 12s (or 12.05s) for now should be fine IMO

contracts/CollateralTracker.sol
```diff
@@ -639,6 +688,163 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+                        totalSupply,
+                        _totalAssets
+                    );
+
+                    uint128 burntInterestValue = userInterestOwed;
+                    if (shares > 0) {
+                        address _owner = owner;
+                        uint256 userBalance = balanceOf[_owner];
+                        if (shares > userBalance) {
+                            // update the acrual of interest paid
```
flanagansteve:
```suggestion
                            // update the accrual of interest paid
```

contracts/CollateralTracker.sol
```diff
@@ -639,6 +688,163 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+                    .mulDivWadRoundingUp(currentBorrowIndex, borrowIndex)
+                    .toUint128();
+            }
+        }
+    }
+
+    /// @notice returns the interest rate per second
+    function interestRate() public view returns (uint128) {
+        uint256 utilization = _poolUtilization();
+        return utilization == 0 ? uint128(1) : uint128(6341958396); // 0.2 * 10**18/(365*24*60*60) = 20% per year;
```
flanagansteve, line 844: Hmmm, so can I ever get `((s_inAMM + unrealizedGlobalInterest) * DECIMALS) / totalAssets()` to round down to 0 (maybe via flashloaning/looped calls) to get crazy low interest, perhaps then enabling me to borrow cheaply and pay off my flash loans

contracts/CollateralTracker.sol
```diff
@@ -639,6 +688,163 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+    }
+
+    function _getUserInterest(
```
flanagansteve, line 852: You probably already know this, but making a note that these and below need docstrings

contracts/CollateralTracker.sol
```diff
@@ -639,6 +688,163 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+        int128 netBorrows = userState.leftSlot();
+        uint128 userBorrowIndex = uint128(userState.rightSlot());
+        if (netBorrows <= 0 || userBorrowIndex == 0 || currentBorrowIndex == userBorrowIndex) {
+            return 0;
+        }
+        unchecked {
+            interestOwed = Math
+                .mulDivRoundingUp(
+                    uint128(netBorrows),
+                    currentBorrowIndex - userBorrowIndex,
```
flanagansteve, line 865: This line is what made much of the above finally click for me - i had been wondering what an "index" meant, and so on. We should improve the language here - should be verbose and specific. Maybe we introduce a concept of "borrowingEpochs" or "loanEpochs", and therefore the rightSlot of a userState is the epoch their borrowing began during? 

We could then store this value in a mapping from epoch indices:

```solidity
// Update borrow index
                uint128 borrowIndex = 10 ** 18 + rawInterest;
                currentBorrowIndex = Math
                    .mulDivWadRoundingUp(currentBorrowIndex, borrowIndex)
                    .toUint128();
```

And call it something like `interestAccruedToThisPoint`, or something slightly more concise but equally precise

contracts/CollateralTracker.sol
```diff
@@ -1145,7 +1370,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                     .toLeftSlot(
                         positionBalanceArray.length > 0
                             ? (_getTotalRequiredCollateral(atTick, positionBalanceArray) +
-                                longPremium).toUint128()
+                                longPremium +
+                                _owedInterest(user)).toUint128()
```
flanagansteve, line 1430: Note to self: Audit the other PanopticPool and CollateralTracker methods to see if there are other methods that contain a concept of "what does this user owe Panoptic", and ensure owedInterest is included

contracts/CollateralTracker.sol
```diff
@@ -1050,6 +1269,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         int128 swappedAmount,
         int128 realizedPremium
     ) external onlyPanopticPool returns (int128) {
```
flanagansteve, line 1326: note to self: see if any other methods affect a user's collateral balance and whether they need to also accrueInterest

contracts/CollateralTracker.sol
```diff
@@ -1029,8 +1241,15 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // however, any intrinsic value is paid for by the users, so we only add the portion that comes from PLPs: the short/long amounts
             // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
             s_poolAssets = uint256(updatedAssets).toUint128();
-            s_inAMM = uint256(int256(uint256(s_inAMM)) + (shortAmount - longAmount)).toUint128();
+            int128 netBorrows = shortAmount - longAmount;
+            s_inAMM = uint256(int256(uint256(s_inAMM)) + netBorrows).toUint128();
 
+            {
+                address _optionOwner = optionOwner;
+                s_interestState[_optionOwner] = s_interestState[_optionOwner].toLeftSlot(
```
flanagansteve, line 1304: note to self: double check that toLeftSlot here is _additive_, not an overwrite. 

I'm pretty sure it is - if so, we should rename `toLeftSlot` to reflect that it is additive, as it currently reads like "store this value in the left slot, regardless of what's currently there"

contracts/libraries/Math.sol
```diff
@@ -1082,4 +1166,18 @@ library Math {
         }
         return data;
     }
+
+    /*//////////////////////////////////////////////////////////////
+                         EXPONENTIAL MATH
+    //////////////////////////////////////////////////////////////*/
+
+    /// @dev Returns the sum of the first three non-zero terms of a Taylor expansion of e^(nx) - 1, to approximate a
```
flanagansteve, line 1174: What precision does 3 terms get us? Is it enough for our usecase? Would be good to document here

contracts/CollateralTracker.sol
```diff
@@ -261,6 +276,31 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         currentPoolUtilization = _poolUtilization();
     }
 
+    /// @notice Returns current borrowIndex.
+    /// @return The borrowIndex
+    function borrowIndex() external view returns (uint96) {
+        return uint96(s_interestRateAccumulator.rightSlot());
+    }
+
+    /// @notice Returns the last block at which interest rates were compounded.
```
flanagansteve:
```suggestion
    /// @notice Returns the last time at which interest rates were compounded.
```

would be correct for current impl; comment above might prompt us to change this though

contracts/CollateralTracker.sol
```diff
@@ -261,6 +276,31 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+    /// @return The last block at which the interest rates were compounded
```
flanagansteve:
```suggestion
    /// @return The last time at which the interest rates were compounded
```

(similarly, may change back to block)

contracts/CollateralTracker.sol
```diff
@@ -261,6 +276,31 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+    function lastInteractionTimestamp() external view returns (uint256) {
+        return (s_interestRateAccumulator.rightSlot() >> 96);
+    }
+
+    /// @notice Returns the last block at which interest rates were compounded.
```
flanagansteve: incorrect docstring, copypastaed from `lastInteractionTimestamp`

contracts/CollateralTracker.sol
```diff
@@ -261,6 +276,31 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+    /// @return The last block at which the interest rates were compounded
+    function unrealizedGlobalInterest() external view returns (uint256) {
+        return s_interestRateAccumulator.leftSlot();
+    }
+
+    /// @notice Returns the baseIndex and the net borrows for the user
```
flanagansteve:
```suggestion
    /// @notice Returns the epoch at which this user's interest accumulation began, and the net borrowed amount for the user
```

contracts/CollateralTracker.sol
```diff
@@ -261,6 +276,31 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+    /// @return The base index of the user in the current state
```
flanagansteve:
```suggestion
    /// @return The epoch this user's borrowing began in
```

contracts/CollateralTracker.sol
```diff
@@ -261,6 +276,31 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+    /// @return The net borrowing amounts for the user (if negative, do not pay interest)
```
flanagansteve:
```suggestion
    /// @return The net borrowed amount for the user (if negative, they are owed money on net, and accumulate interest rather than paying it)
```

contracts/CollateralTracker.sol
```diff
@@ -639,6 +688,163 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+                uint128 unrealizedGlobalInterest,
+                uint256 currentTime,
+                uint128 deltaTime
+            ) = _calculateCurrentInterestState(inAMM);
+
+            // USER
+            LeftRightSigned userState = s_interestState[owner];
+            int128 netBorrows = userState.leftSlot();
+            int128 userBorrowIndex = int128(currentBorrowIndex);
+            if (netBorrows != 0) {
```
flanagansteve:
```suggestion
            if (netBorrows < 0) {
```

Is this possibly what you meant? I don't think you need to do anything inside this `if` branch for users that are net-creditors

contracts/CollateralTracker.sol
```diff
@@ -639,6 +688,163 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+            (
+                uint128 currentBorrowIndex,
+                uint128 unrealizedGlobalInterest,
+                uint256 currentTime,
+                uint128 deltaTime
+            ) = _calculateCurrentInterestState(inAMM);
+
+            // USER
+            LeftRightSigned userState = s_interestState[owner];
+            int128 netBorrows = userState.leftSlot();
```
flanagansteve, line 743: You actually don't need this netBorrows variable more than once - instead, you could make the `if` below `if (userState.leftSlot() < 0)` (or != 0 if that's what you meant). And then when you do your final effects after the `if` instead of:

```solidity
s_interestState[owner] = LeftRightSigned
                .wrap(0)
                .toRightSlot(userBorrowIndex)
                .toLeftSlot(netBorrows);
```

do

```
s_interestState[owner] = userState
                .toRightSlot(userBorrowIndex);
```

If it is cheaper to make the shallow copy of netBorrows and use it in the `if`, you'll still save an SSTORE (from 0, nonetheless!) by taking the latter recommendation.

flanagansteve: > s_interestState[owner] = userState
>                .toRightSlot(userBorrowIndex);

Agh wait - If .toRightSlot is additive rather than an overwrite, you'll have to do this:

```solidity
s_interestState[owner] = userState
                .toRightSlot(userBorrowIndex - userState.rightSlot());
```

e.g., increase its rightSlot by the correct amount to make its value the end-of-function `userBorrowIndex` value. but that should still be cheaper i think

contracts/CollateralTracker.sol
```diff
@@ -639,6 +688,163 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+                        : unrealizedGlobalInterest - burntInterestValue;
+                }
+            }
+
+            s_interestState[owner] = LeftRightSigned
+                .wrap(0)
+                .toRightSlot(userBorrowIndex)
+                .toLeftSlot(netBorrows);
+
+            s_interestRateAccumulator = LeftRightUnsigned
```
flanagansteve, line 788: Similarly, i think it may be cheaper to modify the existing s_interestRateAccumulator value rather than do an overwrite with a newly initialised LeftRight.wrap(0) - e.g., if .toLeftSlot / .toRightSlot are additive:

```solidity
s_interestRateAccumulator
    .toLeftSlot(unrealizedGlobalInterest - s_interestRateAccumulator.leftSlot())
    .toRightSlot(uint128((currentTime << 96) + uint96(currentBorrowIndex)) - s_interestRateAccumulator.rightSlot())
```

Worth comparing forge outputs of the gas usage in tests of this method to see which is cheaper - claude can whip up a simple benchmark

contracts/CollateralTracker.sol
```diff
@@ -346,7 +390,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @return The total amount of assets managed by the CollateralTracker vault
     function totalAssets() public view returns (uint256) {
```
flanagansteve, line 418: Note to self: Its very important that people who are owed interest have that reflected in their collateral balance somehow. I am pretty sure that the `unrealizedGlobalInterest` being included in totalAssets accomplishes that. And that the amounts they have borrowed is correctly counted against them by including that in the getAccountMarginDetails.

But should double check this

contracts/CollateralTracker.sol
```diff
@@ -639,6 +688,163 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         emit Withdraw(msg.sender, receiver, owner, assets, shares);
     }
 
+    /// @notice Accrues protocol-wide interest.
+    function accrueInterest() public {
```
flanagansteve, line 719: The dynamic in the current implementation is that you can choose to voluntarily pay interest at any time using this method (and also stamp your netBorrows), but if someone else wants to force you to pay interest (or update your netBorrows), it must be part of some other action.

One of those actions is transferFrom, so a hacky way for me to do that is to transfer a trivial amount of my CT shares to you.

I think it probably makes more sense to let anybody accrueInterest(address target) somebody else at any time.

contracts/CollateralTracker.sol
```diff
@@ -1077,8 +1297,13 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             // any intrinsic value is paid for by the users, so we do not add it to s_inAMM
             // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
             s_poolAssets = uint256(updatedAssets + realizedPremium).toUint128();
-            s_inAMM = uint256(int256(uint256(s_inAMM)) - (shortAmount - longAmount)).toUint128();
 
+            // update the inAMM value, removing the part that was paid as interest (CHECK Math)
+            {
+                int128 netBorrows = shortAmount - longAmount;
```
flanagansteve: This is a very vague thought but: is this system is too generous to owners of short positiosn? Arguably, you're kind of getting double credit - you're receiving collateral shares as credit for your short positions, _and_ the shortAmount helps decrease the netBorrows you are tracked as borrowing. Maybe you should be charged interest on the longAmounts regardless of countervailing shortAmounts?

contracts/CollateralTracker.sol
guil-lambert: My understanding is that the timestamps can only be manipulated within that 12s block time: the new timestamp must be larger than the previous one but not more than 12s away.

In the context of interest rate accumulation, that 12s of play doesn't really have any impact.

- Euler uses `block.timestmap` when accruing interest: https://github.com/euler-xyz/euler-vault-kit/blob/5b98b42048ba11ae82fb62dfec06d1010c8e41e6/src/Synths/EulerSavingsRate.sol#L194
- Morpho also uses `block.tmestamp`: https://medium.com/coinmonks/smart-contract-security-block-timestamp-manipulation-baec1b95c921

Using block_to_timstamp conversion would also break whenif Unichain or Base decide to decrease their block times, which they have done already in the past.

contracts/CollateralTracker.sol
guil-lambert: Changed

contracts/CollateralTracker.sol
guil-lambert: Not sure what you mean here by `epoch`. This slot tracks the snapshot of the borrow index accumulator stored in `s_interestRateAccumulator.rightSlot()`, which is updated at every poke according to:

```
currentBorrowIndex = Math
    .mulDivWadRoundingUp(currentBorrowIndex, 10 ** 18 + rawInterest)
    .toUint128();
```

contracts/CollateralTracker.sol
guil-lambert: > they are owed money on net, and accumulate interest rather than paying it

This is not quite accurate. Negative borrows (ie. buying options) does indeed result in no interest being paid instead of being owed money

contracts/CollateralTracker.sol
guil-lambert: Yes, here's where it's done, line 1374:

https://github.com/panoptic-labs/panoptic-next-core-private/blob/7571cd0f2e61377fd07015ec23013eb68c3e4918/contracts/CollateralTracker.sol#L1368-L1376

The owed interest (interest they should be paying) is added to the left slot, which tracks collateral requirements

contracts/CollateralTracker.sol
guil-lambert: Ok, that could be added. As you said, the transferFrom path already allows other anyone to force you to pay interest, so might as well make it more straightforward

contracts/CollateralTracker.sol
guil-lambert: I'm not sure if that'll be cheaper: the `s_interestState` needs to be updated regardless to write the new `userBorrowIndex`, so whether the full 256 bits are updated or just the lower 128bits, the cost should be the same.

And `int128 netBorrows = userState.leftSlot();` being called twice (line 711 and line 811) might be redundant, but it's only a bitshifting operation, so not much to optimize besides that `3` gas

contracts/CollateralTracker.sol
flanagansteve: > Using block_to_timstamp conversion would also break whenif Unichain or Base decide to decrease their block times, which they have done already in the past.

Yes, I noted this. The seconds-per-block value could be updateable, though I know this introduces an admin which isn't ideal. It could be a tightly-bound admin function, with restrictions such as only allowing the SECONDS_PER_BLOCK to be altered 1 second at a time, so that a leaked admin key couldn't be used to rapidly alter the value. Could go farther and only allow reduction of SECONDS_PER_BLOCK, since realistically that's the only direction it would ever make sense for it to move.

However, it's all a moot point:

> My understanding is that the timestamps can only be manipulated within that 12s block time: the new timestamp must be larger than the previous one but not more than 12s away.

RE Mainnet:

My age is showing - I learned not to trust block.timestamp back in the PoW days, but actually, block.timestamp has _great_ guarantees now - it shouldn't even vary by 12s, it should be precisely the same as my proposed slot-based calculation above.

Source in lodestar (the software validators are using to validate blocks): https://github.com/ChainSafe/lodestar/blob/248b252a94a7abcde2ea85cd4b204dede70fec29/packages/state-transition/src/block/processExecutionPayload.ts#L41

Source in eth2Book, with commentary: https://eth2book.info/capella/part3/transition/block/#process_execution_payload

RE Base/Unichain:

_Technically_ we're at the mercy of the central sequencer for an accurate timestamp, with effectively no guarantees, and you could argue that a timestamp bug is more likely than a block number bug. But I think that's a risk I'm willing to take.

contracts/CollateralTracker.sol
flanagansteve: Looks like you added it. Should we consider removing the argument-less `accrueInterest()`, then, since someone could just pass their own address into the `accrueInterest(address from)`? Or do you like that there's a slightly cheaper one people can use to accrue interest on themselves?

contracts/CollateralTracker.sol
flanagansteve: I was referring to my proposal [above](https://github.com/panoptic-labs/panoptic-next-core-private/pull/159#discussion_r2383377518):

```solidity
Maybe we introduce a concept of "borrowingEpochs" or "loanEpochs", and therefore the rightSlot of a userState is the epoch their borrowing began during?
```

But your additional documentation on how the borrow indexing works helps a lot, so let me re-review and reconsider whether I'd recommend a change like that anymore

contracts/CollateralTracker.sol
flanagansteve: Confirmed - although i still think we should rename this method, we even have comments in LeftRight.sol that try to get ahead of this confusion:

```solidity
  /// All toLeftSlot functions add bits to the left slot without clearing it first
    // Typically, the slot is already clear when writing to it, but if it is not, the bits will be added to the existing bits
    // Therefore, the assumption must not be made that the bits will be cleared while using these helpers
    // Note that the values *within* the slots are allowed to overflow, but overflows are contained and will not leak into the other slot
```

So a better method name, even just `addToLeftSlot` / `addToRightSlot`, would be great for improving our clarity (and therefore our security - auditors get slowed down by misleading names like this one)

Doesn't have to be in this PR, but let's record the task somewhere if you're in alignment that we should make this change (recorded in either your spreadsheet or the kanban)

contracts/CollateralTracker.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.24;
-
+import "forge-std/Test.sol";
```
flanagansteve, line 3:
```suggestion

```

Reminder to remove this before merging into a `dev` branch

contracts/CollateralTracker.sol
```diff
@@ -346,7 +419,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @return The total amount of assets managed by the CollateralTracker vault
     function totalAssets() public view returns (uint256) {
         unchecked {
-            return uint256(s_poolAssets) + s_inAMM;
+            uint128 unrealizedGlobalInterest = s_interestRateAccumulator.leftSlot();
```
flanagansteve, line 420: These two lines could be consolidated to:

```solidity
            return uint256(s_poolAssets) + s_inAMM + s_interestRateAccumulator.leftSlot();
```

Which should be cheaper gas-wise I believe. I know `totalAssets` is `view`, but it's also by many non-view methods, so the impact of dropping an `MSTORE` would be pretty noticeable IMO.

contracts/CollateralTracker.sol
```diff
@@ -392,8 +466,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param receiver User to receive the shares
     /// @return shares The amount of Panoptic pool shares that were minted to the recipient
     function deposit(uint256 assets, address receiver) external returns (uint256 shares) {
+        _accrueInterest(receiver);
         if (assets > type(uint104).max) revert Errors.DepositTooLarge();
-
         shares = previewDeposit(assets);
```
flanagansteve, line 469:
```suggestion

        shares = previewDeposit(assets);
```

Nit: Generally, it's better style to leave a newline following a guard clause, IMO

contracts/CollateralTracker.sol
```diff
@@ -505,6 +580,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         address receiver,
         address owner
     ) external returns (uint256 shares) {
+        _accrueInterest(owner);
```
flanagansteve, line 581: Should we also `_accrueInterest(receiver)`?

I think probably not - it's analogous to the `msg.sender` of `deposit()`, too - that party isn't actually doing anything with their collateral here. But just checking

contracts/CollateralTracker.sol
flanagansteve: > the s_interestState needs to be updated regardless to write the new userBorrowIndex, so whether the full 256 bits are updated or just the lower 128bits, the cost should be the same.

It depends on the order of operations that the `solc` compiler pursues. The way I read the code, it looks to me like you create a new LeftRightSigned (probably via MSTORE), then write its lower 128 bits, then its upper 128 bits.

What I'm suggesting would mean skipping that temp variable and just overwriting 128 bits of a value that you already have to MLOAD & MSTORE to anyways

> And int128 netBorrows = userState.leftSlot(); being called twice (line 711 and line 811) might be redundant, but it's only a bitshifting operation, so not much to optimize besides that 3 gas

It's also an MSTORE for the netBorrows variable itself, which is the main thing I'm trying to point out - I have had an eye out to chop off memory vars that aren't necessary

contracts/CollateralTracker.sol
```diff
@@ -746,7 +1013,8 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @return poolUtilization The pool utilization in basis points
     function _poolUtilization() internal view returns (uint256 poolUtilization) {
         unchecked {
-            return (s_inAMM * DECIMALS) / totalAssets();
+            uint128 unrealizedGlobalInterest = s_interestRateAccumulator.leftSlot();
```
flanagansteve, line 1014: Same comment here about shifting unrealizedGlobalInterest to be used in-line instead of as its own variable:

```solidity
return ((s_inAMM + s_interestRateAccumulator.leftSlot()) * DECIMALS) / totalAssets();
```

contracts/PanopticPool.sol
```diff
@@ -889,7 +889,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
         (LeftRightSigned longAmounts, LeftRightSigned shortAmounts) = PanopticMath
             .computeExercisedAmounts(tokenId, positionSize);
-
         {
```
flanagansteve, line 892: nit: would restore this newline
```suggestion

        {
```

contracts/PanopticPool.sol
```diff
@@ -1566,7 +1565,6 @@ contract PanopticPool is ERC1155Holder, Multicall {
             // deduct the paid premium tokens from the owner's balance and add them to the cumulative settled token delta
             s_collateralToken0.exercise(owner, 0, 0, 0, -realizedPremia.rightSlot());
             s_collateralToken1.exercise(owner, 0, 0, 0, -realizedPremia.leftSlot());
-
             bytes32 chunkKey = keccak256(
```
flanagansteve, line 1568: nit: would restore this newline

```suggestion

            bytes32 chunkKey = keccak256(
```

contracts/CollateralTracker.sol
flanagansteve: New docstrings resolves this concern

contracts/CollateralTracker.sol
```diff
@@ -639,6 +717,195 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+    /// @return The amount of interest that would be owed if accrued at current timestamp
+    function previewOwedInterest(address owner) public view returns (uint128) {
+        uint256 simulatedBorrowIndex = _calculateCurrentBorrowIndex();
+        LeftRightSigned userState = s_interestState[owner];
+        return _getUserInterest(userState, simulatedBorrowIndex);
+    }
+
     /*//////////////////////////////////////////////////////////////
                             ACCOUNTING LOGIC
     //////////////////////////////////////////////////////////////*/
```
flanagansteve, line 909: Below, there are `onlyPanopticPool` methods that affect user collateral balances - `exercise`, `takeCommissionAndAddData`, and `settleLiquidation` in straightforward manners, and then `delegate`, `revoke`, and `refund` in more edgecasey ways - should those force the user to `accrueInterest`, too? Is it OK to take from a user to settle the liquidation when their interest owed is outdated, for example? Or for their positions to be exercised?

contracts/CollateralTracker.sol
flanagansteve: We could also consider just making `_getUserInterest` return 0 if netBorrows is <= 0, which would eliminate your need to even have a netBorrows if branch (saving gas + the MSTORE of netBorrows), and would look nicer too. I think that's my favourite path here

contracts/CollateralTracker.sol
```diff
@@ -639,6 +717,195 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+            ) = _calculateCurrentInterestState(inAMM);
+
+            // USER
+            LeftRightSigned userState = s_interestState[owner];
+            int128 netBorrows = userState.leftSlot();
+            int128 userBorrowIndex = int128(currentBorrowIndex);
+            if (netBorrows > 0) {
+                uint128 userInterestOwed = _getUserInterest(userState, currentBorrowIndex);
+                if (userInterestOwed != 0) {
+                    uint256 _totalAssets = s_poolAssets + inAMM + unrealizedGlobalInterest;
```
flanagansteve, line 748: Is the reason to redo this calc here gas savings? Would prefer to reuse the existing totalAssets method here to avoid drift between the two implemenations

contracts/CollateralTracker.sol
```diff
@@ -639,6 +717,195 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+
+                    uint256 shares = Math.mulDivRoundingUp(
+                        userInterestOwed,
+                        totalSupply,
+                        _totalAssets
+                    );
+
+                    uint128 burntInterestValue = userInterestOwed;
```
flanagansteve, line 756: You don't use `userInterestOwed` after this - you can just overwrite its value instead of introducing the new `burntInterestValue` variable

contracts/CollateralTracker.sol
```diff
@@ -639,6 +717,195 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+            // USER
+            LeftRightSigned userState = s_interestState[owner];
+            int128 netBorrows = userState.leftSlot();
+            int128 userBorrowIndex = int128(currentBorrowIndex);
+            if (netBorrows > 0) {
+                uint128 userInterestOwed = _getUserInterest(userState, currentBorrowIndex);
+                if (userInterestOwed != 0) {
+                    uint256 _totalAssets = s_poolAssets + inAMM + unrealizedGlobalInterest;
+
+                    uint256 shares = Math.mulDivRoundingUp(
```
flanagansteve, line 750:
```suggestion
                    uint256 sharesOwed = Math.mulDivRoundingUp(
```

How about a more specific varname than `shares` here? Could even go for like `sharesOwedToPayInterest`. I'm trying to bring more specificity into our codebase in general, to help auditors

contracts/CollateralTracker.sol
```diff
@@ -639,6 +717,195 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+                        userInterestOwed,
+                        totalSupply,
+                        _totalAssets
+                    );
+
+                    uint128 burntInterestValue = userInterestOwed;
+                    if (shares > 0) {
+                        address _owner = owner;
```
flanagansteve, line 758: I assume this one was just because of stack-too-deep - if you take some of the reccs above and remove the whole `if (netBorrows > 0)` branch you may not need to do this

contracts/CollateralTracker.sol
```diff
@@ -639,6 +717,195 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+                            /// @dev DO NOT update index. By keeping the user's old baseIndex, their debt continues to compound correctly from the original point in time.
+                            userBorrowIndex = userState.rightSlot();
+                        } else {
+                            // Solvent case: Pay in full.
+                            _burn(_owner, shares);
+                        }
+                    }
+
+                    unrealizedGlobalInterest = burntInterestValue > unrealizedGlobalInterest
+                        ? 0
```
flanagansteve, line 778: Why does this go to 0 if the user's interest owed exceeds the global unrealised interest? Couldn't they be insolvent, for example, and still have unrealised interest that is busy compounding away? I think I'm misunderstanding something here

flanagansteve: Also, under what circumstance _can_ the user owe more than unrealizedGlobalInterest?

contracts/CollateralTracker.sol
```diff
@@ -639,6 +717,195 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+                        address _owner = owner;
+                        uint256 userBalance = balanceOf[_owner];
+                        if (shares > userBalance) {
+                            // update the accrual of interest paid
+                            burntInterestValue = Math
+                                .mulDiv(userBalance, _totalAssets, totalSupply)
+                                .toUint128();
+
+                            /// Insolvent case: Pay what you can
+                            _burn(_owner, userBalance);
```
flanagansteve, line 767: Is this method a good candidate to guard with a minimum userBalance burnt, like your DevDacian PR implemented? Seems like there could be nasty bugs where an insolvent user continually forces this path of accrueInterest, which does affect global vars like `unrealizedGlobalInterest` (flooring it to 0, for example, if the user owed more than the unrealizedGlobalInterest)

contracts/CollateralTracker.sol
```diff
@@ -639,6 +717,195 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
+    /// @return currentTime Current block timestamp
+    /// @return deltaTime Seconds elapsed since last interest accrual
+    function _calculateCurrentInterestState(
+        uint128 inAMM
+    )
+        internal
+        view
+        returns (
+            uint128 currentBorrowIndex,
+            uint128 unrealizedGlobalInterest,
```
flanagansteve, line 809: You actually only use the latter 3 return values here in _accrueInterest - all other callers of this method only care about the `currentBorrowIndex`.

You could instead just return currentBorrowIndex and interestOwed, and then let _accrueInterest:

- calculate unrealizedGlobalInterest as s_interestRateAccumulator.leftSlot() + interestOwed
- currentTime as block.timestamp
- deltaTime as currentTime - s_interestRateAccumulator.rightSlot() >> 96

Which would save all 4 callers gas, I believe (by removing returnval variables)

PR #160, opened 2025-09-12
Fix0: long premia only at mint + virtual shares in settleLongPremium

guil-lambert:
This PR added the same features that were implemented for v1.1's launch:

- feat: add virtual share delegation to `settleLongPremium` https://github.com/panoptic-labs/panoptic-v1-core/pull/25
- feat: allow users to skip earned premium calculation https://github.com/panoptic-labs/panoptic-v1-core/pull/26

@coderabbitai: do not review this PR

## Summary by CodeRabbit

- **New Features**
  - Added premia-as-collateral mode across mint, burn, withdraw, settle and force-exercise flows (new trailing flag) and ERC‑1155 receipt support.

- **Refactor**
  - Unified premia-as-collateral computations and collateral-aware solvency checks; moved pricing to oracle-driven ticks and renamed stale-price error to StaleOracle; refund/substitution logic refined.

- **Tests**
  - Updated tests for new APIs and oracle flows; added token-substitution settlement test.

- **Chores**
  - Adjusted CI optimizer settings.

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -6,7 +6,6 @@ import {CollateralTracker} from "@contracts/CollateralTracker.sol";
 import {SemiFungiblePositionManager} from "@contracts/SemiFungiblePositionManager.sol";
 import {IUniswapV3Pool} from "univ3-core/interfaces/IUniswapV3Pool.sol";
 // Inherited implementations
-import {ERC1155Holder} from "@openzeppelin/contracts/token/ERC1155/utils/ERC1155Holder.sol";
 import {Multicall} from "@base/Multicall.sol";
 // Libraries
 import {Constants} from "@libraries/Constants.sol";
@@ -23,7 +22,7 @@ import {TokenId} from "@types/TokenId.sol";
 /// @title The Panoptic Pool: Create permissionless options on a CLAMM.
 /// @author Axicon Labs Limited
 /// @notice Manages positions, collateral, liquidations and forced exercises.
-contract PanopticPool is ERC1155Holder, Multicall {
+contract PanopticPool is Multicall {
     /*//////////////////////////////////////////////////////////////
                                 EVENTS
     //////////////////////////////////////////////////////////////*/
@@ -98,7 +97,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     int24 internal constant MAX_SWAP_TICK = Constants.MAX_V3POOL_TICK + 1;
 
     /// @notice Flag that signals to compute premia for both the short and long legs of a position.
-    bool internal constant COMPUTE_ALL_PREMIA = true;
+    bool internal constant COMPUTE_PREMIA_AS_COLLATERAL = true;
 
     /// @notice Flag that indicates only to include the share of (settled) premium that is available to collect when calling `_calculateAccumulatedPremia`.
     bool internal constant ONLY_AVAILABLE_PREMIUM = false;
@@ -292,6 +291,24 @@ contract PanopticPool is ERC1155Holder, Multicall {
     }
 
     /*//////////////////////////////////////////////////////////////
+                              EIP SUPPORT
+    //////////////////////////////////////////////////////////////*/
+
+    // note: this contract does not need to accept batch ERC1155 transfers from the SFPM or supply ERC-165 calls
+    // thus, `supportsInterface` and `onERC1155BatchReceived` are left unimplemented to reduce contract size
+
+    /// @notice Returns magic value when called by the `SemiFungiblePositionManager` contract to indicate that this contract supports ERC1155.
+    function onERC1155Received(
+        address,
+        address,
+        uint256,
+        uint256,
+        bytes memory
+    ) external pure returns (bytes4) {
+        return this.onERC1155Received.selector;
+    }
+
+    /*//////////////////////////////////////////////////////////////
                              QUERY HELPERS
     //////////////////////////////////////////////////////////////*/
 
@@ -314,11 +331,13 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @dev Reverts if account is not solvent with `BP_DECREASE_BUFFER`.
     /// @param user The account to check for collateral withdrawal eligibility
     /// @param positionIdList The list of all option positions held by `user`
+    /// @param usePremiaAsCollateral Whether to compute accumulated premia for all legs held by the user for collateral (true), or just owed premia for long legs (false)
     function validateCollateralWithdrawable(
         address user,
-        TokenId[] calldata positionIdList
+        TokenId[] calldata positionIdList,
+        bool usePremiaAsCollateral
     ) external view {
-        _validateSolvency(user, positionIdList, BP_DECREASE_BUFFER);
+        _validateSolvency(user, positionIdList, BP_DECREASE_BUFFER, usePremiaAsCollateral);
     }
 
     /// @notice Returns the total amount of premium accumulated for a list of positions and a list containing the corresponding `PositionBalance` information for each position.
@@ -341,7 +360,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
             _calculateAccumulatedPremia(
                 user,
                 positionIdList,
-                COMPUTE_ALL_PREMIA,
+                COMPUTE_PREMIA_AS_COLLATERAL,
                 includePendingPremium,
                 currentTick
             );
@@ -350,7 +369,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @notice Calculate the accumulated premia owed from the option buyer to the option seller.
     /// @param user The holder of options
     /// @param positionIdList The list of all option positions held by user
-    /// @param computeAllPremia Whether to compute accumulated premia for all legs held by the user (true), or just owed premia for long legs (false)
+    /// @param usePremiaAsCollateral Whether to compute accumulated premia for all legs held by the user for collateral (true), or just owed premia for long legs (false)
     /// @param includePendingPremium If true, include premium that is owed to the user but has not yet settled; if false, only include premium that is available to collect
     /// @param atTick The current tick of the Uniswap pool
     /// @return shortPremium The total amount of premium owed (which may `includePendingPremium`) to the short legs in `positionIdList` (token0: right slot, token1: left slot)
@@ -359,7 +378,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
     function _calculateAccumulatedPremia(
         address user,
         TokenId[] calldata positionIdList,
-        bool computeAllPremia,
+        bool usePremiaAsCollateral,
         bool includePendingPremium,
         int24 atTick
     )
@@ -379,8 +398,10 @@ contract PanopticPool is ERC1155Holder, Multicall {
         for (uint256 k = 0; k < pLength; ) {
             TokenId tokenId = positionIdList[k];
 
-            balances[k][0] = TokenId.unwrap(tokenId);
-            balances[k][1] = PositionBalance.unwrap(s_positionBalance[c_user][tokenId]);
+            balances[k] = [
+                TokenId.unwrap(tokenId),
+                PositionBalance.unwrap(s_positionBalance[c_user][tokenId])
+            ];
 
             (
                 LeftRightSigned[4] memory premiaByLeg,
@@ -389,7 +410,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
                     tokenId,
                     LeftRightUnsigned.wrap(balances[k][1]).rightSlot(),
                     c_user,
-                    computeAllPremia,
+                    usePremiaAsCollateral,
                     atTick
                 );
 
@@ -406,16 +427,17 @@ contract PanopticPool is ERC1155Holder, Multicall {
                         );
 
                         (uint256 totalLiquidity, , ) = _getLiquidities(tokenId, leg);
-                        LeftRightUnsigned availablePremium = _getAvailablePremium(
-                            totalLiquidity,
-                            s_settledTokens[chunkKey],
-                            s_grossPremiumLast[chunkKey],
-                            LeftRightUnsigned.wrap(
-                                uint256(LeftRightSigned.unwrap(premiaByLeg[leg]))
-                            ),
-                            premiumAccumulatorsByLeg[leg]
+                        shortPremium = shortPremium.add(
+                            _getAvailablePremium(
+                                totalLiquidity,
+                                s_settledTokens[chunkKey],
+                                s_grossPremiumLast[chunkKey],
+                                LeftRightUnsigned.wrap(
+                                    uint256(LeftRightSigned.unwrap(premiaByLeg[leg]))
+                                ),
+                                premiumAccumulatorsByLeg[leg]
+                            )
                         );
-                        shortPremium = shortPremium.add(availablePremium);
                     } else {
                         shortPremium = shortPremium.add(
                             LeftRightUnsigned.wrap(
@@ -480,14 +502,16 @@ contract PanopticPool is ERC1155Holder, Multicall {
         uint128 positionSize,
         uint64 effectiveLiquidityLimitX32,
         int24 tickLimitLow,
-        int24 tickLimitHigh
+        int24 tickLimitHigh,
+        bool usePremiaAsCollateral
     ) external {
         _mintOptions(
             positionIdList,
             positionSize,
             effectiveLiquidityLimitX32,
             tickLimitLow,
-            tickLimitHigh
+            tickLimitHigh,
+            usePremiaAsCollateral
         );
     }
 
@@ -496,15 +520,22 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @param newPositionIdList The new positionIdList without the token being burnt
     /// @param tickLimitLow The lower bound of an acceptable open interval for the ending price
     /// @param tickLimitHigh The upper bound of an acceptable open interval for the ending price
+    /// @param usePremiaAsCollateral Whether to compute accumulated premia for all legs held by the user for collateral (true), or just owed premia for long legs (false)
     function burnOptions(
         TokenId tokenId,
         TokenId[] calldata newPositionIdList,
         int24 tickLimitLow,
-        int24 tickLimitHigh
+        int24 tickLimitHigh,
+        bool usePremiaAsCollateral
     ) external {
         _burnOptions(COMMIT_LONG_SETTLED, tokenId, msg.sender, tickLimitLow, tickLimitHigh);
 
-        uint256 medianData = _validateSolvency(msg.sender, newPositionIdList, NO_BUFFER);
+        uint256 medianData = _validateSolvency(
+            msg.sender,
+            newPositionIdList,
+            NO_BUFFER,
+            usePremiaAsCollateral
+        );
 
         // Update `s_miniMedian` with a new observation if the last observation is old enough (returned medianData is nonzero)
         if (medianData != 0) s_miniMedian = medianData;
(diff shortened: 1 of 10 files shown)


contracts/PanopticPool.sol
```diff
@@ -1401,31 +1497,31 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+        uint256 removedLiquidity;
         (totalLiquidity, netLiquidity, removedLiquidity) = _getLiquidities(tokenId, leg);
 
         // compute and return effective liquidity. Return if short=net=0, which is closing short position
         if (netLiquidity == 0 && removedLiquidity == 0) return totalLiquidity;
 
         uint256 effectiveLiquidityFactorX32;
         unchecked {
-            effectiveLiquidityFactorX32 = (uint256(removedLiquidity) * 2 ** 32) / netLiquidity;
+            effectiveLiquidityFactorX32 = (removedLiquidity * 2 ** 32) / netLiquidity;
```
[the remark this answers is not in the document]
guil-lambert: removedLiquidity is <uint128

contracts/PanopticPool.sol
```diff
@@ -1289,7 +1385,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
         }
 
         // revert if fingerprint for provided `_positionIdList` does not match the one stored for the `_account`
-        if (fingerprintIncomingList != currentHash) revert Errors.InputListFail();
+        if (fingerprintIncomingList != s_positionsHash[account]) revert Errors.InputListFail();
```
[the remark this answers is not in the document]
guil-lambert: :grimacing:

guil-lambert:
@coderabbitai can you compare the two PR linked in the first comment and make sure that the changes I made in this PR port all of those changes ~1-to-1?

contracts/PanopticPool.sol
```diff
@@ -379,8 +402,10 @@ contract PanopticPool is ERC1155Holder, Multicall {
         for (uint256 k = 0; k < pLength; ) {
```
flanagansteve, line 402: In #26 , we killed the pLength and c_user variables, replacing them with:

```solidity
balances = new uint256[2][](positionIdList.length);
for (uint256 k = 0; k < positionIdList.length; ) {
```

and simply using `user` everywhere `c_user` is used.

You should try porting that change here too and seeing whether there's actually any gas efficiency improvements in the current pLength / c_user usages - if not, the version without pLength and c_user is certainly more readable, and will reduce the divergence between v1 and v1.1

contracts/PanopticPool.sol
```diff
@@ -799,7 +851,7 @@ contract PanopticPool is ERC1155Holder, Multicall {
             lastObservedTick
         );
 
-        _checkSolvency(user, positionIdList, tickData, buffer);
+        _checkSolvency(user, positionIdList, tickData, buffer, usePremiaAsCollateral);
```
flanagansteve, line 854:
```suggestion
        _checkSolvency(
            user,
            positionIdList,
            PositionBalanceLibrary.packTickData(
                currentTick,
                fastOracleTick,
                slowOracleTick,
                lastObservedTick
            ),
            buffer,
            usePremiaAsCollateral
        );
```

Another change not ported over from #26 was killing the tickData variable and inlining it in the _checkSolvency call - you could try doing the same here, seems like it could be more gas efficient to remove a memory var:

contracts/PanopticPool.sol
```diff
@@ -1511,17 +1607,29 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @param positionIdList Exhaustive list of open positions for `owner` used for solvency checks where the tokenId to settle is placed at the last index
     /// @param owner The owner of the option position to make premium payments on
     /// @param legIndex the index of the leg in tokenId that is to be collected on (must be isLong=1)
+    /// @param usePremiaAsCollateral Whether to compute accumulated premia for all legs held by the settlee for collateral (true), or just owed premia for long legs (false)
```
flanagansteve, line 1610: > /// @dev This feature is only available when `owner` is solvent and has the requisite tokens to settle the premium.

We actually missed this in #25 , but this docstring is now kind of out of date - perhaps:

> /// @dev This feature is only available when `owner` is solvent and has the requisite collateral (in either token) to settle the premium.

contracts/PanopticPool.sol
```diff
@@ -1067,75 +1133,101 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-            (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
-
             uint128 positionSize = s_positionBalance[account][tokenId].positionSize();
 
             (LeftRightSigned longAmounts, ) = PanopticMath.computeExercisedAmounts(
                 tokenId,
                 positionSize
             );
 
+            TokenId _tokenId = tokenId;
```
flanagansteve, line 1176: When you compile on the latest solc version, do you actually save gas with this? If not, would kill this variable

contracts/libraries/PanopticMath.sol
```diff
@@ -1026,35 +1040,35 @@ library PanopticMath {
...
-    /// @param atTick The tick at which to convert between token0/token1 when redistributing the exercise fees
-    /// @param ct0 The collateral tracker for token0
-    /// @param ct1 The collateral tracker for token1
-    /// @return The LeftRight-packed deltas for token0/token1 to move from the exercisor to the exercisee
-    function getExerciseDeltas(
-        address exercisee,
-        LeftRightSigned exerciseFees,
+    /// @notice Substitutes surplus tokens to a caller in exchange for any potential token shortages prior to revoking virtual shares from a payor.
+    /// @param payor The address of the user being exercised/settled
+    /// @param fees If applicable, fees to debit from caller (rightSlot = currency0 left = currency1), 0 for `settleLongPremium`
```
flanagansteve:
```suggestion
    /// @param fees If applicable, fees to debit from caller (rightSlot = token0 left = token1), 0 for `settleLongPremium`
```

I would not use `currency` here. We made those changes in #25 because we were explicitly trying to swap to v4 language. This is still a v3 contract, though, so we should use v3's `token` verbiage

contracts/libraries/PanopticMath.sol
```diff
@@ -1026,35 +1040,35 @@ library PanopticMath {
...
+    /// @param atTick The tick at which to convert between currency0/currency1 when redistributing the surplus tokens
```
flanagansteve:
```suggestion
    /// @param atTick The tick at which to convert between token0/token1 when redistributing the surplus tokens
```

contracts/libraries/PanopticMath.sol
```diff
@@ -1026,35 +1040,35 @@ library PanopticMath {
...
+    /// @param ct0 The collateral tracker for currency0
```
flanagansteve:
```suggestion
    /// @param ct0 The collateral tracker for token0
```

contracts/libraries/PanopticMath.sol
```diff
@@ -1026,35 +1040,35 @@ library PanopticMath {
...
+    /// @param ct1 The collateral tracker for currency1
```
flanagansteve:
```suggestion
    /// @param ct1 The collateral tracker for token1
```

contracts/libraries/PanopticMath.sol
```diff
@@ -1026,35 +1040,35 @@ library PanopticMath {
...
+    /// @return The LeftRight-packed deltas for currency0/currency1 to move from the caller to the payor
```
flanagansteve:
```suggestion
    /// @return The LeftRight-packed deltas for token0/token1 to move from the caller to the payor
```

contracts/libraries/PanopticMath.sol
```diff
@@ -1026,35 +1040,35 @@ library PanopticMath {
...
+        address payor,
+        LeftRightSigned fees,
         int24 atTick,
         CollateralTracker ct0,
         CollateralTracker ct1
     ) external view returns (LeftRightSigned) {
         uint160 sqrtPriceX96 = Math.getSqrtRatioAtTick(atTick);
         unchecked {
-            // if the refunder lacks sufficient token0 to pay back the virtual shares, have the exercisor cover the difference in exchange for token1 (and vice versa)
+            // if the refunder lacks sufficient currency0 to pay back the virtual shares, have the caller cover the difference in exchange for currency1 (and vice versa)
```
flanagansteve:
```suggestion
            // if the refunder lacks sufficient token0 to pay back the virtual shares, have the caller cover the difference in exchange for token1 (and vice versa)
```

contracts/PanopticPool.sol
```diff
@@ -1067,75 +1133,101 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
         // validate the exercisor's position list (the exercisee's list will be evaluated after their position is force exercised)
         _validatePositionList(msg.sender, positionIdListExercisor, 0);
 
-        int24 twapTick = getUniV3TWAP();
+        int24 currentTick;
+        int24 lastObservedTick;
+        {
+            int24 fastOracleTick;
+            int24 slowOracleTick;
+            (fastOracleTick, slowOracleTick, lastObservedTick, currentTick, ) = PanopticMath
```
flanagansteve: Did you get the return order of these vars right? I'll come back to double check later but in the [panoptic-v1 public repo](https://github.com/panoptic-labs/panoptic-v1-core/blob/v1.0.x/contracts/libraries/PanopticMath.sol) we do:

```solidity
function getOracleTicks(
        IUniswapV3Pool univ3pool,
        uint256 miniMedian
    )
        external
        view
        returns (
            int24 currentTick,
            int24 fastOracleTick,
            int24 slowOracleTick,
            int24 latestObservation,
            uint256 medianData
        )
    {
```

contracts/PanopticPool.sol
```diff
@@ -1067,75 +1133,101 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+                .getOracleTicks(s_univ3pool, s_miniMedian);
+
+            if (
```
flanagansteve, line 1155: This would be a divergence from v1.1, but you could consider putting this in an `unchecked` - these shouldn't under or overflow:

```solidity:
Math.abs(lastObservedTick - fastOracleTick) > MAX_TICK_DELTA_SUBSTITUTION ||
Math.abs(lastObservedTick - slowOracleTick) > MAX_TICK_DELTA_SUBSTITUTION
```

contracts/PanopticPool.sol
flanagansteve: And other note to self: If this _does_ need to be corrected, and yet the tests were passing with the incorrect ordering, then we probably need to expand the test suite to cover divergence between ticks

contracts/PanopticPool.sol
```diff
@@ -1067,75 +1133,101 @@ contract PanopticPool is ERC1155Holder, Multicall {
     /// @param tokenId The position to be force exercised; this position must contain at least one out-of-range long leg
     /// @param positionIdListExercisee Post-burn list of open positions in the exercisee's (`account`) account
     /// @param positionIdListExercisor List of open positions in the exercisor's (`msg.sender`) account
+    /// @param usePremiaAsCollateral Whether to compute accumulated premia for all legs held by the exercisee(right slot)/exercisor(left slot) for collateral (>0), or just owed premia for long legs (0)
     function forceExercise(
         address account,
```
flanagansteve, line 1138: This would be another divergence from v1.1, but a more descriptive varname would be better here - `exercisee` would be good

Also, the docstring for `account` is wrong (probably copypasta from `liquidate`):

> /// @param account Address of the distressed account

a force exercisee is not necessarily distressed. instead, we should do something like:

> /// @param account Address of the account being force exercised

And update from `account` to `exercisee` if you agree there

contracts/PanopticPool.sol
```diff
@@ -1067,75 +1133,101 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
         TokenId tokenId,
```
flanagansteve, line 1139: Similarly, a more specific varname like `exercisedPosition` would be ideal - though I also don't want to push for too much divergence from v1.1

contracts/PanopticPool.sol
```diff
@@ -1582,8 +1689,45 @@ contract PanopticPool is ERC1155Holder, Multicall {
             emit PremiumSettled(owner, tokenId, legIndex, realizedPremia);
         }
 
+        int24 lastObservedTick;
+        uint96 tickData;
+        {
+            int24 fastOracleTick;
+            int24 slowOracleTick;
+            (fastOracleTick, slowOracleTick, lastObservedTick, , ) = PanopticMath.getOracleTicks(
```
flanagansteve: Same comment about ordering of getOracleTicks return vals - will come back and check

flanagansteve (changes requested):
Main concern is about using the returned getOracleTicks values correctly - most other bits are gas related / code quality related / docstrings

I compared this PR with 25 and 26 one to one and can confirm you carried over all changes

flanagansteve:
Confirmed - this branch does return currentOracleTick first, then fast/slow/last: https://github.com/panoptic-labs/panoptic-next-core-private/blob/1c3c37e1493885b2e7237f5bdff3a6dd7e56f1f7/contracts/libraries/PanopticMath.sol#L167

Which means a) we need to fix the two getOracleTicks calls, but more importantly b) we need to add more test cases, if the tests were passing with this incorrect usage - ideally documenting that exercisors / settleLongPremium-ers can manipulate price by 2% and still call successfully

contracts/PanopticPool.sol
```diff
@@ -1246,12 +1335,20 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
-                uint24(medianData >> ((uint24(medianData >> (192 + 3 * 3)) % 8) * 24))
-            ) + int24(uint24(medianData >> ((uint24(medianData >> (192 + 3 * 4)) % 8) * 24)))) / 2;
-
             // If ticks have recently deviated more than +/- ~10%, enforce covered mints
-            return Math.abs(currentTick - medianTick) > MAX_TICKS_DELTA;
+            return
+                Math.abs(
+                    currentTick -
+                        (int24(
+                            uint24(medianData >> ((uint24(medianData >> (192 + 3 * 3)) % 8) * 24))
```
flanagansteve, line 1343: `192 + 3 * 3` is a constant - we could just hardcode that and save gas

contracts/PanopticPool.sol
```diff
@@ -1246,12 +1335,20 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+                        ) +
+                            int24(
+                                uint24(
+                                    medianData >> ((uint24(medianData >> (192 + 3 * 4)) % 8) * 24)
```
flanagansteve, line 1347: Same thing applies to `192 + 3 * 4`

contracts/PanopticPool.sol
```diff
@@ -1246,12 +1335,20 @@ contract PanopticPool is ERC1155Holder, Multicall {
...
+                                )
+                            )) /
+                        2
```
flanagansteve, line 1350: You dropped a paren from the v4 expression:

```solidity
    /// @notice Checks whether the current tick has deviated by `> MAX_TICKS_DELTA` from the slow oracle median tick.
    /// @return Whether the current tick has deviated from the median by `> MAX_TICKS_DELTA`
    function isSafeMode() public view returns (bool) {
        uint256 medianData = s_miniMedian;
        unchecked {
            // If ticks have recently deviated more than +/- ~10%, enforce covered mints
            return
                Math.abs(
                    V4StateReader.getTick(POOL_MANAGER_V4, _V4PoolId()) -
                        ((int24(
                            uint24(medianData >> ((uint24(medianData >> (192 + 3 * 3)) % 8) * 24))
                        ) +
                            int24(
                                uint24(
                                    medianData >> ((uint24(medianData >> (192 + 3 * 4)) % 8) * 24)
                                )
                            )) / 2)
                ) > MAX_TICKS_DELTA;
        }
    }
```

Which should be fine considering that division has precedence over subtraction anyways - but you could consider re-adding for clarity / to match v1.1 better

contracts/PanopticPool.sol
```diff
@@ -1246,12 +1335,20 @@ contract PanopticPool is ERC1155Holder, Multicall {
 
         uint256 medianData = s_miniMedian;
         unchecked {
-            int24 medianTick = (int24(
-                uint24(medianData >> ((uint24(medianData >> (192 + 3 * 3)) % 8) * 24))
-            ) + int24(uint24(medianData >> ((uint24(medianData >> (192 + 3 * 4)) % 8) * 24)))) / 2;
-
             // If ticks have recently deviated more than +/- ~10%, enforce covered mints
-            return Math.abs(currentTick - medianTick) > MAX_TICKS_DELTA;
+            return
```
flanagansteve, line 1339: In fact, it seems possible that we can simplify the whole expression - if we're always just averaging the 3rd and 4th rank observation, can't we just hardcode the amount to bitshift, add those two values, divide by two, and subtract from currentTick?

```solidity
function isSafeMode() public view returns (bool) {
    (, int24 currentTick, , , , , ) = s_univ3pool.slot0();
    uint256 medianData = s_miniMedian;
    unchecked {
        return
            Math.abs(
                currentTick -
                    ((int24(uint24(medianData >> 24)) +  // slot 1 (rank 3)
                      int24(uint24(medianData)))          // slot 0 (rank 4)
                    ) / 2
            ) > MAX_TICKS_DELTA;
    }
}
```

IDK if >> 24 is the correct amount to shift (Claude says so but I don't trust it) but I think you get the idea

Given that the docstring also says you're comparing currentTick to the "slow oracle median tick", we could also arguably just replace that with the slowOracleTick and remove this custom calculation entirely

flanagansteve (commented):
Passed things over one more time to check the math changes too

contracts/PanopticPool.sol
guil-lambert: Will look into it.

`c_user` was probably due to stacks too deeps
`pLength` is also because it is used twice in that context, so maybe gas savings were involved.

contracts/PanopticPool.sol
guil-lambert: This is for stack to deep errors

contracts/PanopticPool.sol
guil-lambert: - `uint24(medianData >> (192 + 3 * 3)) % 8)` <- this finds the index of the 3th largest value
- `uint24(medianData >> (192 + 3 * 4)) % 8)` <- this finds the index of the 4th largest value

Then the bitshifting shifts by the appropriate number of "24bits" to extract them.

contracts/PanopticPool.sol
guil-lambert: I am pretty sure this gets transformed into its value at compile and is treated as a constant

contracts/PanopticPool.sol
guil-lambert: This is fixed in the dispatchFrom PR:

https://github.com/panoptic-labs/panoptic-next-core-private/blob/128966825d06386e74f00ca4d2d10058573e5684/contracts/PanopticPool.sol#L894-L901

contracts/PanopticPool.sol
guil-lambert: Not relevant anymore since we're doing dispatchFrom for those functions now

contracts/PanopticPool.sol
guil-lambert: not relevant w/ dispatchFrom

contracts/PanopticPool.sol
guil-lambert: Fixed in my latest commit!

contracts/PanopticPool.sol
guil-lambert: will look into implementing it in dispatchFrom, now it's the same check as for liquidations

contracts/PanopticPool.sol
guil-lambert: See comment above

contracts/PanopticPool.sol
guil-lambert: Will fix if needed in dispatchFrom

contracts/PanopticPool.sol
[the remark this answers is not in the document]
guil-lambert: Thanks for your input, will fix in a subsequent PR!

contracts/PanopticPool.sol
guil-lambert: Added comment

PR #161, opened 2025-09-15
Fix2: prevent zero-value deposits and withdrawals in CollateralTracker

guil-lambert:
Adds validation to prevent zero-value operations in the CollateralTracker contract's deposit, mint, withdraw, and redeem functions. This prevents potential issues with zero-amount transactions that could lead to unexpected behavior or gas 
waste.

## Motivation
This is in response to suggestions by DevDacian on twitter in response to the bunni_xyz hack:  https://x.com/DevDacian/status/1965727746803544534

3⃣ enforce that deposits must return > 0 shares, redeems must return > 0 tokens and that the inputs must also be > 0

Also, zero-value operations in DeFi protocols can:
1. Waste gas without performing meaningful state changes
2. Potentially bypass certain security checks or assumptions
3. Create edge cases in accounting logic
4. Lead to confusion in event logs and transaction history

### Contract Changes
- Added `BelowMinimumRedemption` error to `Errors.sol` library
- Added zero-value checks to 5 functions in `CollateralTracker.sol`:
  - `deposit()`: Reverts if attempting to deposit 0 assets
  - `mint()`: Reverts if shares would result in 0 assets
  - `withdraw()`: Reverts if attempting to withdraw 0 assets (both overloads)
  - `redeem()`: Reverts if shares would result in 0 assets after `previewRedeem()`

## Summary by CodeRabbit

- New Features
  - Added a clear error when attempting below-minimum (zero-amount) deposit, mint, withdraw, or redeem operations.
- Bug Fixes
  - Standardized behavior across deposit, mint, withdraw, and redeem to consistently reject zero-amount actions with a dedicated error, improving validation and user feedback.
- Tests
  - Expanded test coverage to verify reverts on below-minimum actions across all flows, including on-behalf scenarios, and tightened fuzz bounds to ensure meaningful input ranges.

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -393,6 +393,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @return shares The amount of Panoptic pool shares that were minted to the recipient
     function deposit(uint256 assets, address receiver) external returns (uint256 shares) {
         if (assets > type(uint104).max) revert Errors.DepositTooLarge();
+        if (assets == 0) revert Errors.BelowMinimumRedemption();
 
         shares = previewDeposit(assets);
 
@@ -451,6 +452,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         assets = previewMint(shares);
 
         if (assets > type(uint104).max) revert Errors.DepositTooLarge();
+        if (assets == 0) revert Errors.BelowMinimumRedemption();
 
         // transfer assets (underlying token funds) from the user/the LP to the PanopticPool
         // in return for the shares to be minted
@@ -516,6 +518,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             if (allowed != type(uint256).max) allowance[owner][msg.sender] = allowed - shares; // Saves gas for unlimited approvals.
         }
 
+        if (assets == 0) revert Errors.BelowMinimumRedemption();
         // burn collateral shares of the Panoptic Pool funds (this ERC20 token)
         _burn(owner, shares);
 
@@ -558,6 +561,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             if (allowed != type(uint256).max) allowance[owner][msg.sender] = allowed - shares; // Saves gas for unlimited approvals.
         }
 
+        if (assets == 0) revert Errors.BelowMinimumRedemption();
         // burn collateral shares of the Panoptic Pool funds (this ERC20 token)
         _burn(owner, shares);
 
@@ -620,6 +624,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
 
         assets = previewRedeem(shares);
 
+        if (assets == 0) revert Errors.BelowMinimumRedemption();
         // burn collateral shares of the Panoptic Pool funds (this ERC20 token)
         _burn(owner, shares);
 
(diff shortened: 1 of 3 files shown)


contracts/CollateralTracker.sol
```diff
@@ -393,6 +393,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @return shares The amount of Panoptic pool shares that were minted to the recipient
     function deposit(uint256 assets, address receiver) external returns (uint256 shares) {
         if (assets > type(uint104).max) revert Errors.DepositTooLarge();
+        if (assets == 0) revert Errors.BelowMinimumRedemption();
```
flanagansteve, line 396:
```suggestion
        if (assets <= 1) revert Errors.BelowMinimumRedemption();
```

We could consider going farther and taking the other piece of Dacian's advice:

> 1⃣enforce minimum transaction amounts; eg attackers often do 1 wei transactions to trigger rounding issues but normal users never do this

Even going all the way to 100 units is arguable - for the tokens we support this shouldn't realistically ever be a consequential amount of money, and it would give us significant protection against rounding errors

contracts/CollateralTracker.sol
```diff
@@ -393,6 +393,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
 
         shares = previewDeposit(assets);
```
flanagansteve, line 398: Is there perhaps also additional security benefit in enforcing that the _mint call:

> _mint(receiver, shares);

Is for >= 1 share?

contracts/CollateralTracker.sol
```diff
@@ -619,6 +623,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         }
 
         assets = previewRedeem(shares);
+        if (assets == 0) revert Errors.BelowMinimumRedemption();
 
         // burn collateral shares of the Panoptic Pool funds (this ERC20 token)
         _burn(owner, shares);
```
flanagansteve, line 629: Other methods on CollateralTracker that could benefit from this treatment:

- settleLiquidation - it seems more secure to enforce that the PanopticPool never be allowed to CT.settleLiquidation with a 0-value bonus
- transfer / transferFrom - what would be the legitimate reason to make a transfer call for 0 shares?
  - Maybe some transfer that is part of a loop or batch, in which it would be painful to put a guard in the middle of the loop, such as a contract that integrates the CT?
- refund - .refund gets called from PP.forceExercise or .liquidate, to move funds from a force exercisor / liquidatee to exercisee/liquidator, to pay a fee / bonus (net of protocol-delgated amounts that are temporarily lent to the exercisor/liquidator). Should the .liquidate and .forceExercise flows be prevented from trying to move 0 tokens?
- takeCommissionAddData - Should we let the PP call this method with zero values for _both_ longAmounts and shortAmount? Seems illegitimate
  - Similarly for exercise - will longAmount (The notional value of the long legs of the position (if any)) and shortAmount (The notional value of the short legs of the position (if any)) ever be _simultaneously_ zero? I believe that is illegitimate too

flanagansteve, line 629: I believe we're all set for other contracts - in PanopticPool, we already enforce nonzero amounts for mintOptions, and none of the other methods accept an amount value. 

In SemiFungiblePositionManager, we could arguably enforce nonzero amounts of amount0Owed/amount1Owed in the callbacks, but we only bother doing a safeTransferFrom if amount > 0, so not really any damage someone could do there. We already check if chunkLiquidity == 0 in _createLegInAMM.

flanagansteve (changes requested):
This seems like a fine change to me. Its the kind of thing where I don't know how much security this _really_ buys us, but there seems to be no downside aside from gas impact, which is minimal.

I do note that you probably want to additionally protect that the output of previewDeposit / previewWithdraw is nonzero, not just the user-supplied input, as well as some other notes below

PR #162, opened 2025-09-15
Feat0: remove MEV commission fee from CollateralTracker deposit/mint functions

guil-lambert:
## Summary
Removes the MEV commission fee (1% tax) from the CollateralTracker's deposit and mint operations, simplifying the share calculation logic and making the vault implementation more standard.

### Contract Changes (`CollateralTracker.sol`)
- **`previewDeposit()`**: Removed commission fee calculation, now uses direct proportion (`assets * totalSupply / totalAssets`)
- **`maxMint()`**: Removed commission fee adjustment, returns direct share conversion
- **`previewMint()`**: Simplified to standard vault math without commission fee deduction

## Technical Details

**Before:** 
- Deposits were taxed with a `COMMISSION_FEE` = 20bps
- Share calculations: `shares = assets * (DECIMALS - COMMISSION_FEE) * totalSupply / (totalAssets * DECIMALS)`
- Tests had to artificially inflate balances to compensate

**After:**
- Standard ERC4626 vault math without fees
- Share calculations: `shares = assets * totalSupply / totalAssets`
- Clean test setup without artificial adjustments

## Motivation
1. **Simplification**: Removes complexity from core deposit/withdraw logic
2. **Standard Compliance**: Aligns with standard ERC4626 vault implementations
3. **Gas Optimization**: Simpler calculations mean lower gas costs
4. **Test Clarity**: Tests no longer need workarounds for the fee mechanism

## Impact
- Users will receive more shares for the same asset deposits (no 1% haircut)
- Existing integrations expecting the fee may need updates
- More predictable and transparent vault behavior

## Breaking Changes
⚠️ **This is a breaking change** - removes a fee mechanism that was previously applied to all deposits and mints. Any existing deployments or integrations expecting this fee behavior will need to be updated.

## Summary by CodeRabbit

* **New Features**
  * Conversion previews and limits now use pure pro‑rata calculations with no commission/MEV tax applied, so deposits, mints, withdrawals and previews align exactly to total assets and supply.
  * Estimates are more predictable and consistent across the app.

* **Tests**
  * Test scenarios and expected values updated to reflect tax‑free conversions and exact proportional calculations.

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -374,15 +374,11 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param assets The amount of assets to be deposited
     /// @return shares The amount of shares that can be minted
     function previewDeposit(uint256 assets) public view returns (uint256 shares) {
-        // compute the MEV tax, which is equal to a single payment of the commissionRate on the FINAL (post mev-tax) assets paid
-        unchecked {
-            shares = Math.mulDiv(assets, totalSupply, totalAssets());
-        }
+        shares = Math.mulDiv(assets, totalSupply, totalAssets());
     }
 
     /// @notice Deposit underlying tokens (assets) to the Panoptic pool from the LP and mint corresponding amount of shares.
     /// @dev There is a maximum asset deposit limit of `2^104 - 1`.
-    /// @dev An "MEV tax" is levied, which is equal to a single payment of the commissionRate BEFORE adding the funds.
     /// @dev Shares are minted and sent to the LP (`receiver`).
     /// @param assets Amount of assets deposited
     /// @param receiver User to receive the shares
@@ -413,9 +409,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @notice Returns the maximum shares received for a deposit.
     /// @return maxShares The maximum amount of shares that can be minted
     function maxMint(address) external view returns (uint256 maxShares) {
-        unchecked {
-            return convertToShares(type(uint104).max);
-        }
+        return convertToShares(type(uint104).max);
     }
 
     /// @notice Returns the amount of assets that would be deposited to mint a given amount of shares.
@@ -429,7 +423,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
 
     /// @notice Deposit required amount of assets to receive specified amount of shares.
     /// @dev There is a maximum asset deposit limit of `2^104 - 1`.
-    /// @dev An "MEV tax" is levied, which is equal to a single payment of the commissionRate BEFORE adding the funds.
     /// @dev Shares are minted and sent to the LP (`receiver`).
     /// @param shares Amount of shares to be minted
     /// @param receiver User to receive the shares
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -376,11 +376,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     function previewDeposit(uint256 assets) public view returns (uint256 shares) {
         // compute the MEV tax, which is equal to a single payment of the commissionRate on the FINAL (post mev-tax) assets paid
         unchecked {
-            shares = Math.mulDiv(
-                assets * (DECIMALS - COMMISSION_FEE),
-                totalSupply,
-                totalAssets() * DECIMALS
-            );
+            shares = Math.mulDiv(assets, totalSupply, totalAssets());
         }
     }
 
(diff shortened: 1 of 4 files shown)


contracts/CollateralTracker.sol
```diff
@@ -376,11 +376,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     function previewDeposit(uint256 assets) public view returns (uint256 shares) {
```
flanagansteve, line 376: You can also remove lines 127 to 130 where the var is defined:

```solidity
    /// @notice The commission fee, in basis points, collected from PLPs at option mint.
    /// @dev In Panoptic, options never expire, commissions are only paid when a new position is minted.
    /// @dev We believe that this will eliminate the impact of the commission fee on the user's decision-making process when closing a position.
uint256 immutable COMMISSION_FEE;
```

as well as the constructor arg and logic for it:

```solidity
/// @notice Set immutable parameters for the Collateral Tracker.
    /// @param _commissionFee The commission fee, in basis points, collected from PLPs at option mint
    ...
    constructor(
        uint256 _commissionFee,
        ...
    ) {
        COMMISSION_FEE = _commissionFee;
        ...
    }
```

contracts/CollateralTracker.sol
```diff
@@ -376,11 +376,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
         // compute the MEV tax, which is equal to a single payment of the commissionRate on the FINAL (post mev-tax) assets paid
```
flanagansteve:
```suggestion
```

Can now remove this outdated line

contracts/CollateralTracker.sol
```diff
@@ -376,11 +376,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
         unchecked {
```
flanagansteve: I actually think you don't need an `unchecked` block here anymore, since mulDiv itself is already inside an `unchecked` block - it was previously here for the gas savings we got on the `assets * (...)`, `DECIMALS - COMISSION_FEE`, etc

contracts/CollateralTracker.sol
```diff
@@ -418,7 +414,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @return maxShares The maximum amount of shares that can be minted
     function maxMint(address) external view returns (uint256 maxShares) {
         unchecked {
```
flanagansteve: Same here, no unchecked needed anymore i believe

contracts/CollateralTracker.sol
flanagansteve: Actually, this is wrong - this PR keeps the COMMISSION_FEE so that we can use it in `_getExchangedAmount` - right @guil-lambert ? and then we remove it fully in the interest rate PR?

contracts/CollateralTracker.sol
```diff
@@ -428,16 +424,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
-        // finalAssets - commissionRate * finalAssets = convertedAssets
-        // finalAssets * (1 - commissionRate) = convertedAssets
-        // finalAssets = convertedAssets / (1 - commissionRate)
-        assets = Math.mulDivRoundingUp(
-            shares * DECIMALS,
-            totalAssets(),
-            totalSupply * (DECIMALS - COMMISSION_FEE)
-        );
+        assets = Math.mulDivRoundingUp(shares, totalAssets(), totalSupply);
     }
```
flanagansteve, line 423: There is now an outdated docstring on `deposit` that we should update - it is currently:

```solidity
/// @notice Deposit underlying tokens (assets) to the Panoptic pool from the LP and mint corresponding amount of shares.
    /// @dev There is a maximum asset deposit limit of `2^104 - 1`.
    /// @dev An "MEV tax" is levied, which is equal to a single payment of the commissionRate BEFORE adding the funds.
    /// @dev Shares are minted and sent to the LP (`receiver`).
    /// @param assets Amount of assets deposited
    /// @param receiver User to receive the shares
```

And we should remove:

>     /// @dev An "MEV tax" is levied, which is equal to a single payment of the commissionRate BEFORE adding the funds.

flanagansteve: Same thing on `mint`

flanagansteve (changes requested):
Looks great, just found two `unchecked` blocks you can now remove as well as three docstring removals to make, as well as one thread of me talking to myself, lol

contracts/CollateralTracker.sol
guil-lambert: Yes, commissions still exist. 

It's the MEV tax, which was equal to the commission fee, that is removed

PR #164, opened 2025-09-17
Feat4: refactor: PanopticPool interface with unified dispatch system

guil-lambert:
This PR introduces a major architectural change to the PanopticPool contract, consolidating mint/burn and liquidate/forceExercise/settleLongPremium operations through a unified dispatch system while improving premium handling and oracle validation.

This supersedes this PR: https://github.com/panoptic-labs/panoptic-next-core-private/pull/163 (which has now been closed)

### Core Interface Changes:

- Replace individual mintOptions() and burnOptions() methods with unified dispatch() system
- Add new dispatchFrom() for liquidations, force exercises, and premium settlements
- Consolidate multiple operation types (mint, burn, liquidate, force exercise, settle premium) into single entry points

### Contract Architecture:

- Consolidate solvency checking with unified _checkSolvencyAtTicks() return values
- Streamline position validation and liquidity spread checking

### Test Infrastructure:

- Update all tests to use new dispatch-based helper functions
- Maintain backwards compatibility through wrapper functions
 
TODO:

- [ ] add more tests for multiple mints/burns
- [ ] make sure all existing tests still pass 

## Summary by CodeRabbit

- New Features
  - Toggle to treat accumulated premiums as collateral across withdrawals, settlements, and liquidations.
  - Batch mint/burn via unified dispatch and coordinated multi-list dispatchFrom.
  - ERC1155 single-transfer receipt support.

- Breaking Changes
  - Withdrawal and pool public APIs updated to accept a premiums-as-collateral flag and batch entry points.
  - Public refund API renamed (exercise-deltas → refund-oriented).
  - Oracle staleness error renamed to StaleOracle.

- Improvements
  - Refined solvency, premia settlement, and oracle/tick handling.

- Chores
  - Optimizer runs parameter adjusted.

### DIFF
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -113,10 +113,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @notice Amount of assets moved from the Panoptic Pool to the AMM.
     uint128 internal s_inAMM;
 
-    /// @notice Additional risk premium charged on intrinsic value of ITM positions,
-    /// denominated in hundredths of basis points.
-    uint128 internal s_ITMSpreadFee;
-
     /// @notice The fee of the Uniswap pool in hundredths of basis points.
     uint24 internal s_poolFee;
 
@@ -149,10 +145,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @dev i.e 90% -> 0.9 * 10_000 = 9_000.
     uint256 immutable SATURATED_POOL_UTIL;
 
-    /// @notice Multiplier, in basis points, to the pool fee that is charged on the intrinsic value of ITM positions.
-    /// @dev e.g. ITM_SPREAD_MULTIPLIER = 20_000, s_ITMSpreadFee = 2 * s_poolFee.
-    uint256 immutable ITM_SPREAD_MULTIPLIER;
-
     /*//////////////////////////////////////////////////////////////
                             ACCESS CONTROL
     //////////////////////////////////////////////////////////////*/
@@ -174,15 +166,13 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param _forceExerciseCost Basal cost (in bps of notional) to force exercise an out-of-range position
     /// @param _targetPoolUtilization Target pool utilization below which buying+selling is optimal, represented as percentage * 10_000
     /// @param _saturatedPoolUtilization Pool utilization above which selling is 100% collateral backed, represented as percentage * 10_000
-    /// @param _ITMSpreadMultiplier Multiplier, in basis points, to the pool fee that is charged on the intrinsic value of ITM positions
     constructor(
         uint256 _commissionFee,
         uint256 _sellerCollateralRatio,
         uint256 _buyerCollateralRatio,
         int256 _forceExerciseCost,
         uint256 _targetPoolUtilization,
-        uint256 _saturatedPoolUtilization,
-        uint256 _ITMSpreadMultiplier
+        uint256 _saturatedPoolUtilization
     ) {
         COMMISSION_FEE = _commissionFee;
         SELLER_COLLATERAL_RATIO = _sellerCollateralRatio;
@@ -190,7 +180,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         FORCE_EXERCISE_COST = _forceExerciseCost;
         TARGET_POOL_UTIL = _targetPoolUtilization;
         SATURATED_POOL_UTIL = _saturatedPoolUtilization;
-        ITM_SPREAD_MULTIPLIER = _ITMSpreadMultiplier;
     }
 
     /// @notice Initialize a new collateral tracker for a specific token corresponding to the Panoptic Pool being created by the factory that called it.
@@ -235,11 +224,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
 
         // store whether the current collateral token is token0 (true) or token1 (false)
         s_underlyingIsToken0 = underlyingIsToken0;
-
-        // Additional risk premium charged on intrinsic value of ITM positions
-        unchecked {
-            s_ITMSpreadFee = uint128((ITM_SPREAD_MULTIPLIER * fee) / DECIMALS);
-        }
     }
 
     /*//////////////////////////////////////////////////////////////
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -1062,14 +1062,13 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param longAmount The amount of longs
     /// @param shortAmount The amount of shorts
     /// @param swappedAmount The amount of tokens moved during creation of the option position
-    /// @return The final utilization of the collateral vault
-    /// @return The total amount of commission (base rate + ITM spread) paid
+    /// @return The final utilization of the collateral vault (rightSlot) and the total amount of commission (base rate + ITM spread) paid (leftSlot)
     function settleMint(
         address optionOwner,
         int128 longAmount,
         int128 shortAmount,
         int128 swappedAmount
-    ) external onlyPanopticPool returns (uint32, uint128) {
+    ) external onlyPanopticPool returns (LeftRightUnsigned, int128) {
         (uint32 utilization, uint128 commission, int128 tokenPaid) = _updateBalancesAndSettle(
             true, // isCreation = true
             optionOwner,
@@ -1078,7 +1077,10 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             swappedAmount,
             0 // realizedPremium not used
         );
-        return (utilization, commission);
+        return (
+            LeftRightUnsigned.wrap(0).toRightSlot(utilization).toLeftSlot(commission),
+            tokenPaid
+        );
     }
 
     /// @notice Exercise an option and pay to the seller what is owed from the buyer.
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -984,6 +984,79 @@ contract CollateralTracker is ERC20Minimal, Multicall {
                      OPTION EXERCISE AND COMMISSION
     //////////////////////////////////////////////////////////////*/
 
+    /// @notice Internal function to handle all balance and state updates for position creation and closing.
+    /// @param isCreation A boolean flag to indicate if this is for option creation (true) or closing (false).
+    /// @param optionOwner The user minting the option
+    /// @param longAmount The amount of longs
+    /// @param shortAmount The amount of shorts
+    /// @param swappedAmount The amount of tokens moved during creation of the option position
+    ///
+    function _updateBalancesAndSettle(
+        bool isCreation,
+        address optionOwner,
+        int128 longAmount,
+        int128 shortAmount,
+        int128 swappedAmount,
+        int128 realizedPremium
+    ) internal returns (uint32, uint128, int128) {
+        unchecked {
+            int256 tokenToPay;
+            uint128 commission;
+
+            // current available assets belonging to PLPs (updated after settlement) excluding any premium paid
+            int256 updatedAssets = int256(uint256(s_poolAssets)) - swappedAmount;
+
+            if (isCreation) {
+                {
+                    int256 intrinsicValue = int256(swappedAmount) - (shortAmount - longAmount);
+
+                    // the swap commission is paid on the intrinsic value (if a swap occurred; users who mint covered options with their own collateral do not pay this fee)
+                    commission = uint128(
+                        Math.unsafeDivRoundingUp(
+                            uint256(uint128(shortAmount + longAmount)) * COMMISSION_FEE,
+                            DECIMALS
+                        )
+                    );
+
+                    tokenToPay = intrinsicValue + int128(commission);
+                }
+            } else {
+                // add premium and token deltas not covered by swap to be paid/collected on position close
+                tokenToPay = int256(swappedAmount) - (longAmount - shortAmount) - realizedPremium;
+            }
+
+            // Mint/Burn Shares
+            if (tokenToPay > 0) {
+                uint256 sharesToBurn = Math.mulDivRoundingUp(
+                    uint256(tokenToPay),
+                    totalSupply,
+                    totalAssets()
+                );
+                _burn(optionOwner, sharesToBurn);
+            } else if (tokenToPay < 0) {
+                uint256 sharesToMint = convertToShares(uint256(-tokenToPay));
+                _mint(optionOwner, sharesToMint);
+            }
+
+            // Update Pool Assets
+            if (isCreation) {
+                s_poolAssets = uint256(updatedAssets).toUint128();
+            } else {
+                s_poolAssets = uint256(updatedAssets + realizedPremium).toUint128();
+            }
+
+            // Update s_inAMM
+            s_inAMM = uint256(
+                int256(uint256(s_inAMM)) +
+                    (isCreation ? (shortAmount - longAmount) : -(shortAmount - longAmount))
+            ).toUint128();
+
+            uint32 utilization = isCreation ? uint32(_poolUtilization()) : 0;
+
+            return (utilization, commission, int128(tokenToPay));
+        }
+    }
+
     /// @notice Take commission and settle ITM amounts on option creation.
     /// @param optionOwner The user minting the option
     /// @param longAmount The amount of longs
@@ -997,52 +1070,15 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         int128 shortAmount,
         int128 swappedAmount
     ) external onlyPanopticPool returns (uint32, uint128) {
-        unchecked {
-            // current available assets belonging to PLPs (updated after settlement) excluding any premium paid
-            int256 updatedAssets = int256(uint256(s_poolAssets)) - swappedAmount;
-
-            int256 tokenToPay;
-            uint128 commission;
-
-            {
-                int256 intrinsicValue = int256(swappedAmount) - (shortAmount - longAmount);
-
-                // the swap commission is paid on the intrinsic value (if a swap occurred; users who mint covered options with their own collateral do not pay this fee)
-                commission = uint128(
-                    Math.unsafeDivRoundingUp(
-                        uint256(uint128(shortAmount + longAmount)) * COMMISSION_FEE,
-                        DECIMALS
-                    )
-                );
-
-                tokenToPay = intrinsicValue + int128(commission);
-            }
-
-            // compute tokens to be paid due to swap
-            // mint or burn tokens due to minting in-the-money
-            if (tokenToPay > 0) {
-                // if user must pay tokens, burn them from user balance
-                uint256 sharesToBurn = Math.mulDivRoundingUp(
-                    uint256(tokenToPay),
-                    totalSupply,
-                    totalAssets()
-                );
-                _burn(optionOwner, sharesToBurn);
-            } else if (tokenToPay < 0) {
-                // if user must receive tokens, mint them
-                uint256 sharesToMint = convertToShares(uint256(-tokenToPay));
-                _mint(optionOwner, sharesToMint);
-            }
-
-            // update stored asset balances with net moved amounts
-            // the inflow or outflow of pool assets is defined by the swappedAmount: it includes both the ITM swap amounts and the short/long amounts used to create the position
-            // however, any intrinsic value is paid for by the users, so we only add the portion that comes from PLPs: the short/long amounts
-            // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
-            s_poolAssets = uint256(updatedAssets).toUint128();
-            s_inAMM = uint256(int256(uint256(s_inAMM)) + (shortAmount - longAmount)).toUint128();
-
-            return (uint32(_poolUtilization()), commission);
-        }
+        (uint32 utilization, uint128 commission, int128 tokenPaid) = _updateBalancesAndSettle(
+            true, // isCreation = true
+            optionOwner,
+            longAmount,
+            shortAmount,
+            swappedAmount,
+            0 // realizedPremium not used
+        );
+        return (utilization, commission);
     }
 
     /// @notice Exercise an option and pay to the seller what is owed from the buyer.
@@ -1060,37 +1096,15 @@ contract CollateralTracker is ERC20Minimal, Multicall {
         int128 swappedAmount,
         int128 realizedPremium
     ) external onlyPanopticPool returns (int128) {
-        unchecked {
-            // current available assets belonging to PLPs (updated after settlement) excluding any premium paid
-            int256 updatedAssets = int256(uint256(s_poolAssets)) - swappedAmount;
-
-            // add premium and token deltas not covered by swap to be paid/collected on position close
-            int256 tokenToPay = int256(swappedAmount) -
-                (longAmount - shortAmount) -
-                realizedPremium;
-
-            if (tokenToPay > 0) {
-                // if user must pay tokens, burn them from user balance (revert if balance too small)
-                uint256 sharesToBurn = Math.mulDivRoundingUp(
-                    uint256(tokenToPay),
-                    totalSupply,
-                    totalAssets()
-                );
-                _burn(optionOwner, sharesToBurn);
-            } else if (tokenToPay < 0) {
-                // if user must receive tokens, mint them
-                uint256 sharesToMint = convertToShares(uint256(-tokenToPay));
-                _mint(optionOwner, sharesToMint);
-            }
-
-            // update stored asset balances with net moved amounts
-            // any intrinsic value is paid for by the users, so we do not add it to s_inAMM
-            // premia is not included in the balance since it is the property of options buyers and sellers, not PLPs
-            s_poolAssets = uint256(updatedAssets + realizedPremium).toUint128();
-            s_inAMM = uint256(int256(uint256(s_inAMM)) - (shortAmount - longAmount)).toUint128();
-
-            return (int128(tokenToPay));
-        }
+        (, , int128 tokenPaid) = _updateBalancesAndSettle(
+            false, // isCreation = false
+            optionOwner,
+            longAmount,
+            shortAmount,
+            swappedAmount,
+            realizedPremium
+        );
+        return tokenPaid;
     }
 
     /*//////////////////////////////////////////////////////////////
--- a/contracts/CollateralTracker.sol
+++ b/contracts/CollateralTracker.sol
@@ -991,7 +991,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param swappedAmount The amount of tokens moved during creation of the option position
     /// @return The final utilization of the collateral vault
     /// @return The total amount of commission (base rate + ITM spread) paid
-    function takeCommissionAddData(
+    function settleMint(
         address optionOwner,
         int128 longAmount,
         int128 shortAmount,
@@ -1053,7 +1053,7 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @param swappedAmount The amount of tokens moved during the option close
     /// @param realizedPremium Premium to settle on the current positions
     /// @return The amount of tokens paid when closing that position
-    function exercise(
+    function settleBurn(
         address optionOwner,
         int128 longAmount,
         int128 shortAmount,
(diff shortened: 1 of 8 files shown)


contracts/CollateralTracker.sol
```diff
@@ -113,10 +113,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @notice Amount of assets moved from the Panoptic Pool to the AMM.
```
flanagansteve, line 113: note to self: keep an eye on reentrancy

contracts/CollateralTracker.sol
```diff
@@ -113,10 +113,6 @@ contract CollateralTracker is ERC20Minimal, Multicall {
...
     uint128 internal s_inAMM;
 
-    /// @notice Additional risk premium charged on intrinsic value of ITM positions,
```
flanagansteve, line 116: @guil-lambert trying to do a little work today as i recover - trying to understand why this PR includes the removal of s_ITMSpreadFee.

Maybe these changes are not part of this PR? Is [feat/long-premia-only-at-mint](https://github.com/panoptic-labs/panoptic-next-core-private/tree/feat/long-premia-only-at-mint) the correct base branch for this PR, or is another one?

If it is - then I think the reason you're removing this variable is that it's used in `_getExchangedAmount`, which was only used in `takeCommissionAddData` to calculate a commission on the option. You're replacing `takeCommissionAddData` with `_updateBalancesAndSettle`, and you still calculate a commission:

```solidity
commission = uint128(
                        Math.unsafeDivRoundingUp(
                            uint256(uint128(shortAmount + longAmount)) * COMMISSION_FEE,
                            DECIMALS
                        )
                    );
```

But you now use the `COMMISSION_FEE` const instead of `s_ITMSpreadFee`. Are these equivalent values? Or does this PR aim to both alter how commissions are calculated _and_ unify actions into a dispatch method? Could you document the economic changes you're hoping to make if so? Just want to make sure that my PR review confirms the functionality you're aiming for correctly

flanagansteve: (My brain is only at like 75%, so apologies if these are dumb questions!)

contracts/PanopticPool.sol
```diff
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.24;
-
```
flanagansteve, line 3: nit: forge lint will get mad if you don't restore this newline

contracts/PanopticPool.sol
```diff
@@ -494,82 +493,71 @@ contract PanopticPool is Multicall {
...
+    /// @param usePremiaAsCollateral Whether to compute accumulated premia for all legs held by the user for collateral (true), or just owed premia for long legs (false)
+    function dispatch(
         TokenId[] calldata positionIdList,
-        uint128 positionSize,
-        uint64 effectiveLiquidityLimitX32,
-        int24 tickLimitLow,
-        int24 tickLimitHigh,
+        TokenId[] calldata finalPositionIdList,
+        uint128[] calldata positionSizes,
+        uint64[] calldata effectiveLiquidityLimitsX32,
```
flanagansteve, line 509: Personally, I've had my eye on removing this variable for a long time. It literally only provides 1 block of protection - someone could see your purchase and just wait 1 block to then yank their liquidity out. If we're going to keep it, we should maybe consider a way to actually enforce it over the lifetime of the position instead of only at position creation.

contracts/PanopticPool.sol
```diff
@@ -494,82 +493,71 @@ contract PanopticPool is Multicall {
...
+    /// @notice Mints or burns each `tokenId` in `positionIdList.
+    /// @param positionIdList The list of tokenIds for the option positions to be minted or burnt
+    /// @param finalPositionIdList The final positionIdList after all the tokens have been minted/burnt
+    /// @param positionSizes The list of positionSize for the position to be minted (0 for burns)
+    /// @param effectiveLiquidityLimitsX32 Maximum amount of "spread" defined as `removedLiquidity/netLiquidity` for a new position and
     /// denominated as X32 = (`ratioLimit * 2^32`)
-    /// @param tickLimitLow The lower bound of an acceptable open interval for the ending price
-    /// @param tickLimitHigh The upper bound of an acceptable open interval for the ending price
-    function mintOptions(
+    /// @param tickLimits The lower and lower bounds of an acceptable open interval for the ending price
```
flanagansteve:
```suggestion
    /// @param tickLimits The lower and upper bounds of an acceptable open interval for the ending price
```

contracts/PanopticPool.sol
```diff
@@ -494,82 +493,71 @@ contract PanopticPool is Multicall {
...
-    function burnOptions(
-        TokenId tokenId,
-        TokenId[] calldata newPositionIdList,
-        int24 tickLimitLow,
-        int24 tickLimitHigh,
-        bool usePremiaAsCollateral
-    ) external {
-        _burnOptions(COMMIT_LONG_SETTLED, tokenId, msg.sender, tickLimitLow, tickLimitHigh);
+            // make sure the tokenId is for this Panoptic pool
+            if (tokenId.poolId() != SFPM.getPoolId(address(s_univ3pool)))
```
flanagansteve, line 519: I need to check, but I am pretty sure the SFPM already does this check - if so, I would remove it here and just let the SFPM handle this check + do its own revert

contracts/PanopticPool.sol
```diff
@@ -494,82 +493,71 @@ contract PanopticPool is Multicall {
...
+                revert Errors.InvalidTokenIdParameter(0);
 
-        uint256 medianData = _validateSolvency(
-            msg.sender,
-            newPositionIdList,
-            NO_BUFFER,
-            usePremiaAsCollateral
-        );
+            PositionBalance positionBalanceData = s_positionBalance[msg.sender][tokenId];
```
flanagansteve, line 522:
```suggestion
            PositionBalance positionBalanceData = s_positionBalance[msg.sender][i];
```

contracts/PanopticPool.sol
```diff
@@ -494,82 +493,71 @@ contract PanopticPool is Multicall {
...
 
-        // Update `s_miniMedian` with a new observation if the last observation is old enough (returned medianData is nonzero)
-        if (medianData != 0) s_miniMedian = medianData;
-    }
+            int24 tickLimitLow = tickLimits[i][0];
```
flanagansteve, line 524: if tickLimits is just an array of 2 element arrays, we should probably just make it a single array of `LeftRightSigned`s instead

contracts/PanopticPool.sol
```diff
@@ -494,82 +493,71 @@ contract PanopticPool is Multicall {
...
-        bool usePremiaAsCollateral
-    ) external {
-        _burnAllOptionsFrom(
-            msg.sender,
-            tickLimitLow,
-            tickLimitHigh,
-            COMMIT_LONG_SETTLED,
-            positionIdList
-        );
+            if (PositionBalance.unwrap(positionBalanceData) == 0) {
```
flanagansteve, line 532:
```suggestion
            if (PositionBalance.unwrap(positionBalanceData) != 0) {
```

contracts/PanopticPool.sol
```diff
@@ -582,177 +570,113 @@ contract PanopticPool is Multicall {
...
-    /// @param positionIdList The list of currently held positions by the user, where the newly minted position(token) will be the last element in `positionIdList`
+    /// @param tokenId The tokenId of the newly minted position
     /// @param positionSize The size of the position to be minted, expressed in terms of the asset
     /// @param effectiveLiquidityLimitX32 Maximum amount of "spread" defined as `removedLiquidity/netLiquidity` for a new position and
     /// denominated as X32 = (`ratioLimit * 2^32`)
+    /// @param owner The owner of the option position to be minted
     /// @param tickLimitLow The lower bound of an acceptable open interval for the ending price
     /// @param tickLimitHigh The upper bound of an acceptable open interval for the ending price
-    /// @param usePremiaAsCollateral Whether to compute accumulated premia for all legs held by the user for collateral (true), or just owed premia for long legs (false)
     function _mintOptions(
```
flanagansteve, line 581:
```suggestion
    function _mintOption(
```

Since this method only mints a singular option, it should be named to reflect that

contracts/PanopticPool.sol
```diff
@@ -794,31 +720,58 @@ contract PanopticPool is Multicall {
...
     /// @param tokenId The option position to burn
-    /// @param owner The owner of the option position to be burned
+    /// @param positionSize The size of the position to burn
     /// @param tickLimitLow The lower bound of an acceptable open interval for the ending price on each option close
     /// @param tickLimitHigh The upper bound of an acceptable open interval for the ending price on each option close
+    /// @param owner The owner of the option position to be burned
     /// @param commitLongSettled Whether to commit the long premium that will be settled to storage (disabled during liquidations)
     /// @return paidAmounts The net amount of tokens paid after closing the position
     /// @return premiaByLeg The amount of premia settled by the user for each leg of the position
     function _burnOptions(
```
flanagansteve, line 732:
```suggestion
    function _burnOption(
```

Same point here - this method only ever burns a single option, not options

contracts/PanopticPool.sol
```diff
@@ -582,177 +570,113 @@ contract PanopticPool is Multicall {
...
-            .mintTokenizedPosition(tokenId, positionSize, tickLimitLow, tickLimitHigh);
-
-        _updateSettlementPostMint(
-            tokenId,
-            collectedByLeg,
+        PositionBalance balanceData = PositionBalanceLibrary.storeBalanceData(
             positionSize,
-            effectiveLiquidityLimitX32
+            poolUtilizations,
+            tickData
```
flanagansteve, line 619: You're passing in an empty tickData value here - all you do is:

```solidity
uint96 tickData;
```

contracts/PanopticPool.sol
```diff
@@ -582,177 +570,113 @@ contract PanopticPool is Multicall {
...
     /// @return The total amount of commissions (base rate + ITM spread) paid for token0 (right) and token1 (left)
     function _payCommissionAndWriteData(
         TokenId tokenId,
         uint128 positionSize,
-        LeftRightSigned totalSwapped,
-        bool isCovered
-    ) internal returns (uint32, LeftRightUnsigned) {
+        address owner,
+        LeftRightSigned totalSwapped
+    ) internal returns (uint32, LeftRightUnsigned, LeftRightSigned) {
```
flanagansteve, line 642: Need docstring that you now return paidAmounts here too

contracts/PanopticPool.sol
```diff
@@ -902,98 +855,42 @@ contract PanopticPool is Multicall {
...
-    /// @notice Liquidates a distressed account. Will burn all positions and issue a bonus to the liquidator.
-    /// @dev Will revert if liquidated account is solvent at one of the oracle ticks or if TWAP tick is too far away from the current tick.
-    /// @param positionIdListLiquidator List of positions owned by the liquidator
-    /// @param liquidatee Address of the distressed account
-    /// @param positionIdList List of positions owned by the user. Written as `[tokenId1, tokenId2, ...]`
-    function liquidate(
-        TokenId[] calldata positionIdListLiquidator,
-        address liquidatee,
-        TokenId[] calldata positionIdList
+    function dispatchFrom(
```
flanagansteve, line 893: Need docstring

contracts/PanopticPool.sol
```diff
@@ -1014,17 +911,107 @@ contract PanopticPool is Multicall {
...
             );
+            numberOfTicks = atTicks.length;
+        }
+        {
+            // if account is solvent at all ticks, this is a force exercise or a settleLongPremium.
+            if (solvent == numberOfTicks) {
+                uint256 toLength = positionIdListTo.length;
+                uint256 finalLength = positionIdListToFinal.length;
+                unchecked {
+                    tokenId = positionIdListTo[toLength - 1];
```
flanagansteve, line 948: you should only populate this `tokenId` memory var in this branch:

```solidity
else if (toLength == (finalLength + 1)) {
                        // final is one element shorter, that's a force exercis
```

as its only used there

contracts/PanopticPool.sol
```diff
@@ -1014,17 +911,107 @@ contract PanopticPool is Multicall {
...
+                                revert Errors.InputListFail();
+                            }
+                        }
+                        path = 0;
+                    } else if (toLength == (finalLength + 1)) {
+                        // final is one element shorter, that's a force exercise
+                        tokenId.validateIsExercisable(twapTick);
+                        path = 1;
+                    } else if (finalLength == 0) {
+                        // if final length was zero, this was intended to be liquidaton, but revert because not marging called
```
flanagansteve:
```suggestion
                        // if final length was zero, this was intended to be liquidation, but revert because not margin called
```

flanagansteve: and in fact, better wording might be: "but revert because the liquidatee was only insolvent at _some_ ticks, not all", because it is true that solvent != numberOfTicks. I would also use a clearer error message as some would-be liquidators might be confused about a liquidation not going through when their target is in fact insolvent at some ticks

contracts/PanopticPool.sol
```diff
@@ -1014,17 +911,107 @@ contract PanopticPool is Multicall {
...
+                }
+
+                // if path = 1, force exercise
+
+                if (path == 1) {
+                    // to be eligible for force exercise, the price *must* be outside the position's range for at least 1 leg
+                    exchangedAmounts = _forceExercise(account, tokenId, twapTick, currentTick);
+                }
+
+                // ensure the callee is still solvent after the operation
```
flanagansteve: I'm not sure we should do this - it seems to me like i should be able to force exercise a counterparty even if the closure of that position pushes them into insolvency. Also, I can't think of a scenario where that would happen anyways - being force exercised can only ever increase your collateral balance + reduce your collateral requirement

contracts/PanopticPool.sol
```diff
@@ -1014,17 +911,107 @@ contract PanopticPool is Multicall {
...
+            } else {
+                // if path = 0, settleLongPremium
+                if (path == 0) {
+                    exchangedAmounts = _settleLongPremium(account, tokenId, twapTick, currentTick);
+                }
+
+                // if path = 1, force exercise
+
+                if (path == 1) {
+                    // to be eligible for force exercise, the price *must* be outside the position's range for at least 1 leg
```
flanagansteve: I assume you wanted to keep the scope of this PR down and wanted to avoid introducing the "Force exercisable at all prices" + "partial liquidation" improvements?

contracts/PanopticPool.sol
```diff
@@ -1014,17 +911,107 @@ contract PanopticPool is Multicall {
...
+                    if (toLength == finalLength) {
+                        // same length, that's a settle
+                        {
+                            bytes32 toHash = keccak256(abi.encodePacked(positionIdListTo));
+                            bytes32 finalHash = keccak256(abi.encodePacked(positionIdListToFinal));
+                            if (toHash != finalHash) {
+                                revert Errors.InputListFail();
+                            }
+                        }
+                        path = 0;
```
flanagansteve: We should drop this `path` variable and instead just make the actual calls that correspond to each in the appropriate `if` branch - e.g. actually make this call here:

```solidity
exchangedAmounts = _settleLongPremium(account, tokenId, twapTick, currentTick);
```

contracts/PanopticPool.sol
flanagansteve, line 893: If we wanted to go farther, it could be useful to let this one include more than 1 action per call like we did with `dispatch` above - e.g. execute multiple settleLongPremiums in one call, followed by a forceExercise at the end

Or, compose between `dispatch` and `dispatchFrom` calls - maybe i want to atomically force exercise a buyer and then close a position in one call

Just a half-baked thought for now - obviously you can kind of accomplish this with regular `.multicall()`, but maybe there's some usefulness in not forcing solvency checks in between (beyond just gas savings)

contracts/PanopticPool.sol
```diff
@@ -1286,8 +1328,9 @@ contract PanopticPool is Multicall {
             }
         }
 
-        if (expectedSolvent && solvent != numberOfTicks) revert Errors.AccountInsolvent();
-        if (!expectedSolvent && solvent != 0) revert Errors.NotMarginCalled();
+        return solvent;
+        //if (expectedSolvent && solvent != numberOfTicks) revert Errors.AccountInsolvent();
```
flanagansteve:
```suggestion
```

contracts/PanopticPool.sol
```diff
@@ -1286,8 +1328,9 @@ contract PanopticPool is Multicall {
...
+        //if (!expectedSolvent && solvent != 0) revert Errors.NotMarginCalled();
```
flanagansteve:
```suggestion
```

contracts/CollateralTracker.sol
guil-lambert: So the `s_ITMSpreadFee` is a fee that was only applied to the amount of in-the-money funds that were swapped when minting a new option. The value of that ITM amount is given by the intrinsicValue:

```
int256 intrinsicValue = int256(swappedAmount) - (shortAmount - longAmount);
```

This commission is still there and hasn't changed. It is computed on the total amount of funds moved, regardless of whether those funds are for buying or selling options:

```
uint256(uint128(shortAmount + longAmount)) * COMMISSION_FEE
```


The part I removed is this one


```
-            ) +
-                (
-                    intrinsicValue == 0 || isCovered
-                        ? 0
-                        : Math.unsafeDivRoundingUp(
-                            s_ITMSpreadFee * uint256(Math.abs(intrinsicValue)),
-                            DECIMALS * 100
-                        )
-                );
```

Which added an extra fee based on the intrinsic value. It used to be the swap fee (hence the weird division by 100 from Uniswap's logic). 

We're removing it because it doesn't add much friction (the user already pays for slippage and the swap fee, which is embedded in `swappedAmount`)

contracts/PanopticPool.sol
guil-lambert: Interesting suggestion! I agree it can easily be defeated by yanking the liquidity in the next block. And this is more like a slippage tolerance, so I it could be removed from here and instead moved to the frontend.

Enforcing it on the lifetime of the position might be difficult though: someone could buy an option and say "I want no more than 1% of that liquidity to be removed", preventing buying forever. And each buyer's preference would need to be considered whenever a new position is created.

We could re-purpose that value into something like: "what is the maximum I am willing to pay for that option over its lifetime"

And if the premia accumulated exceeds that value, the position can be forced-closed at no cost by anyone, regardless of their in-range or out-of-range status? Or maybe the force-exercise is even giving a bonus to the exercisor to incentivize rapid unwinding of positions that exceed that threshold.

Would replicate a more common "pay upfront" system for option buyers, and leave them with an external keeper system that would handle their position for them.

contracts/PanopticPool.sol
guil-lambert: This was in the old _mintOption workflow:

```
        // make sure the tokenId is for this Panoptic pool
        if (tokenId.poolId() != SFPM.getPoolId(address(s_univ3pool)))
            revert Errors.InvalidTokenIdParameter(0);
```

And we need it to ensure someone from the `ETH-PEPE-100bps` pool doesn't supply a tokenId in the `ETH-USDC-5bps` pool

contracts/PanopticPool.sol
guil-lambert: No, it should be `tokenId`, the `s_positionBalance` mapping has this structure:

```
    mapping(address account => mapping(TokenId tokenId => PositionBalance positionBalance))
        internal s_positionBalance;
        ```

File: contracts/PanopticPool.sol (hunk 8 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol:524]: It's making the input curation a bit harder, say someone on abi.ninja wants to rescue their position because our UI is dead, it is much easier to supply this:

```
[[887272, -887272]]
```

instead of this:
```
[0xfffffffffffffffffffffffffff27618000000000000000000000000000d89e8]
```

File: contracts/PanopticPool.sol (hunk 9 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol:532]: Pretty sure that should be `==0`

If `positionBalanceData` is zero, that means this position has not been minted yet, so that branch would mint it using `_mintOptions`

If `positionBalanceData` is NOT zero, that means this position has been minted, so that fallback branch would burn it using `_burnOptions`

File: contracts/PanopticPool.sol (hunk 10 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol:581]: The rationale was that a multi-legged `tokenId` would indeed mint several options. Like an Iron Condor with 4 legs is minting 4 options instead of 1

File: contracts/PanopticPool.sol (hunk 12 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol:619]: Yeah, this is because there is no need to store that anymore (I think), but that `storeBalanceData` still require it.

I kept this so that we could update it later, and maybe this is where the "what is the maximum I am willing to pay for that option over its lifetime" information gets tucked in

File: contracts/PanopticPool.sol (hunk 13 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol:642]: Done!

File: contracts/PanopticPool.sol (hunk 11 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol:732]: See point above

File: contracts/PanopticPool.sol (hunk 14 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol:893]: Done!

Comment (guil-lambert) [reply contracts/PanopticPool.sol:893]: Yes, the rationale was that this could be accomplished using `multicall`, and using `multicall` wouldn't necessarily benefit from the gas savings because more operations must check for that specific flow to validate solvency at the end.

File: contracts/PanopticPool.sol (hunk 15 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol:948]: The `settleLongPremium` also needs that `tokenId`, so declaring it here handles both `forceExercise` and `settleLongPremium`

File: contracts/PanopticPool.sol (hunk 19 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol]: Great suggestion, done!

File: contracts/PanopticPool.sol (hunk 16 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol]: Done!

File: contracts/PanopticPool.sol (hunk 18 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol]: Yes, that'll be in a follow-up PR

File: contracts/PanopticPool.sol (hunk 17 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol]: Good point.

I think a scenario is that exercisors can manipulate the price an/or the premia to force the user to execute at a bad price.

But we're now making sure the account is solvent before the operations (which was not the case before), so I'll have to think about whether those checks are necessary --ie. if they are indeed solvent at the beginning, could they really be made insolvent after closing?

File: contracts/PanopticPool.sol (hunk 20 above)
Comment (guil-lambert) [reply contracts/PanopticPool.sol]: Done!

File: contracts/PanopticPool.sol (hunk 10 above)
Comment (flanagansteve) [reply contracts/PanopticPool.sol:581]: Gotcha, that seems valid!

File: contracts/CollateralTracker.sol (hunk 2 above)
Comment (flanagansteve) [reply contracts/CollateralTracker.sol:116]: Ah! I was misreading the diff in Github's UI, my bad - get's confusing reviewing branches-off-of-branches 😅 

Your PR _restores_ s_ITMSpreadFee after `feat/long-premia-only-at-mint` _removed_ it, that's all

File: contracts/PanopticPool.sol (hunk 8 above)
Comment (flanagansteve) [reply contracts/PanopticPool.sol:524]: Personally, I think that is a very small benefit that affects less than 1% of users, and the tradeoff is increased gas costs for 100% of users. Those users who are already comfortable going to ABI Ninja and constructing raw contract calls won't be scared off by an additional step of converting tick values to a packed uint256, especially if we explain how to do this in the docs / list this value as a copy-pastable value for the widest-possible limit:

```solidity
[0xfffffffffffffffffffffffffff27618000000000000000000000000000d89e8]
```

Especially considering that these users will need to construct these values as well, which will also require some technical knowledge:

```solidity
        TokenId[] calldata positionIdList,
        TokenId[] calldata finalPositionIdList,
        uint128[] calldata positionSizes,
        uint64[] calldata effectiveLiquidityLimitsX32,
```

contracts/PanopticPool.sol
flanagansteve: You are right! My mistake

contracts/PanopticPool.sol
flanagansteve: You could just hardcode 0 as the value here though and then delete the `tickData` declaration

```suggestion
            0
```

If it's a matter of recordkeeping we could leave a TODO comment too:

```suggestion
            0 // TODO: Perhaps we put the capped premia argument here
```

contracts/PanopticPool.sol
flanagansteve: I don't think you did - I mean that you need an @return docstring, like the one that exists for the returned `LeftRightUnsigned`:

```solidity
/// @return The total amount of commissions (base rate + ITM spread) paid for token0 (right) and token1 (left)
```

Perhaps:

```solidity
/// @return The net amount of tokens paid after minting each position
```

edit: yes you did!

contracts/PanopticPool.sol
flanagansteve: I don't think you did? I don't see one here

edit: Yes you did!

contracts/PanopticPool.sol
flanagansteve: I don't see it - maybe you did this + some other changes locally and need to push?

edit: this was done

contracts/PanopticPool.sol
flanagansteve: I don't see this (might be local on your end)

edit: yes it was done

contracts/PanopticPool.sol
flanagansteve: Ah - makes sense - if you take the other suggestion of killing the `path` var and putting the logic in each if-else then that would become clearer to see too

contracts/PanopticPool.sol
flanagansteve: > ie. if they are indeed solvent at the beginning, could they really be made insolvent after closing?

Yeah to me this is the key point - I don't think so, but we can discuss live and talk through each possible scenario if you like

contracts/PanopticPool.sol
flanagansteve: these two done locally?

contracts/PanopticPool.sol
flanagansteve: note to self: will push a commit with this change once i ensure it doesn't mess up Guillaume's flow

contracts/PanopticPool.sol
flanagansteve: (This is going to be a separate PR 🚀 )

contracts/PanopticPool.sol
flanagansteve: note to self: will push a commit with this change once i ensure it doesn't mess up Guillaume's flow

contracts/PanopticPool.sol
flanagansteve: > using multicall wouldn't necessarily benefit from the gas savings because more operations must check for that specific flow to validate solvency at the end.

I was saying a dispatchFrom that supports multiple aggregated actions would be cheaper than multicalling the same actions as separate function calls - one solvency check at beginning and end instead of between each action. I might be misunderstanding your point here, though:

> because more operations must check for that specific flow to validate solvency at the end.

Either way, this is a minor point though - I still can't think of a reason beyond gas savings and don't think that's worth the significant complexity to make dispatchFrom support multiple operations per call

contracts/PanopticPool.sol
flanagansteve: Yes they were

contracts/PanopticPool.sol
flanagansteve: Perfect - now tokenId is used in the two `if` branches that follow its declaration - all set here

contracts/PanopticPool.sol
flanagansteve: > And we need it to ensure someone from the ETH-PEPE-100bps pool doesn't supply a tokenId in the ETH-USDC-5bps pool

My point was that the SFPM might already perform this exact check, and that both the old and new _mintOptions may be being redundant by re-performing it here

However, surprisingly, the SFPM does _not_ appear to do this check - so instead, I would suggest we move that check there, so that both PanopticPool calls and direct SFPM calls have this protection

Doesn't have to be in this PR, but curious for your thoughts - and if you agree we should do it we can record the task somewhere (either your spreadsheet or the kanban)

contracts/PanopticPool.sol
```diff
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.24;
-
 // Interfaces
```
flanagansteve, line 4:
```suggestion

// Interfaces
```

PR #168, opened 2025-10-01
feat: protocol fee

flanagansteve:
- Charge it in each CT's `takeCommissionAddData`
- Allow a designated `FEE_SWITCH_CONTROLLER` account to update the fee and the recipient
- Set that FEE_SWITCH_CONTROLLER, as well as an initial fee and recipient, in `deployNewPool`
- Emit the commission charged in `OptionMinted`

TODO: update tests (or maybe add tests?)


flanagansteve:
Superceded by #170 170

PR #170, opened 2025-10-02
feat: protocol fee v2

flanagansteve:
- Charge it in each CT's `takeCommissionAddData`
- Allow a designated `FEE_SWITCH_CONTROLLER` account to update the fee and the recipient
- Set that FEE_SWITCH_CONTROLLER, as well as an initial fee and recipient, in `deployNewPool`
- Emit the commission charged in `OptionMinted`

Now off of the `dev` branch, where i get the unifying of `exercise` and `takeCommissionAddData`'s duplicated logic for free. Still need to fight stack too deep, however.

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -77,12 +77,14 @@ contract PanopticPool is Multicall {
     /// @param recipient User that minted the option
     /// @param tokenId TokenId of the created option
     /// @param balanceData The `PositionBalance` data for `tokenId` containing the number of contracts, pool utilizations, and ticks at mint
-    /// @param commissions The total amount of commissions (base rate + ITM spread) paid for token0 (right) and token1 (left)
+    /// @param plpCommissions The total amount of commissions (base rate + ITM spread) paid for token0 (right) and token1 (left) to PLPs
+    /// @param protocolCommissions The total amount of commissions paid for token0 (right) and token1 (left) to the protocol
     event OptionMinted(
         address indexed recipient,
         TokenId indexed tokenId,
         PositionBalance balanceData,
-        LeftRightUnsigned commissions
+        LeftRightUnsigned plpCommissions
+        LeftRightUnsigned protocolCommissions
     );
 
     /*//////////////////////////////////////////////////////////////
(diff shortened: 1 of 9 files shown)


contracts/PanopticPool.sol
```diff
@@ -631,53 +634,57 @@ contract PanopticPool is Multicall {
...
         uint128 positionSize,
         address owner,
         LeftRightSigned totalSwapped
-    ) internal returns (uint32, LeftRightUnsigned, LeftRightSigned) {
+    ) internal returns (uint32, LeftRightUnsigned, LeftRightUnsigned, LeftRightSigned) {
         // compute how much of tokenId is long and short positions
         (LeftRightSigned longAmounts, LeftRightSigned shortAmounts) = PanopticMath
             .computeExercisedAmounts(tokenId, positionSize);
 
-        uint32 commissions;
```
flanagansteve, line 646: I was pretty surprised that `commissions`, which formerly represented the plpCommissions alone, was a uint32 - it could definitely exceed that for 18-decimal tokens. I also noticed we were returning this value in the first return slot, but then consuming it as `utilizations` above, so i assume this was just a mistake - i took the liberty of popping `utilizations1 back in there and making plpCommissions a LeftRightUnsigned

contracts/PanopticPool.sol
```diff
@@ -620,7 +623,7 @@ contract PanopticPool is Multicall {
 
         s_positionBalance[owner][tokenId] = balanceData;
```
flanagansteve, line 623: This is the line that gives me stack too deep - not sure if that actually tells us anything useful, but i'll start out by trying to reduce the number of variables we read in `_mintOptions`

contracts/PanopticPool.sol
```diff
@@ -609,18 +612,17 @@ contract PanopticPool is Multicall {
...
             positionSize,
             poolUtilizations,
-            tickData
+            0
         );
 
         s_positionBalance[owner][tokenId] = balanceData;
 
-        emit OptionMinted(owner, tokenId, balanceData, commissions);
+        emit OptionMinted(owner, tokenId, balanceData, plpCommissions, protocolCommissions);
```
flanagansteve, line 625: TODO: Move this to be emitted from the CT

PR #171, opened 2025-10-03
feat9: Decouple Risk Logic into a Modular RiskEngine Contract

guil-lambert:
#### Summary

This PR introduces a major architectural refactor that decouples risk management logic from the `CollateralTracker` and `PanopticMath` contracts into a new, dedicated `RiskEngine.sol` contract.

The primary motivation is to make the protocol's risk parameterization modular and extensible. By treating the risk model as a swappable component, this change enables a significant new feature: the ability for the `PanopticFactory` to deploy multiple, distinct `PanopticPool`s for the same underlying Uniswap V3 pool, each configured with a different `RiskEngine`.

#### Key Changes

* **New `RiskEngine.sol` Contract:**
    * A new contract has been created to centralize all logic for risk management.
    * All `immutable` risk parameters (`SELLER_COLLATERAL_RATIO`, `TARGET_POOL_UTIL`, etc.) have been moved from `CollateralTracker` to `RiskEngine`.
    * All collateral calculation functions (`getMargin`, `exerciseCost`, `getLiquidationBonus`, `_getTotalRequiredCollateral`, etc.) have been migrated from `CollateralTracker` and `PanopticMath` into this new contract.

* **`PanopticFactory` Enhancement:**
    * The factory's pool mapping has been updated to `mapping(IUniswapV3Pool => mapping(RiskEngine => PanopticPool))`.
    * `deployNewPool` now takes a `RiskEngine` address as an argument, making it a key component for identifying and creating a unique `PanopticPool`.
    * The `CREATE2` salt calculation now includes the `RiskEngine` address to ensure deterministic and unique addresses for pools with different risk models.

* **`PanopticPool` Integration:**
    * `PanopticPool` now holds an `immutable` reference to its `s_riskEngine`.
    * All calls related to margin checks, liquidations, and exercises are now delegated to the associated `RiskEngine` contract.

* **`CollateralTracker` Simplification:**
    * The `CollateralTracker` contract has been significantly simplified, removing all complex risk calculations. Its role is now focused on being an ERC-4626-like vault for asset tracking and interest accrual.

* **Increased Precision:**
    * The `DECIMALS` constant used for risk calculations has been increased from `10,000` to `10,000,000`, allowing for more granular risk parameter settings.

#### New Functionality Unlocked

This refactor enables the creation of parallel markets for the same asset pair with varying risk profiles. For example, for a single WETH/USDC Uniswap pool, we can now deploy:
1.  A "Conservative" `PanopticPool` linked to a `RiskEngine` with high collateral requirements.
2.  A "Degenerate" `PanopticPool` linked to a different `RiskEngine` with lower collateral requirements and higher leverage.

#### Testing

* All existing unit and integration tests have been updated to align with the new architecture.
* New tests have been added to `PanopticFactory.t.sol` to specifically validate:
    * The ability to deploy multiple pools for the same Uniswap pair using different `RiskEngine` contracts.
    * Rejection of deployments that use a zero-address for the `RiskEngine`.

## Summary by CodeRabbit

- New Features
  - Compound-interest accrual via a global borrow index with on-demand accrual and new views for interest and redeemable assets.
  - Option to use premia as collateral; withdrawals/mints auto-accrue interest and enforce minimum redemption.
  - Risk Engine integration: pools expose their risk engine; multiple pools can exist per Uniswap pair using different risk engines.

- Breaking Changes
  - Updated constructor and withdrawal parameters.
  - Factory APIs/events now include a risk engine; pool retrieval requires specifying it.

- Error Handling
  - Added BelowMinimumRedemption and ZeroAddress; renamed StaleTWAP to StaleOracle.


contracts/PanopticFactory.sol
```diff
@@ -69,7 +71,8 @@ contract PanopticFactory is FactoryNFT, Multicall {
     //////////////////////////////////////////////////////////////*/
 
     /// @notice Mapping from address(UniswapV3Pool) to address(PanopticPool) that stores the address of all deployed Panoptic Pools.
-    mapping(IUniswapV3Pool univ3pool => PanopticPool panopticPool) internal s_getPanopticPool;
+    mapping(IUniswapV3Pool univ3pool => mapping(RiskEngine riskEngine => PanopticPool panopticPool))
```
flanagansteve, line 74: I'm going to try and be a devil's advocate here:

You have made each PanopticPool 1 to 1 with a {UniPool, RiskEngine} combination. The same UniPool may correspond to many PanopticPools, one for each RiskEngine out there.

Do you actually foresee many RiskEngines coexisting at once, if our ecosystem is healthy? Why would we not instead expect 1 RiskEngine per PanopticFactory? The latter would still support:

- Changing the RiskEngine - the RiskEngine contract itself can be upgradeable
- Bespoke behaviour for different markets (e.g., allow greater capital efficiency in stable-stable pools) - you would implement this as business logic within a getAccountMarginDetails call. E.g. `if stablecoin[ct0] && stablecoin[ct1] return 20x leverage, otherwise return 10x.`

But maybe I'm missing some key reason one UniPool might need to have multiple PanopticPools with independent RiskEngine contracts.

Maybe one thing you were considering is: "What if in 3 months we decide we want to treat a type of asset class or Uniswap pair differently in the RiskEngine? Wouldn't a more transparent flow for endusers be: Panoptic writes new RiskEngine for that usecase, and deploys a new PanopticPool that uses that RiskEngine?"

But when I think about this, it feels just as transparent to our users for us to write a diff to `getAccountMarginDetails`, and push the update through normal contract-upgrade devops. And on the business side, getting users to switch from the old PanopticPool to the new one seems like it could fail, just like the v3 to v4 upgrade on Uniswap has failed. And then we're stuck maintaining double work - our subgraph needs to play nice with both of the live RiskEngines, our UI might need to have a conditional switch in each transaction method, etc.

The tradeoff of course is that updating the getAccountMarginDetails to handle the new asset class will also risk screwing the update up and affecting unrelated PanopticPools. And more generally, give us admin risk for the upgrade keys. Which is why I only consider myself a devil's advocate for this, rather than a real advocate. But want to hear your counterargument here - I think risk of update-screwups & admin key leak can be straightforwardly mitigated.

contracts/RiskEngine.sol
```diff
@@ -0,0 +1,1001 @@
...
+    /// @param fees If applicable, fees to debit from caller (rightSlot = currency0 left = currency1), 0 for `settleLongPremium`
+    /// @param atTick The tick at which to convert between currency0/currency1 when redistributing the surplus tokens
+    /// @param ct0 The collateral tracker for currency0
+    /// @param ct1 The collateral tracker for currency1
+    /// @return The LeftRight-packed deltas for currency0/currency1 to move from the caller to the payor
+    function getRefundAmounts(
+        address payor,
+        LeftRightSigned fees,
+        int24 atTick,
+        CollateralTracker ct0,
```
flanagansteve, line 111: You could consider, instead of passing in ct0 and ct1, passing in:

- ctSharePrice
- payorBalance
- ctTotalAssets

Which would then let you skip a bunch of read calls in here and save gas - convertToShares/Assets == multiplying or dividing by sharePrice, totalSupply = totalAssets / sharePrice, etc.

Could keep it from being too many arguments by packing into LeftRightUnsigned as appropriate

contracts/RiskEngine.sol
```diff
@@ -0,0 +1,1001 @@
...
+    /// @notice Compute the pre-haircut liquidation bonuses to be paid to the liquidator and the protocol loss caused by the liquidation (pre-haircut).
+    /// @param tokenData0 LeftRight encoded word with balance of token0 in the right slot, and required balance in left slot
+    /// @param tokenData1 LeftRight encoded word with balance of token1 in the right slot, and required balance in left slot
+    /// @param atSqrtPriceX96 The oracle price used to swap tokens between the liquidator/liquidatee and determine solvency for the liquidatee
+    /// @param netPaid The net amount of tokens paid/received by the liquidatee to close their portfolio of positions
+    /// @param shortPremium Total owed premium (prorated by available settled tokens) across all short legs being liquidated
+    /// @return The LeftRight-packed bonus amounts to be paid to the liquidator for both tokens (may be negative)
+    /// @return The LeftRight-packed protocol loss (pre-haircut) for both tokens, i.e., the delta between the user's starting balance and expended tokens
+    function getLiquidationBonus(
+        LeftRightUnsigned tokenData0,
```
flanagansteve, line 295: Trying really hard to avoid naming/docstring/gas stuff, but the one I'll allow myself: Let's find a specific name here. `collateralInfo0` would be a little better, but a better noun than "info", which fits both balance & requirement conceptually into it, would be better - will mull on it

contracts/CollateralTracker.sol
```diff
@@ -71,13 +71,12 @@ contract CollateralTracker is ERC20Minimal, Multicall {
     /// @notice Prefix for the token name (i.e POPT-V1 USDC LP on ETH/USDC 30bps).
```
flanagansteve, line 71: This contract now, essentially, tracks:

- Shares that represent the value of deposited collateral
- Accumulated PLP earnings
- Accumulated interest (owed or earned)

And action-wise, handles:

- commissions (to protocol and PLPs)
- deposits + withdrawals + transfers

To me, the name CollateralTracker implies this is the place you query to get the margin balance+requirement for the user. But that has now shifted to the RiskEngine.

This contract is now more of a "earnings, deposits and fees tracker" - your full collateral, including open position value, is found elsewhere

- Little question: Should we consider renaming it, given the above?
- Bigger question: Should we consider slimming it down further? Could the above actually just be tracked in the PanopticPool itself, and we kill the CollateralTracker contract entirely?

I'm being a little facetious with the Bigger Question, but it's worth considering something like it

flanagansteve (commented):
SO PSYCHED ABOUT THIS!

Restricted myself to just higher-level comments here. I didn't inspect the RiskEngine inner function logic too thoroughly - I assume a lot of it is copy-paste from the CollateralTracker and PanopticMath methods you are replacing. Let me know if there's anything you _did_ change there, though.

PR #174, opened 2025-10-03
WIP: feat: capped premia

flanagansteve:
- Remove `effectiveLiquidityLimitX32`
- Specify the total token0 / token1 you are willing to let your long legs owe (in sum) for your position in mintOptions, replacing effectiveLiquidityLimitX32
- Store it
- Expose a exercisePositionExceedingPremiaCap, which when called will: a) close any position that has exceeded its premia cap in either token0 or token1, and b) 

Notes:

- I decided that it's preferable to have a window where the sellers don't suffer a penalty, but the position is still force-closeable for free. The way i see this working is that the UI shows a cap of 1 ETH, in the contracts we pass that in as 0.99 ETH, and sellers have the 0.99 to 1 ETH window as their chance to close the position without penalty.

Design decisions:

- Should the cap be an either/or cap on each token amount, or a cross-balance cap? I kind of think cross-balance, but doing it either-or was easier for a first pass
- Should we try to fit the cap into the TokenId, so that we don't have to store it?

TODO:

- Implement the fancier reward/penalty mechanism - caller gets proportionally more as we exceed the cap more severely - in this version, it's just: No penalty if exercised within 1% of cap, and 1% penalty otherwise.
- test

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1083,7 +1083,7 @@ contract PanopticPool is Multicall {
             // It is kind of funky that if liquidatable account owned a position was in the grace period window, third parties have the option
             // of claiming that cap-accumulated bonus there first, then liquidating the account, and getting a different sum bonus
             // than if they had just liquidated
-            (shortPremium, longPremium, , positionBalanceArray) = _calculateAccumulatedPremia(
+            (shortPremium, longPremium, , positionBalanceArray,) = _calculateAccumulatedPremia(
                 liquidatee,
                 positionIdList,
                 COMPUTE_PREMIA_AS_COLLATERAL,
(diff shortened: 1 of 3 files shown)


contracts/PanopticPool.sol
```diff
@@ -1775,6 +1794,143 @@ contract PanopticPool is Multicall {
...
+                MIN_SWAP_TICK,
+                MAX_SWAP_TICK,
+                owner,
+                COMMIT_LONG_SETTLED
+            );
+
+            uint128 token0Overage = owedPremia0 - token0Cap;
+            uint128 token1Overage = owedPremia1 - token1Cap;
+
+            // this is equivalent to: if overage >= 1% of cap, pay out a reward
```
flanagansteve: If you're in this grace period, transfer `cap - accumulated` to caller

If you're over cap, transfer premium (ONLY overage, `accumulated - cap`) from sellers to caller

PR #181, opened 2025-10-09
Feat: Implement loans and credit via width=0 legs

guil-lambert:
### Description

This PR introduces a fundamental new primitive to the protocol: **loans and credit**. This is accomplished by allowing option legs to be created with a `width` of 0. This change unlocks the ability to create a wide range of "funded options" and other novel financial products.

A leg with `width=0` represents a fungible claim on the underlying collateral pool, where:

  - A **short `width=0` leg** is a **loan** (borrowing from the pool).
  - A **long `width=0` leg** is a **credit** (lending to the pool).

### Key Changes

  - **`TokenId.sol`**: The validation logic has been updated to permit `width=0`, officially enabling this new type of leg.
  - **`RiskEngine.sol`**:
      - Implemented new collateral logic for standalone `width=0` positions:
          - Loans (short) require collateral of 100% of the notional value plus the base `SELLER_COLLATERAL_RATIO`.
          - Credits (long) have a collateral requirement of 0.
      - The risk partnering logic has been significantly expanded to recognize and provide capital efficiency for composite strategies that combine options and loans/credit, such as:
          - **Funded Options**: Cash-Secured Puts, Prepaid Calls.
          - **Option-Protected Loans**.
          - **Delayed Token Swaps**.
  - **`SemiFungiblePositionManager.sol`**: The `_mintOrBurn` flow now identifies `width=0` legs and treats them as direct token transfers (accounted for in `itmAmounts`), bypassing Uniswap v3 liquidity operations.
  - **`CollateralTracker.sol`**:
      - Now actively tracks the net available liquidity for lending.
      - A new `InsufficientLiquidity` error prevents users from borrowing more tokens than have been supplied by lenders, ensuring solvency.
  - **`PanopticPool.sol`**: Loops that iterate over position legs now correctly skip `width=0` legs when performing actions related to Uniswap V3 liquidity and premium calculations.
  - **`PanopticMath.sol`**: The `getAmountsMoved` function is updated to handle `width=0` legs by temporarily using a minimal width for the sole purpose of calculating their correct notional value.
  - **Testing**: Added an extensive suite of Foundry tests covering the new loan/credit functionality, including collateral calculations, minting/burning, and insufficient liquidity scenarios.


contracts/RiskEngine.sol
```diff
@@ -597,28 +598,37 @@ contract RiskEngine {
...
+                // start with base requirement, which is based on isLong value
+                required += _getRequiredCollateralAtUtilization(
+                    amountMoved,
+                    isLong,
+                    poolUtilization
+                );
 
-        // required collateral is at least 1
-        required = 1;
+                if (isLong == 0) {
```
flanagansteve, line 627: did you intentionally shift this from `unchecked` to `checked`, or was that a mistake?

contracts/RiskEngine.sol
```diff
@@ -597,28 +598,37 @@ contract RiskEngine {
...
+                    // if position is short, check whether the position is out-the-money
 
-        // start with base requirement, which is based on isLong value
-        required += _getRequiredCollateralAtUtilization(amountMoved, isLong, poolUtilization);
+                    (int24 tickLower, int24 tickUpper) = tokenId.asTicks(index);
```
flanagansteve, line 630: it also seems like you also dropped the conditional logic here to calculate one CR for the ITM case, and another for the OTM case, and instead just calculate 3 CRs:

- Return val of _getRequiredCollateralAtUtilization (which was previously used to get CR of a long leg
- r1, which is the max of:
  - amountMoved + (2 * (Return val of _getRequiredCollateralAtUtilization) / sqrtRatio(2 * (atTick - strike))) for long legs, and flipping atTick-strike to strike-AtTick for short legs
  - amountMoved + (2 * (amountsMoved) / sqrtRatio(2 * (atTick - strike))) for long legs, and flipping atTick-strike to strike-AtTick for short legs
- (Return val of _getRequiredCollateralAtUtilization) + (amountsMoved * (scaleFactor - ratio) / (scaleFactor + a constant))
  - where ratio is, similarly to above, sqrtRatio(2 * (atTick - strike)) for long legs, or strike-AtTick for short legs
  - and scale factor is sqrtRatio(position.width)

And take the max of those 3 values. And i think, somehow, the latter two values represent these values in your comments:

> //    - 10% of the notional value at the strike price
> //    - 20% of the underlying price MINUS the out-the-money amount

I am having some trouble understanding: 

- how those latter two values map to the reg-T values
- how taking the max of all 3 is equivalent to the previous behaviour where we had if-else branches that treated short legs VS long legs differently, and treated calls VS puts differently based on their ITM-ness

If this is some cool algebraic simplification of the previous logic, I'm really excited, but want to make sure I understand the equivalency before I sign off on it

contracts/RiskEngine.sol
```diff
@@ -708,24 +702,150 @@ contract RiskEngine {
...
+        // -Upfront short option (short put or call + loan) "Upfront payment to seller" = requirement is max(loan, short option)
+        // -Option-protected loan (long put or call + loan) "Get a loan with an embedded long option for capital protection" = requirement is max(loan, short option)
+        // -Cash-Secured Option (short put or call + credit) "Allocate collateral to that specific option" = requirement is max(short - credit, 1)
+        //
+        // TOKEN TRANSFERS
+        // - Delayed Swap (credit at one strike, loan at another; different amounts = effective swap) = requirement is max(loan0 - convert1to0(credit), 1) or max(loan1 - convert0to1(credit), 1)
+        {
+            // only proceed if the partners have the same asset
+            if (tokenId.asset(partnerIndex) == tokenId.asset(index)) {
+                // witdh of associated legs, true if greater than 0 (ie. it is an option leg)
```
flanagansteve, line 730:
```suggestion
                // width of associated legs, true if greater than 0 (ie. it is an option leg)
```

contracts/libraries/PanopticMath.sol
```diff
@@ -786,6 +785,12 @@ library PanopticMath {
         uint128 amount0;
         uint128 amount1;
 
+        // if the width is zero, add 1 to the width to allow liquidity amounts to be computes
```
flanagansteve, line 788: Just double checking - I don't think we need to modify `haircutPremia` in the same way - we try to get the settled premia based on the chunk key:

```solidity
bytes32 chunkKey = keccak256(
                            abi.encodePacked(
                                tokenId.strike(leg),
                                tokenId.width(leg),
                                tokenId.tokenType(leg)
                            )
                        );

                        // The long premium is not committed to storage during the liquidation, so we add the entire adjusted amount
                        // for the haircut directly to the accumulator
                        settledTokens[chunkKey] = settledTokens[chunkKey].add(
                            (LeftRightSigned.wrap(0).sub(_premiasByLeg[i][leg])).subRect(
                                haircutAmounts
                            )
                        );
```

And i think this will play fine with a width of 0

But just flagging

contracts/libraries/PanopticMath.sol
```diff
@@ -786,6 +785,12 @@ library PanopticMath {
...
+        /// @dev this is just for accounting purposes, the actual tokenId will remain with a width = 0
+        if (tokenId.width(legIndex) == 0) {
+            tokenId = tokenId.addWidth(2, legIndex);
```
flanagansteve, line 791: Technically you could skip this mstore and just replace this sequence:

```solidity
if (tokenId.width(legIndex) == 0) {
            tokenId = tokenId.addWidth(2, legIndex);
        }

        LiquidityChunk liquidityChunk = getLiquidityChunk(tokenId, legIndex, positionSize);
```

with:

```solidity
        LiquidityChunk liquidityChunk = getLiquidityChunk(tokenId.width(legIndex) == 0 ?  tokenId.addWidth(2, legIndex) : tokenId, legIndex, positionSize);
```

contracts/libraries/PanopticMath.sol
flanagansteve: However - on the topic of handling liquidation of accounts holding loan positions - should lenders arguably have higher-prioritised claims when clawing back from a liquidated account than other parties? I think if someone opens a loan and gets liquidated their loans are paid off just like the borrowing associated with any other option leg - but if we're treating this like margin borrowed from your broker, typically that'd have higher priority I think

flanagansteve: One way to accomplish that could be adding a fourth dispatchFrom action, `liquidateLoanLegs`, and require that a victim of regular-`liquidate` has first all their loan legs  `liquidateLoanLegs`ed

contracts/PanopticPool.sol
```diff
@@ -434,45 +433,47 @@ contract PanopticPool is Multicall {
...
-                            LeftRightSigned.unwrap(
-                                LeftRightSigned
-                                    .wrap(int256(LeftRightUnsigned.unwrap(longPremium)))
-                                    .sub(premiaByLeg[leg])
-                            )
-                        )
-                    );
                 }
                 unchecked {
                     ++leg;
```
flanagansteve, line 479: `dispatchFrom`'s `_forceExercise` path will fail for 0-width positions:

```solidity
} else if (toLength == (finalLength + 1)) {
    // final is one element shorter, that's a force exercise
    tokenId.validateIsExercisable(twapTick);
    exchangedAmounts = _forceExercise(account, tokenId, twapTick, currentTick);
}
```

`validateIsExercisable` will incorrectly get 0 for the `rangeUp` and `rangeDown`, and this check will therefore always fail: `if ((currentTick >= _strike + rangeUp) || (currentTick < _strike - rangeDown)) {`

More generally, we should decide if we want to permit the force exercise of loan/credit positions at all

I think probably not - so long as you are solvent you should be permitted to hold the credit / loan as long as you want. If you're in agreement, then the fix here is as simple as adding the same `if (tokenId.width(leg) != 0)` guard in `_forceExercise` that you added in `_settlePremium` and friends.

contracts/RiskEngine.sol
```diff
@@ -597,28 +598,37 @@ contract RiskEngine {
...
 
         uint256 isLong = tokenId.isLong(index);
+        unchecked {
+            // if the width is 0, then this is a loan/credit
+            if (tokenId.width(index) == 0) {
+                if (isLong == 0) {
+                    // buying power requirement for a Loan position is 100% + sellCollateralRatio(utilization = 0)
+                    required = Math.mulDivRoundingUp(
+                        amountMoved,
+                        SELLER_COLLATERAL_RATIO + DECIMALS,
```
flanagansteve, line 608: Just wondering, what is the rationale behind hardcoding a collateralisation requirement of 120%? Is it just convenient to see us reusing SELLER_COLLATERAL_RATIO consistently?

Aave seems to be less conservative than that for ETH:

Screenshot 2025-10-16 at 10 25 00 PM

But moreso for USDC:

Screenshot 2025-10-16 at 10 25 42 PM

(Unless i have these flipped in my head - but 75% LTV == .75/1 = 133% CR, i think, right? whereas ETH ranged from 90->95 == 110% CR?)

Anyway - i could see it being worthwhile to at least have a few better-hardcoded values for the most popular assets, maybe updateable by an admin. Or, going farther, something dynamic - but that's a whole protocol in itself, of course

contracts/CollateralTracker.sol
```diff
@@ -1050,8 +1048,12 @@ contract CollateralTracker is ERC20Minimal, Multicall {
             }
             int128 netBorrows = isCreation ? shortAmount - longAmount : longAmount - shortAmount;
 
+            int256 _inAMM = int256(uint256(s_inAMM));
+
+            if (_inAMM + netBorrows < 0) revert Errors.InsufficientLiquidity();
```
flanagansteve, line 1053: Normally i type these kinds of comments out, delete them when I realise everything is fine, and don't bother making the comment. However, this one made me realise is that you could also take a loan out by doing a long call, which we should maybe document - you phrase it as:

`A short width=0 leg is a loan (borrowing from the pool).`

When its really:

`A short width=0 put, or a long width=0 call, is a loan (borrowing from the pool).`

or maybe, more general wording:

`A bullish width=0 leg is a loan`

And also, the thought process & code snippets might be useful content for our docs regarding why this system actually functions correctly.

---

So just checking: Loans are short positions and borrow from the pool, and should increase netBorrows. Longs are credits and supply capital to the pool, and therefore should decrease netBorrows. When each is closed, the reverse should happen.

On mint, netBorrows is increased by `shortAmount - longAmount`. Those two values ultimately come from:

```
```

Which in our zero-width case would be, from _calculateIOAmounts:

```
LeftRightUnsigned amountsMoved = getAmountsMoved(tokenId, positionSize, legIndex);

        bool isShort = tokenId.isLong(legIndex) == 0;

        if (tokenId.tokenType(legIndex) == 0) {
            if (isShort) {
                // if option is short, increment shorts by contracts
                shorts = LeftRightSigned.wrap(0).addToRightSlot(
                    Math.toInt128(amountsMoved.rightSlot())
                );
            } else {
                // is option is long, increment longs by contracts
                longs = LeftRightSigned.wrap(0).addToRightSlot(
                    Math.toInt128(amountsMoved.rightSlot())
                );
            }
        } else {
            if (isShort) {
                // if option is short, increment shorts by notional
                shorts = LeftRightSigned.wrap(0).addToLeftSlot(
                    Math.toInt128(amountsMoved.leftSlot())
                );
            } else {
                // if option is long, increment longs by notional
                longs = LeftRightSigned.wrap(0).addToLeftSlot(
                    Math.toInt128(amountsMoved.leftSlot())
                );
            }
        }
```

getAmountsMoved has your modification where you pass in a 2-width liquidity chunk and then getAmount0/1 for the liquidity chunk.

What this has revealed to me:

- Short puts would have a shortAmounts of the notionalAmount and a longAmounts of zero. We increment netBorrows by shortAmounts-longAmounts, which means we increment it by the notional value of this leg, which is correct - that's how much got borrowed.
- Long puts would have a longAmounts of the notionalAmount and a shortAmounts of zero, which means we increment by shortAmounts-longAmounts = -longamounts = -notionalValue, which is correct - netBorrows should go down since value has been credited to the pool
- and, netBorrows is signed, so the negative value is fine

Looks like we are all good here. Main thing this made me realise is that you could _also_ take a loan out by doing a long call, which we should maybe document - you phrase it as:

`A short width=0 leg is a loan (borrowing from the pool).`

When its really:

`A short width=0 put, or a long width=0 call, is a loan (borrowing from the pool).`

or maybe, more general wording:

`A bullish width=0 leg is a loan`

contracts/RiskEngine.sol
```diff
@@ -901,6 +1032,107 @@ contract RiskEngine {
         }
     }
 
+    function _computeLoanOptionComposite(
```
flanagansteve, line 1035: @coderabbitai bless us with some docstrings please:

```solidity
    function _computeLoanOptionComposite(
        TokenId tokenId,
        uint128 positionSize,
        uint256 index,
        uint256 partnerIndex,
        int24 atTick,
        int16 poolUtilization
    ) internal view returns (uint256) {
        uint256 _required = _getRequiredCollateralSingleLegNoPartner(
            tokenId,
            index,
            positionSize,
            atTick,
            poolUtilization
        );
        uint256 requiredPartner = _getRequiredCollateralSingleLegNoPartner(
            tokenId,
            partnerIndex,
            positionSize,
            atTick,
            poolUtilization
        );

        if (tokenId.isLong(index) == 0) {
            return _required + requiredPartner;
        } else {
            // return the max of the requirement between a loan and the long option position
            return Math.max(_required, requiredPartner);
        }
    }

    function _computeCreditOptionComposite(
        TokenId tokenId,
        uint128 positionSize,
        uint256 index,
        uint256 partnerIndex,
        int24 atTick,
        int16 poolUtilization
    ) internal view returns (uint256) {
        // can only be called when partnerIndex is the credit
        // required amount for short option leg
        uint256 _required = _getRequiredCollateralSingleLegNoPartner(
            tokenId,
            index,
            positionSize,
            atTick,
            poolUtilization
        );

        // amount moved in the credit leg
        LeftRightUnsigned amountsMovedCredit = PanopticMath.getAmountsMoved(
            tokenId,
            positionSize,
            partnerIndex
        );

        // get correct
        uint128 amountMoved = tokenId.tokenType(partnerIndex) == 0
            ? amountsMovedCredit.rightSlot()
            : amountsMovedCredit.leftSlot();

        if (_required > amountMoved) {
            return _required - amountMoved;
        } else {
            return 0;
        }
    }

    function _computeDelayedSwap(
        TokenId tokenId,
        uint128 positionSize,
        uint256 index,
        uint256 partnerIndex,
        int24 atTick
    ) internal view returns (uint256) {
        LeftRightUnsigned amountsMoved = PanopticMath.getAmountsMoved(tokenId, positionSize, index);

        LeftRightUnsigned amountsMovedP = PanopticMath.getAmountsMoved(
            tokenId,
            positionSize,
            partnerIndex
        );

        uint256 loanAmount = tokenId.tokenType(index) == 0
            ? amountsMoved.rightSlot()
            : amountsMoved.leftSlot();
        uint256 creditAmount = tokenId.tokenType(partnerIndex) == 0
            ? amountsMovedP.rightSlot()
            : amountsMovedP.leftSlot();

        uint256 convertedCredit = tokenId.tokenType(partnerIndex) == 0
            ? PanopticMath.convert0to1RoundingUp(creditAmount, Math.getSqrtRatioAtTick(atTick))
            : PanopticMath.convert1to0RoundingUp(creditAmount, Math.getSqrtRatioAtTick(atTick));

        if (loanAmount > convertedCredit) {
            return loanAmount - convertedCredit;
        } else {
            return 0;
        }
    }
```

flanagansteve (commented):
This is awesome. The main thing to fix is deciding whether these positions should be force-exercisable, and if not, ensure we skip width-0 positions in _forceExercise

I also toss out the idea of treating these loan positions as higher-priority liquidation claims, but that's up for discussion ofc

Other highlights: one comment where I'd appreciate some help understanding the collateral-requirement-calculation, another about how we should set the collateralisation ratio for borrow TokenIds

But nothing blocking aside from the FE thing!

contracts/RiskEngine.sol
flanagansteve: Oooo - what if the collateralisation ratio was 100% plus some scaling of the [adaptive interest rate](https://github.com/panoptic-labs/panoptic-next-core-private/pull/176)? More scarce assets have more stringent collateralisation requirements?

contracts/SemiFungiblePositionManager.sol
```diff
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.24;
-
```
flanagansteve, line 3: If we want positions with loan/credit legs to be transferrable, you will also need to add a zero-width guard in `registerTokenTransfer` (called in `safeTransferFrom` / `safeBatchTransferFrom`) - it calls PanopticMath.getLiquidityChunk which calls Math.getLiquidityForAmount0 and would then get a divide-by-zero in:

```solidity
/// @notice Returns a LiquidityChunk at the provided tick range with `liquidity` corresponding to `amount0`.
    /// @param tickLower The lower tick of the chunk
    /// @param tickUpper The upper tick of the chunk
    /// @param amount0 The amount of token0
    /// @return A LiquidityChunk with `tickLower`, `tickUpper`, and the calculated amount of liquidity for `amount0`
    function getLiquidityForAmount0(
        int24 tickLower,
        int24 tickUpper,
        uint256 amount0
    ) internal pure returns (LiquidityChunk) {
        unchecked {
            uint160 lowPriceX96 = getSqrtRatioAtTick(tickLower);
            uint160 highPriceX96 = getSqrtRatioAtTick(tickUpper);

            uint256 liquidity = mulDiv(
                amount0,
                mulDiv96(highPriceX96, lowPriceX96),
                highPriceX96 - lowPriceX96
            );

            // This check guarantees the following uint128 cast is safe.
            if (liquidity > type(uint128).max) revert Errors.LiquidityTooHigh();

            return LiquidityChunkLibrary.createChunk(tickLower, tickUpper, uint128(liquidity));
        }
    }
```

Because highPriceX96 will == lowPriceX96 and `highPriceX96 - lowPriceX96` is the denominator

PR #182, opened 2025-10-15
Feat(refactor): safe mode and account solvency math in risk engine

guil-lambert:
#### Summary

This PR introduces an architectural refactoring to improve the separation of concerns within the protocol. All core risk management and solvency calculation logic has been abstracted out of the `PanopticPool.sol` contract and encapsulated within the `RiskEngine.sol` contract.

**Note:** This is a pure refactoring effort. The underlying mathematical formulas and risk parameters for solvency assessment have **not** been altered. The goal is to improve code structure, readability, and maintainability without changing core functionality.

---

#### Description of Changes

The `PanopticPool` contract has been simplified to delegate these checks to the `s_riskEngine` instance.

The following logic has been moved from `PanopticPool` to `RiskEngine`:

* **`isAccountSolvent` Logic:** The full check to determine if an account has sufficient collateral has been moved into a new external function, `RiskEngine.isAccountSolvent`. This includes fetching the margin, calculating cross-balances, and applying the collateral buffer.
* **`getSolvencyTicks` Logic:** The logic to determine which oracle ticks to use for solvency checks (i.e., using multiple ticks during periods of high volatility) is now encapsulated in `RiskEngine.getSolvencyTicks`.
* **`isSafeMode` Logic:** The calculation to determine if the protocol should enter "safe mode" based on the deviation between the current tick and the median tick is now handled by `RiskEngine.isSafeMode`.
* **`MAX_TICKS_DELTA` Constant:** This constant has been moved to `RiskEngine` to reside alongside the logic that uses it, cleaning up the constants in `PanopticPool`.

---

#### Key Benefits

This refactoring provides several key advantages:

* **Improved Readability & Maintainability:** `PanopticPool` is now easier to understand, as its functions clearly state their intent (e.g., `s_riskEngine.isAccountSolvent(...)`) rather than containing complex implementation details.
* **Single Responsibility Principle:** This change better adheres to the SRP. The `RiskEngine` is responsible for *how* risk is calculated, while the `PanopticPool` is responsible for *what* to do with that information.
* **Enhanced Testability:** The complex mathematical logic for solvency can now be unit-tested in isolation within the `RiskEngine`, leading to more robust and targeted tests.
* **Modularity & Upgradability:** Future modifications to the risk model will be cleaner and less intrusive, as changes can be scoped primarily to the `RiskEngine` contract.

---

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1390,22 +1390,7 @@ contract PanopticPool is Multicall {
     function isSafeMode() public view returns (bool) {
         int24 currentTick = SFPM.getCurrentTick(s_univ3pool);
         uint256 medianData = s_miniMedian;
-        unchecked {
-            // If ticks have recently deviated more than +/- ~10%, enforce covered mints
-            return
-                Math.abs(
-                    currentTick -
-                        (int24(
-                            uint24(medianData >> ((uint24(medianData >> (192 + 3 * 3)) % 8) * 24))
-                        ) +
-                            int24(
-                                uint24(
-                                    medianData >> ((uint24(medianData >> (192 + 3 * 4)) % 8) * 24)
-                                )
-                            )) /
-                        2
-                ) > MAX_TICKS_DELTA;
-        }
+        return s_riskEngine.isSafeMode(currentTick, medianData);
     }
 
     /*//////////////////////////////////////////////////////////////
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -127,10 +127,6 @@ contract PanopticPool is Multicall {
     /// @dev Ensures token substitution between two accounts is settled safely at a price close to market.
     int256 internal constant MAX_TICK_DELTA_SUBSTITUTION = 203;
 
-    /// @notice The maximum allowed cumulative delta between the fast & slow oracle tick, the current & slow oracle tick, and the last-observed & slow oracle tick.
-    /// @dev Falls back on the more conservative (less solvent) tick during times of extreme volatility, where the price moves ~10% in <4 minutes.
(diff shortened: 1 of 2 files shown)


contracts/PanopticPool.sol
```diff
@@ -1366,46 +1347,25 @@ contract PanopticPool is Multicall {
...
-        );
-        (uint256 balanceCross, uint256 thresholdCross) = PanopticMath.getCrossBalances(
-            tokenData0,
-            tokenData1,
-            Math.getSqrtRatioAtTick(atTick)
-        );
-
-        // compare balance and required tokens, can use unsafe div because denominator is always nonzero
-        return balanceCross >= Math.mulDivRoundingUp(thresholdCross, buffer, 10_000);
+        return
```
flanagansteve, line 1350: Now that this is just a one-liner, i would just kill `_isAccountSolvent` and call `s_riskEngine.isAccountSolvent` directly

contracts/RiskEngine.sol
```diff
@@ -45,6 +45,10 @@ contract RiskEngine {
     /// @dev uint type for composability with unsigned integer based mathematical operations.
     uint256 internal constant DECIMALS = 10_000_000;
 
+    /// @notice The maximum allowed cumulative delta between the fast & slow oracle tick, the current & slow oracle tick, and the last-observed & slow oracle tick.
+    /// @dev Falls back on the more conservative (less solvent) tick during times of extreme volatility, where the price moves ~10% in <4 minutes.
```
flanagansteve, line 49: It _could_ be neat to make this configurable per-pool - store a `mapping(address panopticPoolAddress => uint maxTicksDelta)`. Not a must-have in my opinion though since we generally don't want this to kick in anyways

flanagansteve (approved):
Awesome!

PR #184, opened 2025-10-20
Feat: Optimize gas for margin and solvency calculations

guil-lambert:
**Description:**

This PR introduces a significant refactoring of the margin and solvency check logic to reduce gas consumption during these critical operations.

The optimizations are achieved through two primary changes:

### 0. Streamlined Position Balance Array

The `positionBalanceArray` no longer returns the `tokenId` (which the caller already has) and now provides a simple `uint256[]` of `positionBalance` data. This allows for the use of `calldata` instead of `memory` in several function signatures, saving gas.

### 1. Batched Collateral Calls
A new `external view` function, `assetsAndInterest`, has been added to `CollateralTracker.sol`. This function returns both the user's asset balance (via `convertToAssets`) and their `_owedInterest` in a single call.

The `RiskEngine` now uses this function, reducing what was **two separate external calls** (`assetsOf` and `owedInterest`) per collateral token down to **one single call**. This effectively halves the gas cost for fetching a user's collateral and interest data.

### 2. Consolidated Requirement Loop
The `_getTotalRequiredCollateral` function in `RiskEngine.sol` has been refactored to compute the requirements for **both token0 and token1 in a single loop**.

Previously, this function was called *twice* (once for each token), forcing the contract to loop through all of a user's positions two separate times. This change halves the gas cost of the position-looping logic.

---

### Other Changes:

* **Data Flow:** The `positionIdList` is now passed down the call chain (`_checkSolvencyAtTicks` -> `_isAccountSolvent` -> `getMargin`, etc.) to facilitate the new `_getTotalRequiredCollateral` logic.
* **Natspec & Style:** All Natspec comments have been updated to reflect the new function signatures, and minor style improvements (like caching `PositionBalance` in a local variable) have been made for readability.
* **`unchecked`:** The main `_getMargin` function logic is now wrapped in an `unchecked` block for additional gas savings, as the arithmetic is safe from overflow/underflow.

## Release Notes

* **New Features**
  * Added `assetsAndInterest()` function to retrieve collateral assets and interest owed in a single query

* **Refactor**
  * Streamlined balance data structure for improved clarity in position and margin calculations
  * Enhanced margin and solvency checking to use explicit position tracking lists
  * Updated collateral calculations to separately track position IDs and balance data

* **Bug Fixes**
  * Added validation to ensure position ID and balance data arrays remain synchronized


guil-lambert:
@coderabbitai please review this PR

PR #189, opened 2025-10-31
Feat: Factor Net Liquidation Value (NLV) into liquidation bonus

guil-lambert:
### Summary

This PR refines the liquidation bonus calculation to make it more accurate and fair.

Previously, the liquidation bonus was calculated based on the user's collateral deficit (`required_collateral - available_collateral`) alone. This approach did not account for the intrinsic value of the user's option portfolio itself. As a result, a liquidator could be overpaid if they were liquidating a user who was undercollateralized but held a portfolio with a positive Net Liquidation Value (NLV) (e.g., long in-the-money options).

This change introduces the portfolio's NLV into the bonus calculation. The bonus is now based on the user's *net* deficit, effectively calculated as:

`Deficit = required_collateral - (available_collateral + max(0, NLV))`

This ensures that any positive value in the user's option book is used to offset their collateral requirement *before* a bonus is calculated, leading to smaller, more appropriate bonuses. That bonus could be zero for some accounts (eg. with deep ITM long options).

### Changes

1.  **`RiskEngine.sol`:**
    * Adds a new `public view` function `getNetLiquidationValue`. This function iterates through a user's `positionIdList` to calculate the portfolio's total intrinsic value (NLV) in token0 and token1 at a given `atTick`.
    * Updates `getLiquidationBonus` to accept a new `LeftRightSigned netLiq` parameter.
    * The core bonus logic is modified to use this `netLiq`. It calculates a `netLiqCross` (the price-adjusted NLV) and subtracts it from the collateral deficit. If the user is net solvent (`collateral + NLV >= required`), the bonus is correctly set to `0`.

2.  **`PanopticPool.sol`:**
    * In the `_liquidate` function, the pool now calls `s_riskEngine.getNetLiquidationValue` to get the portfolio's NLV *before* calculating the bonus.
    * This `netLiq` is then passed into the `s_riskEngine.getLiquidationBonus` function.

3.  **`PanopticMath.sol`:**
    * Adds an overloaded `getCrossBalances` function that accepts a `LeftRightSigned` struct. This is a helper function needed to convert the two-token `netLiq` (which can be positive or negative) into a single, price-adjusted `int256` value for the bonus calculation.

---

### ⚠️ To-Do (WIP)

This PR contains the core logic but is still a work-in-progress. The following are required before it's ready for merge:

* [ ] Add new unit tests for `RiskEngine.getNetLiquidationValue`.
* [ ] Update existing tests for `RiskEngine.getLiquidationBonus` to cover new scenarios with positive, negative, and zero NLV.
* [ ] Update integration/fork tests for `PanopticPool._liquidate` to ensure the end-to-end flow is correct.

### DIFF
--- a/contracts/PanopticPool.sol
+++ b/contracts/PanopticPool.sol
@@ -1013,6 +1013,7 @@ contract PanopticPool is Multicall {
         LeftRightUnsigned tokenData0;
         LeftRightUnsigned tokenData1;
         LeftRightUnsigned shortPremium;
+        LeftRightSigned netLiq;
         {
             uint256[] memory positionBalanceArray = new uint256[](positionIdList.length);
             LeftRightUnsigned longPremium;
@@ -1022,6 +1023,13 @@ contract PanopticPool is Multicall {
                 COMPUTE_PREMIA_AS_COLLATERAL,
                 ONLY_AVAILABLE_PREMIUM,
                 currentTick
+            );
+            netLiq = s_riskEngine.getNetLiquidationValue(
+                positionIdList,
+                positionBalanceArray,
+                twapTick,
+                shortPremium,
+                longPremium
             );
             (tokenData0, tokenData1) = s_riskEngine.getMargin(
                 liquidatee,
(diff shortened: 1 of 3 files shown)


guil-lambert:
Will not merge, too much impact on liquidation economics

PR #232, opened 2026-01-13
Fix deploy scripts

zjesko:


0xTARC (approved):

0xTARC (commented):
This looks great, though I believe we will want the Uniswap V4 - compatible versions of these contracts, not V3

Guillaume mentioned this a few days ago in standup. Waiting on his response in slack..

Also, when you successfully deploy this, can you post the factory addresses in a comment here or in slack so i can add them to the indexer?

0xTARC (changes requested):
OK confirmed we do indeed want to launch only on V4 to start

https://panoptic-global.slack.com/archives/C04F33PCH0S/p1768406787206599?thread_ts=1768405848.097989&cid=C04F33PCH0S

